PR #18, opened 2018-12-29
Corrected value for scrollTo behaviour parameter

DvdGiessen:
TypeScript was (rightfully so) complaining about the parameter passed to the `window.scrollTo` function. The value `instant` is not a recognized behaviour value, and a quick search did not turn up any browser that currently recognized this value. Per the CSS standard draft, `auto` is used to specify instant scrolling.

See also https://developer.mozilla.org/en-US/docs/Web/API/ScrollToOptions/behavior and https://drafts.csswg.org/cssom-view/#enumdef-scrollbehavior.

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -470,7 +470,7 @@ export class Replayer {
           this.iframe.contentWindow!.scrollTo({
             top: d.y,
             left: d.x,
-            behavior: isSync ? 'instant' : 'smooth',
+            behavior: isSync ? 'auto' : 'smooth',
           });
         } else {
           try {


Yuyz0112:
Interesting, I think you are right but typescript also hints `instant` is a valid value in my editor.

image

DvdGiessen:
Hm, looking into it that seems to be changed in the 3.2 release of TypeScript about a month ago, specifically in this commit: https://github.com/Microsoft/TypeScript/commit/40bd7c89abbdf79493e2ea20dd80d9eb657dc82e#diff-46fd87925e4552c166ec188712741c3fL17754

Yuyz0112:
Thanks for the investigation!

Issue #16, opened 2018-12-28
Ability to stop recording


ChuckJonas:
Seems like it would make sense to have the ability to stop recording.

Yuyz0112:
Oh, my fault about missing this in the doc.

The recorder already implemented the stop method like this:
```ts
```

PR #20, opened 2018-12-29
demonstrating how to stop recording

ChuckJonas:
Fix for #16

### DIFF
--- a/guide.md
+++ b/guide.md
@@ -58,18 +58,27 @@ rrweb.record({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


Yuyz0112:
Thanks! How about adding a standalone code example about stop recording between 'the simple one' and 'the real world one'? Which I think is more straightforward to understand:)

ChuckJonas:
@Yuyz0112 done

Yuyz0112:
Thanks!

PR #21, opened 2018-12-29
REPL tool enhancement

pd4d10:


Yuyz0112:
Actually, I'm planning to make the REPL tool better, such as users can re-input URLs to record different web pages without restart the REPL tool.

So I think I'm not going to merge this PR at this moment because things may change later. But really appreciate any suggestion to improve the REPL tool.

pd4d10:
Good idea. Are you already working on this? If not, I can do some tweak in this PR to match this proposal

In addition, Shall we add it to `bin` in package.json? So users can simply use `npx rrweb` to use CLI tool, instead of cloning the whole repo

Yuyz0112:
@pd4d10 good to hear that, actually I'm not coding something about that but would like to share my ideas.

I think rrweb should be a base of many toolings, and the REPL tool may finally become to something like rrweb-cli. The CLI can have some options like just record a URL and replay it, or save the recorded events into a static HTML file that can be used for debugging, or generate a JS file that will init a replayer and replay the session like a video file.

pd4d10:
Sounds great. Can I take this one? Seems we should start it at another repo.

Yuyz0112:
Finally, it can be a single repo, but currently another repo may make it inconvenient to develop.
For example, if you need to change something in the rrweb repo, you need update it first and wait for the publishing of new version, and then update it as a dependency of the new project.

Personally, I suggest to develop the tool as an npm script now and move it to another repo when things are more stable and clear.

pd4d10:
OK, let's do it in this PR. `readline` module is a bit low level, use [inquirer](https://github.com/SBoudrias/Inquirer.js) instead

Yuyz0112:
@pd4d10 Good to see the progress! Temporarily I add a `WIP` label to this PR, feel free to remove it when you think it is finished and I will start to review the code.

pd4d10:
Upgraded. Now users can do multiple records in one session

scripts/repl.ts
```diff
@@ -116,8 +136,10 @@ function getCode(): string {
     `);
   }
 
-  const tempFolder = path.join(__dirname, '../temp');
   function saveEvents() {
+    const tempFolder = path.join(__dirname, '../temp');
+    console.log(tempFolder);
```
Yuyz0112: Should we console temp folder path? We already print saved path with:
```js
console.log(`Saved at ${savePath}`);
```

Yuyz0112:
@pd4d10 LGTM! Would like this to be merged now?

pd4d10:
Yes, I think so. Seems only the owner can remove WIP label

Yuyz0112:
Aww, sorry for that. Thanks for the great work.

PR #25, opened 2018-12-30
Add ignore classname & block config

samjbmason:
This is the second part to this PR, the first part is (https://github.com/rrweb-io/rrweb-snapshot/pull/6) the `rrweb-snapshot` should be published first then this one will need to have the first as a dependency.

### DIFF
--- a/src/record/index.ts
+++ b/src/record/index.ts
@@ -46,7 +46,7 @@ function record(options: recordOptions = {}): listenerHandler | undefined {
           },
         }),
       );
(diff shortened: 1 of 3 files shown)


Yuyz0112:
@samjbmason sorry for the late, since there are some conflicts in this patch, I added the feature in https://github.com/rrweb-io/rrweb/commit/c04e4bf0a56ff4fa4f328c01ef77ac3bcf1c0432 .

Issue #45, opened 2019-01-21
[QUESTION] Best way to play live recording


L3o-pold:
I'd like to play in real time the record. Is that something possible?

I actually record a web page and send events to a backend that's broadcasting events with websocket.

What's the best way to add new fetched events to current replay?

I tried something like this:

```javascript
mixed.on('progress.update', function (data) {
    if (data.events && data.events.length) {
          if (!player) {
            player = new rrwebPlayer({
              target: document.getElementById('player'),
              data: data,
            });
          } else {
            for (var i = 0; i < data.events.length; i++) {
              player.options.data.events.push(data.events[i]);
            }
         }
    }
});
```

If that's not possible is it a planned feature? I'd like to make a live debugging feature for our customer, something similar to [www.fullstory.com](https://www.fullstory.com/).

Yuyz0112:
@L3o-pold Actually I have tried live-mode in a local branch, here is a preview: https://github.com/rrweb-io/rrweb/issues/30#issuecomment-451620595
I think it is definitely possible, an `addEvents` method is needed. Also, need to handle cases like network latency.

So this is in the future plan, and any discussion or contribution will be great to accelerate the progress.

mattc0llier:
First off - amazing work! thank you for sharing, this came just at the right time.

Just so you have some example use cases for live events... I am looking to stream live events to a player with a similar set up to youtube live. 

<img width="758" alt="Screen Shot 2019-03-20 at 17 44 26" src="https://user-images.githubusercontent.com/2234289/54706966-06739280-4b38-11e9-9539-bc6d7f0b0717.png">

I am sending live events to a backend and then broadcasting events using WebSockets for a live view with live pause, instant 30-second replay and then full replay via drag and drop scroll bar.

marcospassos:
This is one of the most wanted features for us.

@Yuyz0112 you demo looks promising! Any plans to make it public?

JagdishThavasi:
> I'd like to play in real time the record. Is that something possible?
> 
> I actually record a web page and send events to a backend that's broadcasting events with websocket.
> 
> What's the best way to add new fetched events to current replay?
> 
> I tried something like this:
> 
> ```js
> mixed.on('progress.update', function (data) {
>     if (data.events && data.events.length) {
>           if (!player) {
>             player = new rrwebPlayer({
>               target: document.getElementById('player'),
>               data: data,
>             });
>           } else {
>             for (var i = 0; i < data.events.length; i++) {
[...]

PR #73, opened 2019-05-26
start to support live mode

Yuyz0112:
1. add a liveMode flag to config, when liveMode is set, the timer
will keep running even though all the actions casted
2. add a public method addEvent, which will cast newly added event
in sync
3. move mouse in sync mode with the latest position info

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -69,6 +69,7 @@ export class Replayer {
       showWarning: true,
       showDebug: false,
       blockClass: 'rr-block',
+      liveMode: false,
     };
     this.config = Object.assign({}, defaultConfig, config);
 
@@ -155,6 +156,11 @@ export class Replayer {
     this.timer.addActions(actions);
     this.timer.start();
     this.emitter.emit(ReplayerEvents.Resume);
+  }
+
+  public addEvent(event: eventWithTime) {
+    const castFn = this.getCastFn(event, true);
+    castFn();
   }
 
   private setupDom() {
@@ -424,8 +430,10 @@ export class Replayer {
         break;
       }
       case IncrementalSource.MouseMove:
-        // skip mouse move in sync mode
-        if (!isSync) {
+        if (isSync) {
+          const lastPosition = d.positions[d.positions.length - 1];
+          this.moveAndHover(d, lastPosition.x, lastPosition.y, lastPosition.id);
+        } else {
           d.positions.forEach(p => {
             const action = {
               doAction: () => {
--- a/src/replay/timer.ts
+++ b/src/replay/timer.ts
@@ -45,7 +45,7 @@ export default class Timer {
           break;
         }
       }
-      if (actions.length > 0) {
+      if (actions.length > 0 || self.config.liveMode) {
         self.raf = requestAnimationFrame(check);
       }
     }
--- a/src/types.ts
+++ b/src/types.ts
@@ -241,6 +241,7 @@ export type playerConfig = {
   showWarning: boolean;
   showDebug: boolean;
   blockClass: string;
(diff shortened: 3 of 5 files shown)


Yuyz0112:
related to #45

eoghanmurray:
Is this missing an `events.push`?  e.g. `getMetaData().totalTime` reports the wrong time after events have been added;

```        Replayer.prototype.addEvent = function (event) {
            this.events.push(event);
            var castFn = this.getCastFn(event, true);
            castFn();
        };
```

rifaideen:
@Yuyz0112 - First of all, this is great tool. Can you please share the example on iveMode?

```js
    new Player({
      target: document.getElementById('preview'),
      data: {
        events: recordings,
        autoPlay: true,
      },
      liveMode: true,
    })
```
I'm trying to broadcast live event but not sure where I'm wrong.

eoghanmurray:
The `liveMode` is more of an internal flag to keep the playback loop running ... it could probably be removed if the playback loop could be restarted upon subsequent call to `addEvent`.

We got the live/cobrowsing functionality working ok by feeding the output of `emit` to `addEvent` — it wasn't working for us prior to the #191 change we did so I imagine there's not many using it yet.

If you wish to record in one browser and do live playback on a remote browser you likely are going to want to feed the events through a websocket server on both ends.  That'd be out of scope of this project as it's only one of many possible use cases, and there are many ways of implementing it.

rifaideen:
@eoghanmurray - Thank you. I have already figured out and it works well but I see the lagging in mouse pointer movements, while broadcasting the live session.

nikhilrayaprolu:
@rifaideen any example code on how you could implement cobrowsing?

rifaideen:
@nikhilrayaprolu - It should be something like this:

```js
import { Replayer, unpack } from 'rrweb';

let recordings = [];

let player = null;

function livePlay(data) {
  if (player === null && recordings.length >= 2) {
    player = new Replayer(recordings, {
      liveMode: true,
      root: document.getElementById('preview'),
      unpackFn: unpack,
    });
    player.play();
  } else if (recordings.length < 2) {
    recordings.push(JSON.parse(data));
  }

  if (player !== null) {
    player.addEvent(JSON.parse(data));
  }
});

// receive live events from the target page or element and call the below function
livePlay(data);
```

Yuyz0112:
Hi @nikhilrayaprolu @rifaideen 

I suggest you check my new PR https://github.com/rrweb-io/rrweb/pull/226 which refactoring the live mode with a more solid and easy to use API.

rifaideen:
@Yuyz0112 - Actually I was experiencing the lag in replay especially the mouse move events and hopefully this [#226](https://github.com/rrweb-io/rrweb/pull/226) would resolve this issue I think.

Yuyz0112:
@rifaideen Yes, the lag is one of the issues being solved.

rifaideen:
@Yuyz0112 - One more thing which is not related to rrweb, however, it's all about webpage heatmaps. do you have any sample implementations?

Yuyz0112:
@rifaideen Not yet. I do not have too much sense about how to implement a heatmap on the events.

rifaideen:
@Yuyz0112 - okay and one more question. what are the events supported by `rrweb-player`. I have tried to add event handler for `IncrementalSnapshot` event in the `replayer.addEventListener` method but it was not successful.

nikhilrayaprolu:
@rifaideen  it looks like the code sample you provided is an example for a live view but in case of co-browsing, the events are sent in a bidirectional manner. So is there a way to send mouse movements from the second screen to the first screen and first screen to the second screen?

PR #80, opened 2019-06-12
Option to mask inputs

s-jakob:
Added option 'maskAllInputs' to replace all user inputs with an Asterisk. This is important here in the EU to avoid recording of sensitiv user data.
ignoreClass is not always the best option, because sometimes you cant add the class everywhere. In addition the mask is nicer to watch, because you see that the user ist typing some stuff.

### DIFF
--- a/src/record/index.ts
+++ b/src/record/index.ts
@@ -25,6 +25,7 @@ function record(options: recordOptions = {}): listenerHandler | undefined {
     blockClass = 'rr-block',
     ignoreClass = 'rr-ignore',
     inlineStylesheet = true,
+    maskAllInputs = false,
   } = options;
   // runtime checks for user options
   if (!emit) {
@@ -161,6 +162,7 @@ function record(options: recordOptions = {}): listenerHandler | undefined {
             ),
           blockClass,
           ignoreClass,
+          maskAllInputs,
           inlineStylesheet,
         }),
       );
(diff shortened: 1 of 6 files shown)


Yuyz0112:
@s-jakob Thanks for the PR and the code looks good to me!

BTW, I think this option should also work for full snapshot which was recorded by the rrweb-snapshot lib. I'm going to apply this into it later this weekend and will release a new version after that.

s-jakob:
Very nice. I really like this library. I am working on another feature to add custom events (maybe "tags" is the better name to separate it from the recording events) to the recording to mark special events like checkout or add to card. Did you have something like this planed?

Yuyz0112:
@s-jakob Yean I know it is an important feature but not really started to work on that.

And I think a separate issue for some API and implementation discussion will help before you start to do that.

s-jakob:
@Yuyz0112 i just added an issue #81

eoghanmurray:
I wonder should we _not_ mask the following:
`<input type="range">`
This is just a single value so could hardly constitute sensitive user data?

Also date types are pretty innocuous to allow?
```
<input type="date">
<input type="datetime-local">
<input type="time">
<input type="week">
<input type="month">
```
https://www.w3schools.com/html/html_form_input_types.asp

PR #96, opened 2019-07-16
added 'pointer-events: none' for Replayer iframe

dphuang2:
'pointer-events: none' prevents the user from interacting with the DOM
inside of the iframe. This prevents accidental tampering by a user when
watching replays such as link clicks or DOM manipulation.

pointer-events supported by 97.15% usage: https://caniuse.com/#feat=pointer-events


Yuyz0112:
Thanks!

remisharrock:
all right, now I found why I can't copy/paste some text in the replayer :) maybe I need to implement a toggle to enable copy/pasting during playback

PR #116, opened 2019-08-10
Add show function to replayer

dphuang2:
The show function allows the replayer to jump to a specific point in time without actually starting the timer. This is useful for a drag functionality (though quite slow cause we rebuilding our snapshot), when you want to display a freeze frame in time, or when you simply do not want the replayer to call timer.start().

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -135,6 +135,20 @@ export class Replayer {
     this.emitter.emit(ReplayerEvents.Start);
   }
 
+  public show(timeOffset: number) {
+    this.timer.clear();
+    this.timer.timeOffset = 0;
+    this.baselineTime = this.events[0].timestamp + timeOffset;
+    for (const event of this.events) {
+      if (event.timestamp < this.baselineTime) {
+        this.getCastFn(event, true)();
+      } else {
+        break;
+      }
+    }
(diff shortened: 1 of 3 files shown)


Yuyz0112:
@dphuang2 You are right. Currently, the "drag to play" will not work because of the naive rebuild implementation.

I have a plan to build a quick `jumpTo` API which has the same functionality as your `show` API. I think this API without performance optimization is a little redundant because we can do this via `play + pause`.

dphuang2:
Good point, play + pause is essentially the same thing. No need to merge this. How do you plan to design the jumpTo API?

Yuyz0112:
@dphuang2 I still need some POC before implement it, but I think there are some performance killers we should avoid:
1. Many kinds of events are not needed to be cast during the jump. Like mouse move, mouse interactions, scroll, etc.
2. Maybe we need to maintain something like a virtual DOM tree and apply DOM mutations to it instead of performing to the real DOM. Because append/remove DOM can be expansive and many times we are experiencing create and delete DOM during the jump. Only the final result of virtual DOM tree need be applied to the real DOM.
3. If there are several full snapshots in the events, we should start from the nearest one instead of the first one.
4. It will be nice if can implement the UNDO of DOM mutations.

Issue #104, opened 2019-08-01
[Observer] Observer doesn't watch for CSSStyleSheet.insertRule


VladimirMilenko:
Currently, it is impossible to track calls to insertRule(which is used in CSS-in-JS libraries).
With this, it's impossible to get the styles, which are added dynamically during render phase.

I believe this can be solved by proxying calls  to `styleElement.sheet.addRule/insertRule`

VladimirMilenko:
I was able to do this in following way:

```
  handleStyleRoot = styleRoot => {
    const handler = {
      apply: function apply(target, thisArg, argumentsList) {
        const result = target.apply(thisArg, argumentsList);
        styleRoot.innerHTML = Array.from(thisArg.rules, x => x.cssText).join('\n');

        styleRoot.sheet.insertRule = new Proxy(styleRoot.sheet.insertRule, handler);

        return result;
      }
    };
    const proxy = new Proxy(styleRoot.sheet.insertRule, handler);
    styleRoot.sheet.insertRule = proxy;
    styleRoot.sheet.insertRule('.rr_track_on {}');
  };
```

jperl:
@VladimirMilenko where did you put this code?

jperl:
@Yuyz0112 @VladimirMilenko If you have guidance on how to add support for this, I would be happy to work on submitting a PR.

dacort:
Is this the same as #58 ?

PR #124, opened 2019-08-25
[WIP] hack together stylesheet observer

jperl:
This is my very hacky fix implementing https://github.com/rrweb-io/rrweb/issues/104

### DIFF
--- a/src/record/index.ts
+++ b/src/record/index.ts
@@ -1,5 +1,5 @@
 import { snapshot } from 'rrweb-snapshot';
-import initObservers from './observer';
+import initObservers, { observeStylesheet } from './observer';
 import {
   mirror,
   on,
@@ -80,9 +80,11 @@ function record(options: recordOptions = {}): listenerHandler | undefined {
       inlineStylesheet,
       maskAllInputs,
     );
+
     if (!node) {
       return console.warn('Failed to snapshot the document');
     }
+
     mirror.map = idNodeMap;
     wrappedEmit(
       wrapEvent({
@@ -181,6 +183,13 @@ function record(options: recordOptions = {}): listenerHandler | undefined {
           inlineStylesheet,
         }),
       );
+
+      for (var i in mirror.map) {
+        const node = mirror.map[i];
+        if ((node as any).tagName == 'STYLE') {
+          observeStylesheet(node as any);
+        }
+      }
     };
     if (
       document.readyState === 'interactive' ||
--- a/src/record/observer.ts
+++ b/src/record/observer.ts
@@ -506,6 +506,26 @@ function initInputObserver(
   };
 }
 
+export function observeStylesheet(styleRoot: HTMLLinkElement) {
+  const handler: ProxyHandler<any> = {
+    apply: function apply(target, thisArg, argumentsList) {
+      const result = target.apply(thisArg, argumentsList);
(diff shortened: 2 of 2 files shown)


marcospassos:
Hi @jperl, thank you for working on this.

One problem I see with your suggestion is that lack of [support of `Proxy`](https://caniuse.com/#feat=proxy). While not sure if this is the best idea to solve the problem, I would recommend using monkey-patching instead of a proxy, even if not ideal, it ensures backward compatibility.

jperl:
@marcospassos Good point I just copied in [this code](https://github.com/rrweb-io/rrweb/issues/104#issuecomment-517323099) without thinking about it -- that is an easy change to make.

Where do you recommend I inject the monkey patch -- for both the initial load and newly created elements?

Also how do you recommend testing this?

marcospassos:
@jperl, I'd take another path for solving this problem:

- Monkey-patch the `insertRule` and `removeRule` globally, so that we don't need to track recently changed or newly added elements.
- Create a mechanism to stop listening to changes as soon as the recorder is turned off
- Create a new [incremental snapshot event](https://github.com/rrweb-io/rrweb/blob/master/src/types.ts#L33) to explicitly report a CSS rule addition or removal
- Change the replayer to apply those changes to the document

Benefits
- Better performance, as we don't have to watch elements being added or changed
- More granularity in the event reporting (we'll only track the changes rather than the entire stylesheet)
- Improved browser support

```js
// How to monkey-patch the insertRule globally
const insertRule = CSSStyleSheet.prototype.insertRule;
CSSStyleSheet.prototype.insertRule = function(rule) {
    dispatchEvent(/*...*/);

    return insertRule.apply(this, arguments);
}
```

jperl:
@marcospassos That sounds like a much better approach. 

@Yuyz0112 if you can confirm that you would be willing to merge the approach @marcospassos recommended, I am happy to take a stab at implementing it.

Yuyz0112:
@marcospassos Thanks for the review!
@jperl I agree with the suggestions from @marcospassos and feel free to ping me if you meet any issue during implement this.

If you want to split this implementation into multiple PRs(which is recommended), I think it can be done with:
1. patch insertRule/deleteRule APIs.
2. add incremental events and observers in recording side.
3. replay the stylesheet events in the replaying side.

dcramer:
@jperl I might take a stab at moving forward with your work unless you were still actively looking into this. Let me know!

jperl:
@dcramer Please do! Sorry I have been too busy to tackle this yet.

dcramer:
I'm working on adapting this patch, and hoping someone can help clarify something.

I've disabled the observers and written a test which uses insertRule/deleteRule, however the test passes successfully where I'd expect it to fail. I believe thats because of the existing element mutation observers. How do I actually reproduce the core issue in isolation?

dcramer:
Actually I think i misunderstood how the snapshot tests were working. I'm guessing (skimming source) its based on an actual dom snapshot event vs the incremental build up events. Back at it!

Yuyz0112:
Looks like we can close this one since #177 has been merged?

@jperl @dcramer

dcramer:
:+1: i just need to figure out how to get things into the replayer but ill work on that next opportunity i get

PR #145, opened 2019-11-07
Changing Array.from to iterators.

IMFIL:
Array.from was leaving some set items out of the loop leading in missing nodes during replay. This PR replaces the Array.from(set) to iterators iterating through the set.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   "dependencies": {
     "@types/smoothscroll-polyfill": "^0.3.0",
     "mitt": "^1.1.3",
-    "rrweb-snapshot": "^0.7.18",
+    "rrweb-snapshot": "IMFIL/rrweb-snapshot#master",
     "smoothscroll-polyfill": "^0.4.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   "dependencies": {
     "@types/smoothscroll-polyfill": "^0.3.0",
     "mitt": "^1.1.3",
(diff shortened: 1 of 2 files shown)


Yuyz0112:
Hi @IMFIL, thanks for the PR!

Could you explain a little more about which situation will the `Array.from` cause missing items?

IMFIL:
It seems that when iterating through the values without using the iterator explicitly causes some nodes to be left out. I tested a few websites with the traditional for loop instead of the Array.From and missing nodes were present when using the tradition for loop. 

The websites include:

https://www.slaters.co.uk/checkout/onepage
https://nrml.ca/
http://www.abchome.com/

Yuyz0112:
Hi @IMFIL , sorry for the late!

After a deep debugging, I found(surprisingly) the site you mentioned overwrite the `Array.from` method.

In a normal site, if you print `Array.from`, you will get
```
Array.from
ƒ from() { [native code] }
```

And in these sites, you will get
```
Array.from
ƒ $A(e){if(!e)return[];if("toArray"in Object(e))return e.toArray();for(var t=e.length||0,n=new Array(t);t--;)n[t]=e[t];return n}
```

And if you look into the implementation of the second one, it was even wrong! It will convert any Set into an empty array.

I'm not sure which script overwrite this function, but google point to some PayPal cart code.

Anyway, I would like to merge the PR and tweak the code a little.

Yuyz0112:
https://github.com/prototypejs/prototype/blob/master/src/prototype/lang/array.js#L59
Looks like a prototypejs bug:(

Issue #54, opened 2019-02-14
Configurable threshold for events throttling


mariusrak:
It would be nice to be able to configure the value, so a user would be able to set how detailed and how often would events be emitted for higher precision or higher data savings.

https://github.com/rrweb-io/rrweb/blob/65fd31487c746c025223f877b75dacc40ae29c46/src/record/observer.ts#L232

Issue #111, opened 2019-08-06
Specific node snapshot


VladimirMilenko:
I think it could be a nice idea to make a specific node snapshot, this is especially useful for #104 related things.

In some sense it could be implemented like this:
```js
const someObject = new SomeClass({
  onEvent: (changedNode) => {
    rrweb.snapshot(changedNode);
   }
})
```
This is mostly helpful for cases, when changes are not tracked by mutation observer and are changing the inner props of the node, like `CSSStyleSheet` on `style` node.

Yuyz0112:
@VladimirMilenko Agree with that! BTW, snapshot a specific node will unlock the potential of rrweb-snapshot which can do a lot of things.

Currently, the concern is the first argument of snapshot actually has two semantic:
1. The document object of the scope.
2. The root node.

Maybe we need to separate it into two arguments, one for the document object and the other one for the root node which is optional. When the root node was not passed in, it will use the document node as a default value.

But the API of rrweb.snapshot is a little poor at this moment, which cannot extend another argument after document node. Maybe I need to update the API into an object-style argument and release a new minor version for rrweb-snapshot.

Yuyz0112:
The most tricky part of this feature is about CSS styles.

For example, if the DOM looks like this:
```
html
  body
    div #1
    div #2
```

and some style sheets like this:
```css
body > div {
  background: grey;
}

div:nth-child(2) {
  color: white;
}
```

And we only record div#2, then all the style sheets will not be applied to the recorded div#2.

randy-johnson:
I would be willing to live with that.   It could be easily documented.  Now I have to do some trickory like search / replace a classname in the player HTML before it plays to remove the divs.  It works fine but it would be nice to be able to record the div that we want.

romain130492:
Any update on this?

PR #147, opened 2019-11-15
Try add configurable threshold for events throttling

mpstv:
Tried to make an improvement #54. Prease review my code.

### DIFF
--- a/src/record/index.ts
+++ b/src/record/index.ts
@@ -35,6 +35,7 @@ function record(options: recordOptions = {}): listenerHandler | undefined {
     inlineStylesheet = true,
     maskAllInputs = false,
     hooks,
+    mousemoveWait = 50
   } = options;
   // runtime checks for user options
   if (!emit) {
@@ -181,6 +182,7 @@ function record(options: recordOptions = {}): listenerHandler | undefined {
             ignoreClass,
             maskAllInputs,
             inlineStylesheet,
+            mousemoveWait
           },
           hooks,
         ),
--- a/src/record/observer.ts
+++ b/src/record/observer.ts
@@ -279,7 +279,7 @@ function initMutationObserver(
   return observer;
 }
 
-function initMoveObserver(cb: mousemoveCallBack): listenerHandler {
+function initMoveObserver(cb: mousemoveCallBack, mousemoveWait: number): listenerHandler {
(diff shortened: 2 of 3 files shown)


Yuyz0112:
Hi @mpstv, thanks for the PR!

I think you can follow the implementation of other configurations in the recorder, such as the `blockClass`.

The steps to add a configuration can be:
1. add it to the record options(https://github.com/rrweb-io/rrweb/blob/master/src/record/index.ts#L28), and set a default value if needed(https://github.com/rrweb-io/rrweb/blob/master/src/record/index.ts#L33)
2. add it to the observer params(https://github.com/rrweb-io/rrweb/blob/master/src/record/index.ts#L180)
3. pass it to the specific observer(https://github.com/rrweb-io/rrweb/blob/master/src/record/index.ts#L180). In your case, it could be the second param of `initMoveObserver`.
4. use it in the specific observer.

mpstv:
Hi @Yuyz0112, thanks for fast feedback.
I changed my commit in according your recommendation. (I hope got it right).
Waiting for review my fixes.

Yuyz0112:
LGTM, going to merge this PR. Thanks for your contribution! @mpstv 

BTW, I think I would like to think a bit more about the configuration interface before publishing a new version.

For example, I may introduce a more general configuration like `sampling` and the interface may look like
```
{
  sampling: {
    mousemove: 50,
    scroll: 100,
    ...
  }
}
```

alexwenbj:
> LGTM, going to merge this PR. Thanks for your contribution! @mpstv
> 
> BTW, I think I would like to think a bit more about the configuration interface before publishing a new version.
> 
> For example, I may introduce a more general configuration like `sampling` and the interface may look like
> 
> ```
> {
>   sampling: {
>     mousemove: 50,
>     scroll: 100,
>     ...
>   }
> }
> ```

It is a nice proposal! I don't know are these features add to it?Moreover, is it possible to add some other configures to it?For example,I just want to record the actions inside an element, and ignore all the others,by now,I must configure all the ignore element classes.

Yuyz0112:
@alexwenbj https://github.com/rrweb-io/rrweb/issues/111 discuss record on a specific element.

Would you like to open an issue about sampling config for tracking?

PR #158, opened 2019-12-30
Deduplicate config for BROWSER_ONLY

eoghanmurray:
A little bit of de-duplication so that the BROWSER_ONLY config isn't specified twice.
Also intend to add another command to filter the output to just iife targets, so checking if this approach is a good idea.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
     "test": "npm run bundle:browser && cross-env TS_NODE_CACHE=false TS_NODE_FILES=true mocha -r ts-node/register test/**/*.test.ts",
     "test:watch": "PUPPETEER_HEADLESS=true npm run test -- --watch --watch-extensions js,ts",
     "repl": "npm run bundle:browser && cross-env TS_NODE_CACHE=false TS_NODE_FILES=true ts-node scripts/repl.ts",
-    "bundle:browser": "cross-env BROWSER_ONLY=true rollup --config",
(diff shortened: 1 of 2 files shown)


Yuyz0112:
I'm not sure that making the rollup config file more 'dynamic' is better for maintaining and understanding.

But I'm still open to the change of it if there are some stronger reasons.

eoghanmurray:
Me breaking it up helped me understand what the config was doing. The deduplication in particular helped me clarify that the BROWSER_ONLY was just a subset of the main build (and how it was).
My building needs are embodied by `browser-unpacked`, I'm hoping that this approach makes it easier for others to create similar configs.  I'm not fully happy with the naming though ('browser', 'suppress', 'only')

Yuyz0112:
@eoghanmurray Would you mind to check the latest master branch's rollup config? I think it may be a better base to extend these functionalities.

eoghanmurray:
I've reapplied the changes on top of the current master.
I appreciate my changes may not be easy to understand here with the added special keywords 'suppress' and 'only'.

> Would you mind to check the latest master branch's rollup config? I think it may be a better base to extend these functionalities.

I really don't understand all the different configs or what they are for.
Please go ahead and close this pull request if it is not providing any benefit!

eoghanmurray:
Close as I've abandoned attempts to keep this up to date w.r.t. master and did a custom hack instead to prevent compilation of non `iife` targets.

PR #177, opened 2020-02-21
Add observers for stylesheet mutations

dcramer:
This captures the insertRule and deleteRule actions on CSSStyleSheet, and wires them up to StyleSheetRule events in the recorder.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "record and replay the web",
   "scripts": {
     "test": "npm run bundle:browser && cross-env TS_NODE_CACHE=false TS_NODE_FILES=true mocha -r ts-node/register test/**/*.test.ts",
+    "test:watch": "PUPPETEER_HEADLESS=true npm run test -- --watch --watch-extensions js,ts",
     "repl": "npm run bundle:browser && cross-env TS_NODE_CACHE=false TS_NODE_FILES=true ts-node scripts/repl.ts",
     "bundle:browser": "cross-env BROWSER_ONLY=true rollup --config",
     "bundle": "rollup --config",
--- a/src/record/observer.ts
+++ b/src/record/observer.ts
@@ -21,6 +21,7 @@ import {
   MouseInteractions,
   listenerHandler,
   scrollCallback,
+  styleSheetRuleCallback,
   viewportResizeCallback,
   inputValue,
   inputCallback,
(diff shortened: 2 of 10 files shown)


dcramer:
We should squash this for sure, but let me know if the approach is how you'd go about it.

dcramer:
Also this fixes GH-176

dcramer:
cc @marcospassos

test/integration.test.ts
```diff
@@ -35,14 +35,7 @@ describe('record integration tests', function(this: ISuite) {
...
   before(async () => {
-    this.browser = await puppeteer.launch({
-      defaultViewport: {
-        width: 1920,
-        height: 1080,
-      },
-      headless: false,
-      args: ['--no-sandbox'],
-    });
+    this.browser = await launchPuppeteer();
```
dcramer: these changes are unrelated, but I was annoyed with running --watch and having the browser constantly take focus

test/__snapshots__/record.test.ts.snap
```diff
@@ -331,3 +331,130 @@ exports[`custom-event 1`] = `
...
+        \\"left\\": 0,
+        \\"top\\": 0
+      }
+    }
+  },
+  {
+    \\"type\\": 3,
+    \\"data\\": {
+      \\"source\\": 8,
+      \\"id\\": -1,
```
dcramer: guessing the use of the mapper here is wrong, could someone guide me on the expectation of id and how it needs to function?

dcramer:
Also looks like the test suite is a bit flakey (sometimes the async code doesnt finish running as expected?)

Have seen it on the new tests + on:

```
     AssertionError: Received value does not match stored snapshot "async-checkout 1".
```

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "record and replay the web",
   "scripts": {
     "test": "npm run bundle:browser && cross-env TS_NODE_CACHE=false TS_NODE_FILES=true mocha -r ts-node/register test/**/*.test.ts",
+    "test:watch": "PUPPETEER_HEADLESS=true npm run test -- --watch --watch-extensions js,ts",
```
dcramer: can back this out if folks disagree but found it useful

aside, it doesnt recompile src/ and I couldn't figure out how to achieve that

src/record/observer.ts
```diff
@@ -519,6 +520,34 @@ function initInputObserver(
   };
 }
 
+function initStyleSheetObserver(cb: styleSheetRuleCallback): listenerHandler {
+  const insertRule = CSSStyleSheet.prototype.insertRule;
+  CSSStyleSheet.prototype.insertRule = function(
+    rule: string,
+    index?: number | undefined,
+  ) {
+    cb({
```
dcramer: one question I had here - do people prefer this version of a single type, or should we have two types which means we could better compress the data

Yuyz0112: I think this implementation keeps simplicity. Data compressing stuff can be done by the packer plugin in a more general aspect.

src/record/observer.ts
```diff
@@ -519,6 +520,34 @@ function initInputObserver(
   };
 }
 
+function initStyleSheetObserver(cb: styleSheetRuleCallback): listenerHandler {
+  const insertRule = CSSStyleSheet.prototype.insertRule;
+  CSSStyleSheet.prototype.insertRule = function(
+    rule: string,
+    index?: number | undefined,
```
Yuyz0112: looks like `index?: number` is enough since `?` already make type optional

package.json
Yuyz0112: Could not get that done too...
Combine Mocha and Typescript is not as fluent as jest.

Yuyz0112:
@dcramer Thanks a lot! I would love to merge it after tweak some small changes.

The test things can be done in the following PRs.

dcramer:
@yuzy0112 I fixed the type - were you asking me to pull the test:watch and other changes out?

Yuyz0112:
Nope, it's fine.

PR #178, opened 2020-02-24
Add support for replaying StyleSheetRule events

### LINKED ISSUE #58: Missing styles in recording
Hi!

First I want to say that this project is amazing to use, simple and fast to implement! :)

I have 2 react SPA, one is relativley straight forward, while the other is much more complex. The simple site works perfectly! :) 

But the slightly more complex site looks like this:
image

But in the records it looks like this instead:
image

This site uses styled-components heavily, and the grey color comes from the default index.scss.

My code looks like this:
```js
const useRRWeb = () => {
    const [allow, setAllow] = useState(false)
    const eventsRef = useRef([])

    const allowRecording = () => setAllow(true)
    const getRandom = () => Math.random().toString(36).substring(7);
    import("rrweb").then(({record}) => {
        record({
            emit: (event) => {
                eventsRef.current.push(event)
            }
        })
    })
    useEffect(() => {
        if (!allow)
            return;

        const sessionId = getRandom() + getRandom() + getRandom();
        const save = () => {
            const events = eventsRef.current
            const body = JSON.stringify({events});
            eventsRef.current = [];

            fetch(getUrl(sessionId), {
                method: 'POST',
                headers: {
                    'Content-Type': 'application/json',
                },
                body,
            }).catch(() => {
                eventsRef.current.unshift(events)
            });
        }
        const saveInterval = setInterval(save, 10 * 1000);

        record({
            emit: (event) => {
                eventsRef.current.push(event)
            }
        })

        return () => {
            clearInterval(saveInterval)
            save()
        }
    }, [allow])

    return [allow, allowRecording]
}
export default useRRWeb
```

Any ideas about where I should start to look?

### LINKED ISSUE #104: [Observer] Observer doesn't watch for CSSStyleSheet.insertRule

dcramer:
Fixes #58 
Fixes #104

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -608,6 +608,28 @@ export class Replayer {
         }
         break;
       }
+      case IncrementalSource.StyleSheetRule: {
+        const target = mirror.getNode(d.id);
+        if (!target) {
+          return this.debugNodeNotFound(d, d.id);
+        }
+
+        const styleEl = (target as Node) as HTMLStyleElement;
+        const styleSheet = <CSSStyleSheet>styleEl.sheet;
+
+        if (d.adds) {
+          d.adds.forEach(({ rule, index }) => {
+            styleSheet.insertRule(rule, index);
+          });
+        }
+
+        if (d.removes) {
+          d.removes.forEach(({ index }) => {
+            styleSheet.deleteRule(index);
+          });
+        }
+        break;
+      }
       default:
     }
   }


dcramer:
Tested via this and it seemed to work, though the id recorded was -1 (maybe thats correct?), so I was a bit confused:

https://codesandbox.io/s/beautiful-babbage-hz2wj

Here's the output: https://gist.github.com/dcramer/3b8449cf4388663ab7f0c1dbec399fe6

dcramer:
Aside - willing to help try to get test coverage in for replays (beyond just testing the controller actions), but didn't see any existing.

Yuyz0112:
Interesting.

The example in the gist works because it captures the added DOM:
```
{
  "source": 0,
  "texts": [],
  "attributes": [],
  "removes": [],
  "adds": [
    {
      "parentId": 4,
      "previousId": 52,
      "nextId": null,
      "node": {
        "type": 2,
        "tagName": "style",
        "attributes": {
          "_cssText": "p { color: red; }"
        },
        "childNodes": [],
        "id": 53
      }
    }
  ]
}
```

Yuyz0112:
The replayer code looks fine.

I believe the `id: -1` happens because of the async callback of MutationObserver.

Looks like **the order of the dynamic stylesheet insertion** is:
1. create a `<style>` element
2. call `insertRule` on the style element

But **the order of rrweb's observer callback** is:
1. Capture the `insertRule` calling, because it triggered in sync.
2. Capture the add of `<style>` element, because MutationObserver will fire in the next async loop.

I'm not sure whether this works for all the CSS in JS lib, it seems the first insertion will be replayed by the add of DOM, and the following insertion weill be captured correctly.
But I think put the `insertRule` observer callback into a macro task(maybe by using `setTimeout(, 0)`) will help us always get the right order.

dcramer:
Seems like it’d be nice to denounce insertRule until the style is on the dom. Thoughts? Not sure how complex that’d be.

PR #189, opened 2020-03-30
Fix incorrect restart of playback while artificially paused

eoghanmurray:
Keep track of pause/play state so that player doesn't accidentally 'unpause' a user pause action.

I can reliably recreate the problem on a recording with stylesheets by doing:
`player.on('fullsnapshot-rebuilded', player.pause);`

The playback sometimes pauses, but depending on the timing of how external stylesheets load, the pause may not work, and play will continue.

(Pausing on the first fullsnapshot-rebuilded is very useful in order to present user with a 'screenshot' of the first screen before starting playback)

I haven't created a test case as none of the existing testcases rely on an external stylesheet (which could introduce a delay to the tests?), let me know if it's okay (and how) to introduce such a test case.

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -57,6 +57,8 @@ export class Replayer {
 
   private missingNodeRetryMap: missingNodeMap = {};
 
+  private playing: boolean = false;
+
   constructor(events: eventWithTime[], config?: Partial<playerConfig>) {
     if (events.length < 2) {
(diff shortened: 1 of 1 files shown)


Yuyz0112:
Yep, this issue also makes the rrweb-player's auto play config unstable.

The fix is nice and I'm going to merge it. And then I would like to introduce a state machine to describe the replayer's state and makes it easier to test and integrate.

Issue #182, opened 2020-03-02
Node  id xxx is not found.


arun477:
hey, great tool man, I'm facing some issue with replaying my recorded events. when I try to replay the recorded events, I can see a lot of warning messages "Node with id xxx not found" like this but my video somehow playing but not properly( it happens if a number of event increase for example: if total event size is 1k it playing well but when it increased to more than 1k it started playing improperly).  my application is react based and I'm using styled-components("CSS in js library) and I'm storing every event as a string(JSON.stringify(event)) in DB. at playing time query it from DB parse it using JSON.parse(event) and trying to play. 

thanks.

arun477:
and I have two more questions.
1. is the order of events important or we can provide events in any order to the player.
2. is it ok to use JSON.stringify and JSON.parse (it won't corrupt event data?)

Yuyz0112:
Hi @arun477 , could you provide a sample events log for debugging?

For your questions:
1. Yes, we need the events being ordered by timestamp.
2. Yes, the events are serializable.

arun477:
hey, thank for the quick reply.  my logs are sensitive to share, so I can't share could you tell me what causes this  issue "Nodeid xxx is not found" I try to debug myself( if not I will simulate sample env and email you but it will take some time for me to generate that sample log)

Yuyz0112:
Feel free to re-open this issue if I can help.

eoghanmurray:
@arun477 were you using `player.addEvent` to add events?

jackiehamchen:
Hello @arun477 
are you able to solve the issue? could you please share with me as i'm experiencing the same error message while replaying?

wbcs:
inlineStylesheet:true

PR #191, opened 2020-03-31
Add event async

eoghanmurray:
Enable addEvent to be used while there are still queued actions

 - addEvent was originally added to support 'live mode' (#73)
 - it was failing badly if used to add events before playback of current events has concluded, as assumption was that we were at end of playback
 - event wasn't getting added to list of events leading to incorrect totalTime metadata, and loss of added event upon replay from start
 - in a situation where events are being streamed to the replayer, this change supports asynchronous usage of addEvent (after you've added at least 2 initial events)

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -205,8 +205,21 @@ export class Replayer {
     const event = this.config.unpackFn
       ? this.config.unpackFn(rawEvent as string)
       : (rawEvent as eventWithTime);
-    const castFn = this.getCastFn(event, true);
-    castFn();
+    if (this.service.state.matches('playing')) {
+      const currentTime = this.events[0].timestamp + this.timer.timeOffset;
+      const isSync = event.timestamp < currentTime;
+      const castFn = this.getCastFn(event, isSync);
+      if (isSync) {
+        castFn();
+      } else {
+        const action = {
+          doAction: castFn,
+          delay: event.timestamp - currentTime,
+        };
+        this.timer.addAction(action);
+      }
+    }
+    this.events.push(event);
   }
 
   private setupDom() {


eoghanmurray:
Oh yeah, previously if an addEvent was called too early, before the event that builds the DOM was completed, then an `Node id xxx is not found.` error was the result.  Have questioned @arun477 on #182 to see if that was the case with them.

Yuyz0112:
#182 may also caused by the checkout feature.

Since the DOM observer Is async, we may capture some events after a full snapshot with invalid id.

eoghanmurray:
Cool well you can disregard that reference to #182 .  I gues `Node id xx is not found` is a common generic error when things go wrong. I believe I was encountering because `addEvent` was happening before the first `FULLSNAPSHOT-REBUILDED` completed.

Yuyz0112:
@eoghanmurray Would you like to rebase the master branch?

eoghanmurray:
@Yuyz0112 yes have done so!

Yuyz0112:
@eoghanmurray 

I have picked my work on the live mode to this PR: https://github.com/rrweb-io/rrweb/pull/226

Do you think I've solved the issues in this PR?

eoghanmurray:
@Yuyz0112 #226 looks good; am I correct in saying it allows the option of turning off the buffering of mouse movements?

However 
> event wasn't getting added to list of events 
I don't see that being fixed in #226 ?
One goal of this pull request is to allow calls to `addEvent` to work correctly in terms of pushing the new event to the end of the queue.
As it currently stands, maybe `addEvent` should be renamed `playLiveEventNow`.

E.g. one goal of this PR is when events are being streamed _in_ to the player (even in a non-live context), you want to start playback asap, before you've downloaded all events.

eoghanmurray:
As mentioned, the `addEvent` function in machine.ts now covers all this ground e.g. changes added in https://github.com/rrweb-io/rrweb/commit/8913bcb9d6bc50850474439b218b85a7e434d6a2

We're also now inserting in correct place after events are added with addEvent #411

So it's safe to close this one.

PR #193, opened 2020-04-07
Fix for certain websites which don't scroll on their documentElement

eoghanmurray:
- document.documentElement.scrollTop may be zero, but document.body.scrollTop may have the actual scrolling amount
 - main fallback idea taken from https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX
 - modified as `(document.documentElement || document.body).scrollTop` will incorrectly report zero.
 - version here supported by https://github.com/mochi/mochikit/blob/master/MochiKit/Position.js#L23

### DIFF
--- a/src/record/index.ts
+++ b/src/record/index.ts
@@ -97,8 +97,20 @@ function record<T = eventWithTime>(
         data: {
           node,
           initialOffset: {
-            left: document.documentElement!.scrollLeft,
-            top: document.documentElement!.scrollTop,
+            left: (window.pageXOffset !== undefined) ?
+	      window.pageXOffset : (
+		document!.documentElement.scrollLeft ||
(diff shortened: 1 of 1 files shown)


eoghanmurray:
Example website which previously reported zero for `document.documentElement.scrollTop` despite initial scrolling on a small viewport:
http://dearbhlaanddean.com/

eoghanmurray:
This is a hard bug to notice but you should be able to see that during playback, clicks which seem to be happening on empty space actually are successfully clicking on e.g. links.

I wonder is there any merit for recording the event.src element for a click event (if that's even possible to determine which element in the capture/bubble phase is the 'salient' target of the click).

Yuyz0112:
Thanks for the fix!

PR #198, opened 2020-04-12
Introduce state machine to implement player

Yuyz0112:
In this patch, we add `@xstate/fsm` which is a lightweight state machine library to refactor out player code.

The basic state machine of the player looks like this:
image

Since this is the first patch of our state machine refactoring, we do not make too much change on the code. Instead of it, we only replace the `playing` flag with a machine match.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "rollup-plugin-commonjs": "^9.2.0",
     "rollup-plugin-node-resolve": "^3.4.0",
     "rollup-plugin-postcss": "^1.6.2",
-    "rollup-plugin-terser": "^3.0.0",
+    "rollup-plugin-terser": "^5.3.0",
     "rollup-plugin-typescript": "^1.0.0",
     "ts-node": "^7.0.1",
     "tslib": "^1.9.3",
@@ -59,6 +59,7 @@
   },
   "dependencies": {
     "@types/smoothscroll-polyfill": "^0.3.0",
+    "@xstate/fsm": "^1.3.0",
     "mitt": "^1.1.3",
     "pako": "^1.0.11",
     "rrweb-snapshot": "^0.7.26",
--- a/src/declarations/@xstate/fsm/index.d.ts
+++ b/src/declarations/@xstate/fsm/index.d.ts
@@ -1,0 +1,22 @@
+import * as fsm from '@xstate/fsm';
+
+declare module '@xstate/fsm' {
+  export namespace StateMachine {
+    interface Service<
+      TContext extends object,
+      TEvent extends fsm.EventObject,
+      TState extends fsm.Typestate<TContext> = any
+    > {
+      send: (event: TEvent | TEvent['type']) => void;
+      subscribe: (
+        listener: StateListener<State<TContext, TEvent, TState>>,
+      ) => {
+        unsubscribe: () => void;
+      };
+      start: () => Service<TContext, TEvent, TState>;
+      stop: () => Service<TContext, TEvent, TState>;
+      readonly status: fsm.InterpreterStatus;
+      readonly state: State<TContext, TEvent, TState>;
+    }
+  }
+}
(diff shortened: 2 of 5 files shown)


eoghanmurray:
So I've just updated to latest (0988e453016f6d), and I believe this commit (and subsequent ones e.g. 8913bcb9d6bc508504744) have changed the API interface for rrweb (i.e. broken my client code!)

In a 20 second recording, if I call `player.play(10000)` after playback has completed (e.g. the user drags the scrubber back to the 10 second mark), then nothing happens.  Within the `play` function, inspecting `this.service.state.value` shows that it is `"ended"`.

Stepping through the state machine code in the debugger is also very opaque, I think it's that @xstate/fsm code that is included in rrweb.js already minified?  At any rate I can't figure out exactly why it is not playing by stepping further than the `play` call.  I presume it's related to the fact that the `"ended"` state needs a `replay()` call to get out of that state, but I don't see any `replay` function available.

I'm sure there are sound reasons for adding a fsm but from a collaboration point of view it's not making things easier to hack with the extra layer of indirection.

eoghanmurray:
#235 addresses one thing that this refactor has removed; the inability to directly access `.events` during replaying.
(At least I think this is the change that removed `.events`)

eoghanmurray:
See also #236 for getting the `.play()` method working like it used to

eoghanmurray:
These changes have broken a lot of things!!

Yuyz0112:
@eoghanmurray Did it broken some public APIs?

eoghanmurray:
Yes! Basically everything to do with jumping around a recording is broken since these and the subsequent related changes (what we call 'scrubbing' around a recording, i.e. drag the slider on the video).

I've spent all of today further trying to figure out the purpose of introducing this FSM, i.e. what it solves!

Today I'm trying to fix the following but I can't work out how I should implement it:
 - At the `ended` state, start playing at an arbitrary point in time. (why do we need an ended state? can't we just pause at the end?)  I did #236 to enable the playing, but actually it's still all broken in multiple ways (the `lastPlayedEvent` is incorrectly taken into account - see #250 but it is still not working for me and I cannot work out why)
 
I just don't understand what the FSM accomplishes?  Why should the player be prevented from going to certain states from certain others?

The FSM also completely ignores the _time_ that you wish to play from. Can I pause at time X and play at time Y?  Can I be in the `ended` state when the current time is zero?   What if I am in the `playing` state and wish to jump forward 10s and continue playing, do I have to exit the `playing` state or not?   What does `skipping` mean?

Can we roll this all back?? My boss is suggesting we just freeze/fork from before this change after all the time I've spent on it trying to get the player working again.

Yuyz0112:
@eoghanmurray Really appreciate this feedback and thanks to your continuous contribution to rrweb. I'm sorry this patch cause your issues and I will write down some more details about why I introduce state machine to the replayer.

I manage to do this tonight. And I think we will finally reach an agreement about the implementation.

eoghanmurray:
Any update on this?

Just thinking, instead of a FSM, what would be wrong with modelling playback state as follows:

`playing: boolean`
`currentTime: integer (milliseconds)`
`playbackSpeed: float (default 1x)`
`liveBuffer: integer (milliseconds)`

So if you pause, playing becomes `false`.  Likewise if you get to the end of a recording, playing can become 'false', or in a live playback scenario, it will stay true, and currentTime will keep increasing.

Rather than defined transitions between states (FSM), any of the above should be able to be altered during playback (with the possible exception of `livebuffer`).  

Altering currentTime could cause a rebuild from most recent snapshot if it is decreased, or intervening events immediately applied if it is increased (it doesn't matter if you are playing or paused).  This would be what would happen when the drag&drop timeline scrubber is moved.

There would be no `resume` functionality #248 , as you'd just call `play` with a time if you are paused, and if the new play time matched the currentTime, then no rebuilding would have to occur.

Anything I am missing?

Yuyz0112:
I've just opened the new proposal for dicussing.

Yuyz0112:
https://github.com/rrweb-io/rrweb/issues/274

PR #200, opened 2020-04-15
Only execute events since most recent pageload

eoghanmurray:
On recordings with many full pageloads, dom state and mutations were being applied
only to be discarded when a new pageload came in, resulting in very slow time to
rebuild - and inability to interactively 'scrub' through the later parts of these
recordings.

### DIFF
--- a/src/replay/machine.ts
+++ b/src/replay/machine.ts
@@ -4,6 +4,7 @@ import {
   eventWithTime,
   actionWithDelay,
   ReplayerEvents,
+  EventType,
   Emitter,
 } from '../types';
 import { Timer, getDelay } from './timer';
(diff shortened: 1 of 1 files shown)


eoghanmurray:
The code probably needs a bit of explanation, as I have some 'unknowns':

 - I'm detecting 'new full pageload' by looking for consecutive Meta +FullSnapshot ... maybe I only need to look for Meta? wasn't sure on this
 - I'm also assuming there are no events that when applied before a pageload can have an effect after a FullSnapshot.. (e.g. prior browser window resize events are reset by the data.width/data.height of the Meta event)

eoghanmurray:
Updated and reapplied on top of FSM changes i.e. in machine.ts

Yuyz0112:
Thanks, I'll merge this and tweak a little to avoid mutating the events.

Issue #173, opened 2020-02-18
Certain DOM nodes are not serialized  in Safari but are being serialized in Chrome.


IMFIL:
How to reproduce:

1. Go to this URL:  https://www.skiny.com/pant-2er-pack-white-086835-080500.html
2. Follow steps in video: 
![steps](https://user-images.githubusercontent.com/16093847/74781607-c3ea9680-526f-11ea-98aa-4f2bbf89c775.gif)

Do the exact same steps on Safari and Chrome and the session replay on the: [https://www.skiny.com/checkout/#shipping](https://www.skiny.com/checkout/#shipping) page will be broken on safari and not on chrome.

This is the error message received while replaying the session in Safari. There is no such error during replay time for the session recorded in Chrome.

<img width="723" alt="Screen Shot 2020-02-18 at 4 58 59 PM" src="https://user-images.githubusercontent.com/16093847/74781700-f85e5280-526f-11ea-977d-d056e2522555.png">

Running the latest code for both the player as well as the recorder.

Here are the events for the safari recording:
[safari.txt](https://github.com/rrweb-io/rrweb/files/4221860/safari.txt)

Here are the events for the chrome recording:
[chrome.txt](https://github.com/rrweb-io/rrweb/files/4221862/chrome.txt)

Yuyz0112:
Thanks for the report @IMFIL 

Since my laptop OS is Linux, I think I need some more time to find a safari env to reproduce and fix the issue.

IMFIL:
This PR will prevent the re-player from breaking when this happens. It should buy us more time to fix it at it's core.

https://github.com/rrweb-io/rrweb/pull/174

Issue #86, opened 2019-07-04
[Question] Using for analytics use-case


boazpeer1:
Is it possible to use rrweb for analytics rather than replay?
I would only need the recorder and comprehensive documentation of the events so I can extract the required data which is needed for the analytics.

Is such documentation available?

marcospassos:
@boazpeer1 it's entirely possible, and in fact, it's the use case of some users.  Currently, most of the documentation covers specification details only, but the data structure is available at the [`types.ts`](https://github.com/rrweb-io/rrweb/blob/master/src/types.ts) file.

Notice that some of the information included in the events depends on post-processing on the replay part, so they cannot be consumed without tweaks yet. However, there are some ongoing efforts to fix it, see #82 and #82.

Issue #83, opened 2019-06-28
[Proposal] Resolve node ID's on record-side


marcospassos:
I want to discuss the feasibility involves resolving all node ids at record-time instead of replay-time, as it'd make the code a lot simpler and more consumer-friendly.

I've implemented the following POC that resolves node ids regardless of the order in which the observer reports them:

```typescript
private resolveAddedNodes(mutations: addedNodeMutation[]): addedNodeMutation[] {
    const previous: { [key: number]: number } = {};
    const next: { [key: number]: number } = {};

    for (let mutation of mutations) {
        if (mutation.previousId != null && mutation.previousId >= 0) {
            previous[mutation.previousId] = mutation.node.id;
        }

        if (mutation.nextId != null && mutation.nextId >= 0) {
            next[mutation.nextId] = mutation.node.id;
        }
    }

    let additions: addedNodeMutation[] = [];

    for (let mutation of mutations) {
        let {previousId, nextId} = mutation;

        if (previousId !== null && previousId < 0) {
            previousId = next[mutation.node.id];
        }

        if (nextId !== null && nextId < 0) {
            nextId = previous[mutation.node.id];
        }

        additions.push({
            parentId: mutation.parentId,
            nextId: nextId,
            previousId: previousId,
            node: mutation.node,
        });
    }

    return additions;
}
```

Could you check the feasibility of this proposal?

Yuyz0112:
@marcospassos when would you like to execute this method, after a mutation callback?

marcospassos:
I'd suggest before emitting to listeners, in the recorder

Yuyz0112:
@marcospassos So it will execute in this order, right?
```
1. mutation observer callback
2. loop mutation records
3. resolve nodes
4. emit to listeners
```

According to this [commit](https://github.com/rrweb-io/rrweb/commit/661fa6c7eb2512fe60b592bec58925afd8a39599), sometimes the newly added nodes will not show in the same mutation observer callback loop.

marcospassos:
@Yuyz0112 can you give me a real case or test when it occurs?  When it happens, what is missing? Parents, siblings, or both? Can we use a buffer in the record side, instead of the replay side?

Yuyz0112:
@marcospassos I could not reproduce that with any test suite code. And since mutation observer callback is a microtask in the event loop, your code should work and be safe.

So how about adding this resolve logic in the record side, but also keep the replay side missing node resolver logic as a double check?

[...]

PR #206, opened 2020-05-02
mutation observer v2

Yuyz0112:
There are some long-term issues in rrweb's mutation observer.

A scenario cause problem:
A list of DOM node: n1, n2, n3, n4, n5
Steps of modifying the nodes:
  1. remove n1, n2, n3, n4 sequentially
  2. append n4, n3, n2, n1 after n5 sequentially
Then we got the added node data like this:
  (id: n4, prev: null, next: n3  )
  (id: n3, prev: n4,   next: n2  )
  (id: n2, prev: n3,   next: n1  )
  (id: n1, prev: n2,   next: null)
The problem comes when we try to replay the first add node datum.
Since its prev node is null, we rely on its next sibling n3. But
n3 was not present at this moment, and in previous code, we fallback
to append n4 to the last of its parent node.

The solution is to defer the append of elements that missing
siblings. But it is also hard to tell which node is the first one
that needs to be appended.

Take a step back and rethink the design of the mutation observer,
we've found there are two implementations make things complicated.
1. We set the id to -1 when we seeing some nodes are not serialized yet.
2. We record both previous sibling and next sibling to determine the
position of the node.

But we can do better! First, we can put nodes with un-serialized siblings
to a queue, and try to add it again later. Then we can just record the next
sibling as 'the single truth' so we can be sure which node is the last
one of its parent.

This patch has implemented the new observer strategy. Data recorded with
the new observer should no longer have any node with id -1. But for
compatibility consideration, we still keep some replayer code that helps
solve legacy data.

### DIFF
--- a/src/record/observer.ts
+++ b/src/record/observer.ts
@@ -191,17 +191,14 @@ function initMutationObserver(
     const addQueue: Node[] = [];
     const pushAdd = (n: Node) => {
       const parentId = mirror.getId((n.parentNode as Node) as INode);
-      if (parentId === -1) {
+      const nextId =
+        n.nextSibling && mirror.getId((n.nextSibling as unknown) as INode);
+      if (parentId === -1 || nextId === -1) {
         return addQueue.push(n);
       }
       adds.push({
         parentId,
-        previousId: !n.previousSibling
-          ? n.previousSibling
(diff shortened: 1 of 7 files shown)


Yuyz0112:
I think this patch is also related to:

#173 
#86 
#83

Issue #209, opened 2020-05-07
There is an error when moving to the next page


zhanseitaliev:
An error appears while playing back actions in the player.
When the visitor moves to the next page, the player stops, and an error appears in the console.

Error in console:  `Uncaught TypeError: t.setAttribute is not a function`

Events are sent to the server for storage every second.

Yuyz0112:
Close due to un-reproducible. Feel free to reopen if the problem still happens.

juancruzgardey:
@Yuyz0112 I have the same problem when I replay the interactions in the eBay website. I attach the events that I'm trying to replay. [events_ebay.zip](https://github.com/rrweb-io/rrweb/files/5034928/events_ebay.zip)

It's an error that use to happen in other websites also, especially in the highly dynamic ones.

I use the recorder as a web-extension, starting the capture whenever each page is loaded.

Yuyz0112:
@juancruzgardey Looks like you are using the chrome extension to record across multiple pages, but the DOM map are not reset correctly.

Could you share your record code for debugging?

juancruzgardey:
@Yuyz0112 sorry for the delay. I attach a zip that contains the web-extension.

Thanks!

[ScreenRecorderDebugging.zip](https://github.com/rrweb-io/rrweb/files/5226469/ScreenRecorderDebugging.zip)

PR #213, opened 2020-05-10
Verifying the type of node during mutation of attributes

zhanseitaliev:
To mutate attributes, a node must be of type Element. 
In my case, during the iteration of the attributes for the mutation, some nodes were of type Text. 
A node with type Text does not have a setAttribute method.

Issue: #209

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -499,6 +499,9 @@ export class Replayer {
           if (!target) {
             return this.warnNodeNotFound(d, mutation.id);
           }
+          if(target.nodeType !== 1) {
+            return this.warnNodeIsNotAnElement(d, mutation.id);
+          }
           for (const attributeName in mutation.attributes) {
             if (typeof attributeName === 'string') {
               const value = mutation.attributes[attributeName];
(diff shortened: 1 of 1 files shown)


Yuyz0112:
Hi @zhanseitaliev, thanks for the PR.

Could you provide some events for debugging? I think this should not happen because we only collect attributes changes on elements.

zhanseitaliev:
Hi @Yuyz0112, 

This error appears in some sites with different structures.
Demo player: [link](https://webvisor.shopter.tech/demo) (You can see the error after 5.30 s)
Events: [link](https://webvisor.shopter.tech/api/demo)

Thank you for your response!

Yuyz0112:
@zhanseitaliev Thanks for the help.

I see this is a multi-page website and there are some navigations. Are you putting rrweb on every page to trigger a record? Or maybe you can show some demo code of your record logic.

zhanseitaliev:
@Yuyz0112 Yes, I added rrweb to every page.

```
import { record } from 'rrweb';

let events = [];

record({
    emit(event) {
        events.push(event);
    },
});

function save() {

    if(events.length > 0) {
        const body = JSON.stringify({ events });
        events = [];
        fetch('https://visor.shopter.tech/api/sessions/', {
            method: 'POST',
            headers: {
                'Content-Type': 'application/json'
            },
            body,
        })
    }

}

setInterval(save, 5 * 1000);
```

Yuyz0112:
The events cause this problem is a little wired.

It looks like `https://charlottelowe.co.uk/collections/photo-jewellery` has more than one rrweb instance to record the page.

And I have tried using the REPL tool to record this page and things work well.
Is it possible to reproduce this now? (I see the events are logged on May 8)

Yuyz0112:
Close due to un-reproducible. Feel free to reopen if the problem still happens.

PR #214, opened 2020-05-14
This dimension is in px but is unitless 

eoghanmurray:
according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

didn't have to use `setAttribute`; can revert to direct attribute modification is that better suits project tastes.


Yuyz0112:
Thanks!

PR #222, opened 2020-05-26
Keep `npm run typings` happy

eoghanmurray:
Forgot to convert to string.

Incidentally, should `npm run typings` be part of the test suite?

There are a few unrelated errors when the typings command is run.


Yuyz0112:
Yep, now only the release process will run `npm run typings`

Issue #221, opened 2020-05-25
Throttling/pausing of Mutation Events


eoghanmurray:
I'm just wondering whether there has been any effort, or rejected effort, to add throttling and/or grouping to the MutationObserver.

I'm recording a website which has an image carousel which transitions images via a setInterval script, so every second, dozens of new mutation events are being generated without any user actions.
All the mutations are happening on the `style` attribute, so there's quite a lot of commonality between the events, which suggests a few different 'packing' strategies, but before delving into this level of complexity, I'm wondering whether an API method to pause mutation tracking would be useful?

This could be triggered by e.g. https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API  so that if a user isn't actually on the page, then mutation events will be batched up until they return.

eoghanmurray:
Have started some prep work for this in #223

Yuyz0112:
@eoghanmurray This will be one of the most exciting features for rrweb! I will start to review your patch today. Thanks a lot!

eoghanmurray:
Part 2 is in #224 which builds upon #223

eoghanmurray:
I've gone for adding a new API method `freezePage` (sitting alongside the only other one `addCustomEvent`).  Here's some sample code of how this could be used


    var pause_timeout = false;
        
    rrwebRecord({
        emit: function(event) {
            if (!(event.type == 3 && event.data.source == 0)) {
                if (pause_timeout) {
                    clearTimeout(pause_timeout);
                }
                pause_timeout = setTimeout(function(){
                    // stop recording page mutations after 10s of inactivity
                    rrwebRecord.freezePage();
                }, 10000);
            }
            
             ///  ... rest of emission code
        },
    });

    function handleVisibilityChange() {
        if (document.hidden) {
            rrwebRecord.addCustomEvent('window-hidden');
            rrwebRecord.freezePage();
        } else {
            rrwebRecord.addCustomEvent('window-visible');
        }
    }

    // Warn if the browser doesn't support addEventListener or the Page Visibility API
    if (typeof document.addEventListener === "undefined" || is_hidden_attr === undefined) {
        //pass
    } else {
        document.addEventListener('visibilitychange', handleVisibilityChange, false);
    }

eoghanmurray:
Let me know if you think the above behavior should be made into a config setting, e.g.

   `autofreezeOnBlur: boolean = false;`
   `autofreezeAfterSeconds: float = -1;`

[...]

PR #223, opened 2020-05-26
Move mutation processing into it's own class

eoghanmurray:
This should stand on it's own as a refactor, but is intended as a basis
for exposing the new MutationBuffer object to further outside control e.g.
to 'mute' or batch up mutation emission when the page becomes inactive
from a https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
point of view


eoghanmurray:
Not currently working; output is producing this:
```
MutationBuffer.prototype.processMutations = function (mutations) {
   ...
   var _this = this;
   mutations.forEach(this.processMutation);
```
instead of 
```
mutations.forEach(_this.processMutation);
```

eoghanmurray:
Ok, working now with the 05334ef152b365458d08e0376592031bdbb67661 commit.

eoghanmurray:
Have solved the bind problem by making all methods use the Arrow function expressions.
Let me know if there's a neater way of doing methods (with the ability to use `this`) in typescript.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

https://stackoverflow.com/questions/43724426/this-is-undefined-inside-the-foreach-loop

Yuyz0112:
@eoghanmurray The changes look good to me.

If I'm right, this patch is a 1-1 rewrite of the mutation processing code without any new logic and would be the base of the following changes.

Please let me know if I'm wrong. For example, if there are any logical changes to the processing flow and I would like to review that part again.

(I've already go through the patch, but Github's split view was not quite useful for comparing new files and deleted files)

eoghanmurray:
Yes, this `mutationClass` branch is purely refactoring, and should not have any logical changes.  Apologies for delay.

eoghanmurray:
#224 builds on top of this and has the logical changes discussed in #221 .

Yuyz0112:
@eoghanmurray Thanks for the feedback, I would love to merge this patch and start to review the next one.

PR #224, opened 2020-05-27
Enable external pausing of mutation buffer emissions

eoghanmurray:
- no automatic pausing based on e.g. pageVisibility yet, assuming such a thing is desirable
   https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
 - user code has to call new API method `freezePage` e.g. when page is hidden or after a timeout
 - automatically unpauses when the next user initiated event occurs
   (am assuming everything that isn't a mutation event counts as 'user initiated'
   either way think this is the correct thing to do until I see a counterexample
   of an event that shouldn't cause the mutations to be unbufferred)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,5 +11,6 @@ export {
 } from './types';
 
 const { addCustomEvent } = record;
+const { freezePage } = record;
 
-export { record, addCustomEvent, Replayer, mirror, utils };
+export { record, addCustomEvent, freezePage, Replayer, mirror, utils };
--- a/src/record/index.ts
+++ b/src/record/index.ts
@@ -1,5 +1,5 @@
 import { snapshot, MaskInputOptions } from 'rrweb-snapshot';
-import initObservers from './observer';
+import { initObservers, mutationBuffer } from './observer';
 import {
   mirror,
   on,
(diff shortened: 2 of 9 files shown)


eoghanmurray:
410756e5fda39bf88b5bca791ef998a1752690f2 makes a change to delay processing of the \`addedSet\` until emission.  I'm not sure I understand this all, but the change solves an issue with a build up of additions (when the DOM was just cycling through nodes).

eoghanmurray:
@Yuyz0112  I've rebased based on current master, would it be possible to get another review and merge if this change is okay?

This pull request adds a new API method `.freezePage` which allows mutations to be suspended externally.

We are using this to prevent pages which have e.g. a carousel implemented with DOM mutation from generating massive recordings even though the website visitor is not taking any actions (and might not even have the tab in focus)

eoghanmurray:
Is `suspendMutations` a better name than `freezePage`?

Yuyz0112:
For me, freezePage sounds more intuitive.

BTW, if I understand correctly, #385 contains all the changes in this PR and we can only review and discuss that one?

eoghanmurray:
Yes, #385 is this pull request and more ... all further discussion can go there.

PR #235, opened 2020-06-30
Expose startTime and endTime in the `getMetaData()` call.

eoghanmurray:
I was using `player.events[0].timestamp` but `player.events` has gone away (since a78da77 I believe)

I was doing things like `e.offset = e.timestamp - rrweb_replayer.events[0].timestamp;` in client code for constructing timelines.

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -155,6 +155,8 @@ export class Replayer {
     const firstEvent = events[0];
     const lastEvent = events[events.length - 1];
     return {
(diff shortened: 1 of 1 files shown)


eoghanmurray:
Ok, I see now that `.events` is now available in `.service.state.context.events`!
That is a lot of layers of indirection!  It would take me a long time to find that by poking around in the javascript debugger :(

Yuyz0112:
Thanks @eoghanmurray

PR #236, opened 2020-06-30
`.play()` backwards compatibility

eoghanmurray:
Ensure api method `.play()` can be called by external code without worrying about the internal state of the finite state machine

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -178,7 +178,14 @@ export class Replayer {
    * @param timeOffset number
    */
   public play(timeOffset = 0) {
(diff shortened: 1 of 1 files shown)


eoghanmurray:
I haven't added anything to handle the `skipping` state as I don't know enough about fast forward.

I'm not happy about this commit as it appears to just add more complexity to undo the complexity of the FSM.
Wondering whether this could instead be solved by adding e.g. a PLAY handler to the `ended` state etc.
Really I just don't know what the benefit of using a FSM so maybe an explanation on that will set me right!

Yuyz0112:
Looks reasonable. And we can further simply public API with this approach.

PR #246, opened 2020-07-15
Restore functioning of #200 - this got broken after #242

eoghanmurray:
- What was broken was that it would just play activity from the first page view, but then would stop at the second page view (meta) as actions after that had been discarded
 - This restores the functionality given by the comment 'return the events from last meta to the end.' - we never want to discard events that are after the baseline time
 - I believe 'session' is the incorrect terminology for this function name, as a session in web analytics usually means a series of page views


Yuyz0112:
@eoghanmurray Thanks! I think you are right, my bad.

PR #250, opened 2020-07-15
Reset the `lastPlayedEvent` as it would otherwise be used to discard events

eoghanmurray:
Reset the `lastPlayedEvent` as it would otherwise be used to discard events events in machine.ts as follows:

          for (const event of neededEvents) {
            if (
              lastPlayedEvent &&
              lastPlayedEvent.timestamp > baselineTime &&
              (event.timestamp <= lastPlayedEvent.timestamp ||
                event === lastPlayedEvent)
            ) {
              continue;
              }
             const isSync = event.timestamp < baselineTime;

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -217,6 +217,7 @@ export class Replayer {
    */
   public play(timeOffset = 0) {
     if (this.service.state.value === 'ended') {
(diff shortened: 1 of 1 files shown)


Yuyz0112:
Thanks

PR #264, opened 2020-07-23
Add support for invoking `play` when player is already playing

Juice10:
Looks like calling `play` a second time was broken. I re-implemented it and added some tests.

This now works:
```javascript
      const { Replayer } = rrweb;
      const replayer = new Replayer(events);
      replayer.play(500);
      replayer.play(1500);
```

## Related issues & pull requests

This might make https://github.com/rrweb-io/rrweb/pull/250 redundant, not completely sure.

It fixes some of the things mentioned in https://github.com/rrweb-io/rrweb/pull/198#issuecomment-658895230 including:
 > What if I am in the playing state and wish to jump forward 10s and continue playing

### DIFF
--- a/src/replay/machine.ts
+++ b/src/replay/machine.ts
@@ -20,60 +20,60 @@ export type PlayerContext = {
 };
 export type PlayerEvent =
   | {
-      type: 'PLAY';
-      payload: {
-        timeOffset: number;
-      };
-    }
+    type: 'PLAY';
+    payload: {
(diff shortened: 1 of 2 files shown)


Yuyz0112:
I think this has been implemented in https://github.com/rrweb-io/rrweb/issues/274, please let me know if it does not fit your use case.

Juice10:
Thanks for that @Yuyz0112! Closing this PR

PR #269, opened 2020-07-27
Add replayer.seek(timeOffset) to move the player to a moment in time

Juice10:
Builds on top of https://github.com/rrweb-io/rrweb/pull/264, please review that PR first.

Adds a `seek` method that pauses the player at a given time offset. It's similar to `replayer.play(2500); replayer.pause()` except that it won't trigger a race condition through `waitForStylesheetLoad` (https://github.com/rrweb-io/rrweb/issues/268).

## Example code
```javascript
const { Replayer } = rrweb;
const replayer = new Replayer(events);
replayer.seek(2500); //=> pauses the player at timeOffset 2500
```

Fixes https://github.com/rrweb-io/rrweb/issues/268

PR #288, opened 2020-08-19
Add regression tests and `pause(timeOffset)` test

Juice10:
Port over the tests from #269 and #264 as they can be useful as regression tests.

Includes tests for 
- `play()`-ing multiple times.
- `pause()` at time offset

Also increases timeout as puppeteer has a tendency to take longer than two seconds to start on my 2020 16inch MacBook Pro. (We should probably also do that for `record.test.ts` too)


Yuyz0112:
Thanks!

PR #294, opened 2020-08-20
Use window.top and window.top.document

eoghanmurray:
This pull request is to break out of simple same-domain <iframe>s, i.e. when rrweb.js is loaded from within the iframe.

This is the case for GoDaddy.

I'm not sure whether this is generally a good idea for all of rrweb's users, so maybe it needs further consideration before being accepted?

### DIFF
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -19,6 +19,7 @@ import {
   listenerHandler,
   mutationCallbackParam,
   scrollCallback,
+  IWindow,
 } from '../types';
 import { IframeManager } from './iframe-manager';
 import { ShadowDomManager } from './shadow-dom-manager';
@@ -451,7 +452,7 @@ function record<T = eventWithTime>(
             );
             init();
           },
-          twindow,
+          twindow as IWindow,
         ),
       );
     }
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -2,6 +2,7 @@ import { snapshot, MaskInputOptions, SlimDOMOptions } from 'rrweb-snapshot';
 import { initObservers, mutationBuffers } from './observer';
 import {
   on,
+  getTopWindow,
   getWindowWidth,
   getWindowHeight,
   polyfill,
@@ -66,6 +67,10 @@ function record<T = eventWithTime>(
   if (!emit) {
     throw new Error('emit function is required');
   }
+
+  const twindow = getTopWindow();
+  const tdoc = twindow.document;
+
   // move departed options to new options
   if (mousemoveWait !== undefined && sampling.mousemove === undefined) {
     sampling.mousemove = mousemoveWait;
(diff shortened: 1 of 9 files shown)


Juice10:
Great work eoghanmurray!
I’m wondering if this could potentially work cross-domain, for example in electron if rrweb is injected in all the iframes.

Juice10:
> I just came across this since reading your comment:
> [rrweb-io/rrweb-snapshot#10](https://github.com/rrweb-io/rrweb-snapshot/pull/10)
> Maybe that is unrelated?

@eoghanmurray I'm not sure to be honest.

> Do you have a chance to test out what would happen with multiple iframes in the same document?

Yes I'd be happy to, anything specific you'd like me to do? Just do a recording with this branch? Any specific pages you'd like me to try out?

One thing that might make electron based recording a little trickier is that every iframe/browser-window has its own `rrweb.record({ emit(event) { /* send to central location */  } });` so electron receives events from multiple pages at the same time without knowing which one is which.

eoghanmurray:
The electron (https://www.electronjs.org/ ?) application of rrweb is definitely not on my radar, I was only thinking of the web; so I guess a run through it's paces would be great to make sure this change doesn't break things there.

eoghanmurray:
Should this be a configuration option for it to be accepted?

eoghanmurray:
Does https://github.com/rrweb-io/rrweb/pull/481 change it so that it starts with the top level document, or is it just in whatever document that rrweb is initiated?

https://github.com/rrweb-io/rrweb/blob/1dd91f0737dc326560fec4bccb0264623777103f/src/record/index.ts#L439-L442

I think the latter is the case so maybe this pull request is still relevant?

eoghanmurray:
I've done a full rework as this PR was originally authored before we had nested iframe support (#481)

This is also based on #716, although that is still in progress, as they are related.

Code-Crash:
Hi @eoghanmurray ,

Do we have any small codepen or demo, or if I want to test it, how we can do that?

Thanks

Code-Crash:
> > Hi @eoghanmurray ,
> > Do we have any small codepen or demo, or if I want to test it, how we can do that?
> > Thanks
> 
> I build the rrweb locally and when I tried to use it with an iframe, it was working but scrolling/tab events are not getting captured.


Hi @eoghanmurray,

I tried to build and debug the issue and I found that the DOM tree is getting captured correctly (with scroll/tab also), so I think, the replayer is not able to replay it properly.

samber:
Any update regarding this PR?

How can we move things forward?

I wonder if it would be better to add `window` as a property of `recordOptions`? 🤔  It would let the developer pick the right window: `window` or `window.parent` or `window.top`.

eoghanmurray:
> I wonder if it would be better to add `window` as a property of `recordOptions`? thinking It would let the developer pick the right window: `window`, `window.parent`, `window.top` or child iframe.

I think both will be good, but I think it should default to the highest window it can find

ziyiz-amzn:
@Juice10 @eoghanmurray This is helpful, any plan to merge this CR? or #1023 ?

PR #305, opened 2020-08-25
Added option to be able to modify the dom during replay

lubenard:
Added an option in the config object of the replayer to be able to modify the dom during replay.
The option is named **modifiable** and it's default value is **false**

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -51,6 +51,7 @@ const defaultConfig: playerConfig = {
   liveMode: false,
   insertStyleRules: [],
   triggerFocus: true,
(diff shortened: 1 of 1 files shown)


Yuyz0112:
`enableInteract` and `disableInteract` are public APIs, why not just use them?

lubenard:
I thought it would be good idea, since this is config, to put it in the config too

Yuyz0112:
Okay, how about `intractable`? which is same as the API

lubenard:
My bad. You are right, that was a stupid PR. 
Sorry for the inconvenience.

I'm closing it.

Issue #285, opened 2020-08-16
seeking forward causing error


raminious:
Hey, 

I'm getting error when I'm trying to seek forward the progress   

https://github.com/rrweb-io/rrweb/blob/be4cccd6e373c1a6702130a21a1d4557f2944101/src/replay/index.ts#L627

![image](https://user-images.githubusercontent.com/334716/90345608-b0579700-e02a-11ea-9dfd-6471bf8dc64c.png)

steps of reproducing the error:   
https://streamable.com/7df7v7   

rrweb version: 0.9 (latest)

Here is the repository that I'm working on that. https://github.com/raminious/rrweb-player   

`git clone git@github.com:raminious/rrweb-player.git`    
`npm install`   
`npm start`    
open `http://localhost:1234`

you can find the events array in `src/response.js`

Juice10:
I'm having the same issue as @raminious.

## Steps to reproduce
1. Go to https://codepen.io/juice10/pen/LYNRypv
2. Open your browser's console/debugger.
3. Seek anywhere past 30 seconds.
4. Error is triggered in the console. `Cannot read property 'rules' of null`

## Example of offending event:
```json
{"86":{"data":{"id":105,"adds":[{"rule":".css-1qrqtax{height:0.25rem;background-color:#BEE8F2;opacity:0;-webkit-transition:opacity 0.5s;transition:opacity 0.5s;opacity:1;-webkit-transition:opacity 2.5s;transition:opacity 2.5s;-webkit-animation-name:animation-1hmp8dd;animation-name:animation-1hmp8dd;-webkit-animation-duration:2s;animation-duration:2s;-webkit-animation-timing-function:ease-in-out;animation-timing-function:ease-in-out;-webkit-animation-iteration-count:infinite;animation-iteration-count:infinite;}","index":75}],"source":8},"type":3,"timestamp":1597398354895}}
```

## Lines responsible for error
Line where error happens: https://github.com/rrweb-io/rrweb/blob/master/src/replay/index.ts#L627
Line where `styleSheet` is defined: https://github.com/rrweb-io/rrweb/blob/master/src/replay/index.ts#L620

## What I think is happening
From what I could find the problem happens on `IncrementalSource.StyleSheetRule` events where `styleEl.parentNode` is a DocumentFragment (not yet added to the dom). Apparently you can't call `.sheet` on stylesheet elements in a DocumentFragment.

Juice10:
[...]

PR #307, opened 2020-08-25
Child nodes without __sn now remove without error

Juice10:
When removing a `<style>` element the text nodes inside don't always have a `__sn` attribute which triggers an error in `deepRemoveFromMirror`. 

This PR fixes that and adds a test to prove that it is fixed.

Fix for https://streamable.com/wjnu23 (20 second mark) from https://github.com/rrweb-io/rrweb/issues/285#issuecomment-678808349.


Yuyz0112:
Thanks, could you resolve the conflict?

Juice10:
@Yuyz0112 my pleasure! Just resolved the conflict. Thank you!

PR #310, opened 2020-08-27
Fix live mode

Juice10:
Looks like live mode broke and it wasn't possible to go from paused (or playing) to live. This fixes that and adds a test to prove it.


Yuyz0112:
Thanks! pause -> live should be a valid operation.

Juice10:
Happy to help!

Issue #375, opened 2020-09-24
de-async emissions


eoghanmurray:
I've got a problem I've traced back to processing of the emission event.
Basically I think it's possible for a DOMMutation event to alter the DOM node tree while I'm processing the initial FullSnapshot event.

This is resulting in an application specific bug, but (assuming I've got the diagnosis right) I'm wondering whether this could be a more general source of bugs, e.g. a mutation event appearing in the recording before the snapshot that it should act upon.

Should we introduce locking code or some sort of a queue to ensure that emissions happen sequentially?

Yuyz0112:
I've seen similar problems in the checkout feature.

I've considered adding a general `generation` number to the events, which is the number of the full snapshot. So in the replay, we can check whether an event was happened before the current generation.

eoghanmurray:
I'm having a look at implementing a simple lock in the wrapEmit function so that the next emission can't happen until the client code deals with the previous emission. 
I'm using a queue, and basing my understanding of the js details on https://stackoverflow.com/a/2734311/6691

eoghanmurray:
This is turning out a little more complicated.

I think I have to delay execution of e.g. mutation.ts::emit() itself as it is the calling of this function that modifies the mirror

(my current queue/lock system is just delaying the processing of the emissions)

Any guidance appreciated!

eoghanmurray:
Ok trying to directly pause mutations while taking a full snapshot, building on the code added here:
https://github.com/rrweb-io/rrweb/pull/224
(that pull request hasn't been accepted yet)

eoghanmurray:
I have all tests passing in #385 and believe things are correct.

I have this deployed on 11 websites that were exhibiting the problem (a FullSnapshot already containing the results of DOM Mutation before the rrweb mutation event is emitted), and it solves the problem; no need for any further work to queue emissions as far as I'm concerned.

eoghanmurray:
So to wrap up: #385 (now merged) doesn't do any general queueing of emissions (to ensure that only one event can be processed at a time) but does address the specific interaction of FullSnapshot with DomMutation ... i.e. the latter is delayed until the former is completed.

PR #385, opened 2020-10-01
Suspend mutations during snapshot

eoghanmurray:
See #375 for the motivation for this.

This builds on top of #224

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,5 +11,6 @@ export {
 } from './types';
 
 const { addCustomEvent } = record;
+const { freezePage } = record;
 
-export { record, addCustomEvent, Replayer, mirror, utils };
+export { record, addCustomEvent, freezePage, Replayer, mirror, utils };
--- a/src/record/mutation.ts
+++ b/src/record/mutation.ts
@@ -296,7 +296,6 @@ export default class MutationBuffer {
     ) {
       return;
     }
-    this.emissionCallback(payload);
 
     // reset
     this.texts = [];
@@ -306,6 +305,8 @@ export default class MutationBuffer {
     this.movedSet = new Set<Node>();
     this.droppedSet = new Set<Node>();
     this.movedMap = {};
+
+    this.emissionCallback(payload);
   };
 
   private processMutation = (m: mutationRecord) => {
--- a/src/record/mutation.ts
+++ b/src/record/mutation.ts
@@ -114,6 +114,7 @@ export default class MutationBuffer {
   private texts: textCursor[] = [];
   private attributes: attributeCursor[] = [];
   private removes: removedNodeMutation[] = [];
+  private mapRemoves: Node[] = [];
 
   private movedMap: Record<string, true> = {};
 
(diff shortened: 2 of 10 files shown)


eoghanmurray:
Interesting side note: during development I discovered following change (which I subsequently reverted) is more correct:
https://github.com/rrweb-io/rrweb/pull/385/commits/2600fe737274a122ee171d489ea652803cce539c
basically we record attribute changes in a mutation when those changes are already applied in the 'adds' array of the same mutation.

Yuyz0112:
Cool, I will start to review  the PRs @eoghanmurray

src/record/mutation.ts
```diff
@@ -143,7 +145,7 @@ export default class MutationBuffer {
   private maskInputOptions: MaskInputOptions;
   private recordCanvas: boolean;
 
-  constructor(
+  public init(
```
Yuyz0112, line 148: What's the point of using an init method instead of the constructor?

src/record/mutation.ts
```diff
@@ -109,10 +109,12 @@ function isINode(n: Node | INode): n is INode {
  * controls behaviour of a MutationObserver
  */
 export default class MutationBuffer {
+  public paused: boolean = false;
```
Yuyz0112, line 112: Maybe use private property and provide two public methods to control it?

src/index.ts
```diff
@@ -11,5 +11,6 @@ export {
 } from './types';
 
 const { addCustomEvent } = record;
+const { freezePage } = record;
 
-export { record, addCustomEvent, Replayer, mirror, utils };
+export { record, addCustomEvent, freezePage, Replayer, mirror, utils };
```
Yuyz0112, line 16: What's the use case a user should call freeze page manually?

src/record/mutation.ts
eoghanmurray: It's because we are now exporting a singleton `mutationBuffer` object from observer.ts, for importation into record/index.ts.
This is the only way I knew to create the global object there; `initMutationObserver` subsequently gets the config parameters and uses them to initialize the mutationBuffer via the `init` method. 
Maybe there is a better way?

src/index.ts
eoghanmurray: The usecase for Statcounter is that some pages have a constant stream of mutations happening all the time (an image carousel where transitions are controlled by javascript rather than CSS).   These generate huge recordings.  We've mitigated this by deciding that if there is no user activity on the page, we call `freezePage` after a timeout.  

A second thing we do is to call freezePage when the page loses focus (`document.addEventListener('visibilitychange')`).

So I've gone with implementing this as a function call, rather than as a setting (`autoFreezeAfterSeconds` or similar), as there may be other use cases I haven't thought of, and I felt like there would need to be two config settings for the above two scenarios, which felt like a harder thing to document.

See https://statcounter.com/counter/recorder_uncompressed.js for details on how the two calls are set up; these could be added to documentation as examples if we're gonna keep the `freezePage` method approach.

src/record/mutation.ts
eoghanmurray: Thanks, in 1006d8efeccd98e4bafdaa600f3498b3aa12e929 I've added getter/setter methods and renamed from `paused` to `frozen` for consistency, and to avoid any confusion with pausing playback.

src/index.ts
Yuyz0112: I cannot open the about link but I do understand the use case now. Thanks!

Yuyz0112 (approved):
LGTM

PR #389, opened 2020-10-09
Make destructor/stopper function explicit and add option to clean up DOM

eoghanmurray:
depends on https://github.com/rrweb-io/rrweb-snapshot/pull/49


Juice10 (approved):
LGTM

src/record/index.ts
```diff
@@ -389,9 +390,17 @@ function record<T = eventWithTime>(
...
       );
     }
-    return () => {
+    stop = function(cleanup: boolean) {
       handlers.forEach((h) => h());
+      if (cleanup) {
+        let id: keyof typeof mirror.map;
+        for (id in mirror.map) {
+          delete mirror.map[id].__sn;
+        }
```
Juice10, line 399: The mirror was refactored so this doesn't work anymore

PR #407, opened 2020-10-29
feature: make mirror independent in Replayer

ZzqiZQute:
I found that the `Replayer` of rrweb is now using only one `mirror` object to rebuild the snapshot. But when there are many replayer objects, the mirror with drop the DOM information of the previous replayer. This pull request wants to save the mirror inside of the replayer object itself, so that we can use many replayer object and they don't be conflict with each other.


eoghanmurray:
I was wondering why switching between different replayers seems to take a long time, even when the recording has been played already.  Maybe this is the problem?

provinceinnovation:
This PR looks good to me. I am also using multiple `Replayer`'s in parallel and find that they conflict with each other due to the single `mirror` object. I am aware that removing the global `mirror` object is a breaking change. @Yuyz0112 - how do you feel about this PR?

Yuyz0112:
@provinceinnovation I also love this PR. Since it's an API breaking change, I marked it as a v1.0 candidate.

src/replay/index.ts
```diff
@@ -178,10 +180,27 @@ export class Replayer {
       for (const d of inputMap.values()) {
         this.applyInput(d);
       }
+
+      for (const [frag, parent] of this.fragmentParentMap.entries()) {
```
Yuyz0112, line 184: This is an issue introduced when resolving conflicts.

I'll remove this in the master branch.

PR #409, opened 2020-11-03
Add record option maskInputFn for custom mask input function

yaozulv:
sample code
```
rrweb.record({
  emit(event) {
    // store the event in any way you like
  },
  maskAllInputs: true,
  maskInputFn: text => 'x'.repeat(text.length),
});
```


Yuyz0112:
Thanks!

PR #410, opened 2020-11-03
feat: enable to change the config of mouse tail while playing

Mark-Fenng:
fix a potential bug: handleResize function might change the display style of mouseTail


Yuyz0112:
Thanks!

PR #411, opened 2020-11-04
Sort events upon creation, and keep ordered

eoghanmurray:
Otherwise we risk misidentifying the last event (i.e. calling `finish()` prematurely).

I noticed this when creating a recording which had a custom event at the end of the event list which was coming from a different source, i.e. not generated by rrweb, and had a timestamp earlier in the recording.

This event was getting executed (doAction) at the correct timestamp, but after execution it was identified as the 'last' event, and no further events were being executed.

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -125,6 +125,8 @@ export class Replayer {
     events: Array<eventWithTime | string>,
     config?: Partial<playerConfig>,
   ) {
+    events.sort((a1, a2) => a1.timestamp - a2.timestamp);
+
     if (!config?.liveMode && events.length < 2) {
       throw new Error('Replayer need at least 2 events.');
     }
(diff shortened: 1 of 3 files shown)


eoghanmurray:
I missed pushing the principal commit for this one; apologies if you had already reviewed it as it didn't make much sense.

eoghanmurray:
Recent commits:
- Ensure that ordering is stable. Previously we were inadvertently reversing order of events when there happened to be multiple events with the same timestamp
- Fast track: Don't enter into binary insertion search if the new event has a timestamp greater than the last event

Yuyz0112:
Thanks, @eoghanmurray.

Now I think implemented sorting inside rrweb is a good idea to avoid events that were not sorted in the userland. And I believe a binary search + splice is the only algorithm we can have with the current data structure.

lele0108:
@eoghanmurray @Yuyz0112 

With this PR, in live mode, I don't think we are taking into account the insertion of the first event:

`let end = events.length - 1;` will return -1 if `events.length` is 0. 

Which will cause `if (end === -1 || events[end].timestamp <= event.timestamp) {` to break. Has this behavior been tested?

Thanks!

Yuyz0112:
@lele0108 Thanks for the report and you are right! I've just pushed the fix.

Issue #216, opened 2020-05-15
Render Frame 0


jwbrew:
When initalising a new Replayer, if `play()` is not called, there is just a blank screen with a cursor position available. Should it be possible to show the first frame of the recording?

Yuyz0112:
Good idea, going to add this

Yuyz0112:
Good idea, going to add this

eoghanmurray:
One problem with this is that you don't get a chance to attach a handler for the `fullsnapshot-rebuilded` event.

I see that there is special `setTimeout` treatment for the `Resize` event, but maybe that's for a different purpose?

PR #416, opened 2020-11-11
allow fullsnapshot capture to be added along with constructor

eoghanmurray:
Allow `.on` to be chained directly after the constructor so that event handling for the first full snapshot can be added straight after constructor, and that handler will receive the first automatic fullsnapshot build (#216)

Could do with some documentation so you can see:

```
	var rrweb_replayer = new rrwebReplay.Replayer(
		recording
	).on('fullsnapshot-rebuilded', function(fullSnapshot) {
		// do something with first build
	});
```

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -194,14 +194,17 @@ export class Replayer {
       }, 0);
     }
     if (firstFullsnapshot) {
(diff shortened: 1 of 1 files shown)


Yuyz0112:
LGTM

Thanks

PR #417, opened 2020-11-11
Ensure animations don't continue playing when we've paused playback

eoghanmurray:
Further work:
The animation state won't follow scrubbing/jumping of playback position, but I don't know any way of achieving that.

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -276,6 +276,7 @@ export class Replayer {
       this.service.send({ type: 'PAUSE' });
       this.service.send({ type: 'PLAY', payload: { timeOffset } });
     }
+    this.iframe.contentDocument.getElementsByTagName('html')[0].classList.remove('rrweb-paused');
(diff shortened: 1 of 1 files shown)


Yuyz0112:
LGTM.

And I think this is a try-my-best strategy right @eoghanmurray ? Because there may be some JS animation we could not pause. And basically we are out of control with the animation state.

eoghanmurray:
Yes, it's just handles the a specific CSS animation that I noticed in a recording.

Any JS animations should be paused naturally as they are being rendered via mutations though right?  
Maybe you are thinking of GIF animations?  
If so ya that would be under 'further work' :relaxed:  e.g. via https://github.com/ctrl-freaks/freezeframe.js (which has a typescript implementation)

Yuyz0112:
Thanks!

Mark-Fenng:
@eoghanmurray In my application, I need the CSS transition finished after I paused play. Could we add an option for this?

Issue #317, opened 2020-09-02
Tracking console.error and network requests.


jay-khatri:
Tools like fullstory and logrocket also log the output of the console and the outgoing requests from the browser. Is this something that's being considered at all?

Yuyz0112:
They are nice to have, but maybe not a high priority thing for myself at this moment. Would love to help if anyone wants to implement these.

rifaideen:
@jay-khatri you can implement on your own something like this:

```js
    /**
     * Intercept http requests and track excluding calls to `host` variable.
     */
    const xhrOpen = window.XMLHttpRequest.prototype.open;

    // eslint-disable-next-line no-unused-vars
    window.XMLHttpRequest.prototype.open = function (method, url, async, user, password) {
        const request = {
            url,
            method,
            async,
            headers: {},
        };

        // parse headers
        const getHeaders = (rawHeaders) => {
        const result = {};
        const headers = rawHeaders.trim().split(/[\r\n]+/);
            headers.forEach((line) => {
                const parts = line.split(': ');
                const header = parts.shift();
                const value = parts.join(': ');
                result[header] = value;
            });

            return result;
        };

        // eslint-disable-next-line func-names
        this.addEventListener('load', function () {
            request.headers = getHeaders(this.getAllResponseHeaders());

            const payload = this.responseText;
            request.status = this.status;
            request.status_text = this.statusText;
            request.response_type = 'text/plain';

            if (request.headers['content-type']) {
                request.response_type = request.headers['content-type'];
            } else if (request.headers['Content-Type']) {
                request.response_type = request.headers['Content-Type'];
            }

            if (payload) {
                request.response = new Blob([payload], { type: request.response_type });
            }

            /**
             * do something with collected information. (request)
             */
        });

    
        // eslint-disable-next-line prefer-rest-params
        xhrOpen.apply(this, arguments);
    };
  
  /**
   * Listen for console events.
   */
  // eslint-disable-next-line no-console
  const cerror = console.error;
  // eslint-disable-next-line no-console
  console.error = function () {
    const event = { type: 'error', values: arguments };

    /**
     * do something with event.
     */

    // eslint-disable-next-line prefer-rest-params
    cerror.apply(this, arguments);
  };
```

Yuyz0112:
@Mark-Fenng has a PR for the console part: #424

Yuyz0112:
Let's discuss network tracking in https://github.com/rrweb-io/rrweb/issues/552

Issue #234, opened 2020-06-29
hijack the console API and record corresponding events


Luiz-N:
i'm looking to be able to capture events in the console while using rrweb. Specifically error events so this feature would be perfect. Any suggestions or examples I could look at for "hijacking the console" so that i could give this a try?

Yuyz0112:
Hi @Luiz-N 

I would suggest dividing the development of this feature into two parts:
1. capture console calls as rrweb events
2. display rrweb console events during replaying

For the first part, you can check some existing implementation like [this](https://github.com/getsentry/sentry-javascript/blob/a257a8aa1b46b8f3f3a48eb4d1b4bf89f9893095/packages/integrations/src/captureconsole.ts).

And for the second part, I'm not quite sure what's the best way to do it. I mean maybe we can call the console API during replaying to show the events we captured, or we can display them in part of the replayer. Any suggestions are welcomed and it should not block the first part.

alexcroox:
This is how LogRocket handles it, showing both the console output and network calls in sync with the playback video.

It's incredibly handy for debugging errors while watching the playback

![image](https://user-images.githubusercontent.com/1065835/87186258-19891380-c2e3-11ea-9590-8659877669c4.jpg)

Yuyz0112:
https://github.com/rrweb-io/rrweb/blob/master/docs/recipes/console.md

alexcroox:
Amazing work thank you!!

PR #424, opened 2020-11-22
feat: enable rrweb to record and replay log messages in console

Mark-Fenng:
this is the implementation of new feature request(issue #234 )(issue #317 )

here are a few points of description.
1. users need to set recordLog option in rrweb.record's parameter to record log messages.  The log recorder is off by default.
2. support recording and replaying all kinds of console functions. But the reliability of them should be tested more
3. the stringify function in  stringify.ts needs improvement. e.g. robustness and better support for more kinds of object
4. we can replay the log messages in a simulated html console like LogRocket by implementing the interface "ReplayLogger" in the future


Mark-Fenng:
@Yuyz0112 Could you please review this pull request😂  I've solved merge conflicts two times in the past one week

src/types.ts
```diff
@@ -353,8 +369,84 @@ export type fontParam = {
...
+  table?: (tabularData: any, properties?: ReadonlyArray<string>) => void;
+  time?: (label?: string) => void;
+  timeEnd?: (label?: string) => void;
+  timeLog?: (label?: string, ...data: any[]) => void;
+  trace?: (message?: any, ...optionalParams: any[]) => void;
+  warn?: (message?: any, ...optionalParams: any[]) => void;
+};
+
+// functions to replay log record
+export type ReplayLogger = {
```
Yuyz0112, line 418:
```ts
export type ReplayLogger = Record<LogLevel, (data: logData) => void>
```

Is this an equivalent?

Yuyz0112:
LGTM

Most of the design looks great and just a few things need to be tweaked.

src/types.ts
Mark-Fenng: Yes, this one is better!

eoghanmurray:
There is no easy way to transmit events for when an unhandled exception happens on the page?  
I.e. if we don't have a chance to catch and log with `console.error` because the exception happens in 3rd party js code?

PR #426, opened 2020-11-26
fix: applyScroll function doesn't work in sync mode(play at a specific time)

Mark-Fenng:
applyScroll function should be called after all of new nodes mounted on dom


Yuyz0112:
Thanks!

PR #428, opened 2020-11-27
feat: add an option to determine whether to pause CSS animation when playback is paused

Mark-Fenng:
set pauseAnimation to true by default


Yuyz0112:
Thanks!

PR #435, opened 2020-12-07
docs: add document for console recording and playback

Mark-Fenng:


docs/recipes/console.md
```diff
@@ -0,0 +1,68 @@
...
+});
+replayer.play();
+```
+
+Description of replay option is as follows:
+
+| key          | default                                                                                      | description                                                                                                                                            |
+| ------------ | -------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| level        | ['log','warn','error',...]                                                                   | Same meaning as that option in recordLog. You can set this option to play levels of log you need.                                                      |
+| replayLogger | a console based object that implements the interface [ReplayLogger](../../src/types.ts#L417) | You can also set a replay logger to replay the log messages in a simulated browser console like LogRocket by implementing the interface `ReplayLogger` |
```
Yuyz0112, line 68: I think we'd better not to naming some commercial products in our docs.

Mark-Fenng: ok

PR #437, opened 2020-12-10
Catch SecurityErrors when accessing styleSheet rules

grziwok:
Hello! Thanks for creating this great library!

I'd like to cover one case which happened to us today. When we access a stylesheet rules, it may happen that it throws security error:

```
Uncaught DOMException: Failed to read the 'rules' property from 'CSSStyleSheet'
```

It occurs when we replay a page, which includes an iframe with link to stylesheet that has CORS security rules that we cannot access. I just added a try / catch to handle that because it throws an exception and stops recording.

Please let me know if that makes sense for you. Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode
+.idea
 node_modules
 package-lock.json
 # yarn.lock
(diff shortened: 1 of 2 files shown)


Yuyz0112:
Thanks! Going to merge it.

PR #445, opened 2020-12-18
fix: sometimes currentTime is smaller than the totalTime when player is finished

Mark-Fenng:
fix the problem that sometimes return value of getCurrentTime() is negative

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -508,7 +508,7 @@ export class Replayer {
           // defer finish event if the last event is a mouse move
           setTimeout(() => {
             finish();
-          }, Math.max(0, -event.data.positions[0].timeOffset));
+          }, Math.max(0, -event.data.positions[0].timeOffset + 50)); // Add 50 to make sure the timer would check the last mousemove event. Otherwise, the timer may be stopped by the service before checking the last event.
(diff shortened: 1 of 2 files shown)


src/replay/index.ts
```diff
@@ -508,7 +508,7 @@ export class Replayer {
           // defer finish event if the last event is a mouse move
           setTimeout(() => {
             finish();
-          }, Math.max(0, -event.data.positions[0].timeOffset));
+          }, Math.max(0, -event.data.positions[0].timeOffset + 50)); // Add 50 to make sure the timer would check the last mousemove event. Otherwise, the timer may be stopped by the service before checking the last event.
         } else {
           finish();
```
eoghanmurray, line 513: @Mark-Fenng do you think the 50ms buffer is also valid to be applied to the `} else {` branch here?

I've another unrelated use-case; namely to ensure that we don't prematurely finish when new events are being rapidly added via `addEvent` ... see #606 — I'm thinking to add the `setTimeout ... 50` to before both calls to `finish()`.

Mark-Fenng: @eoghanmurray I think setTimeout at else branch can work out.

Issue #381, opened 2020-09-29
Consider alternative compression library 


alexcroox:
Initially when Pako was introduced to this library there were concerns over its huge size. Thought I’d throw a tiny, more performant alternative into the mix for future consideration:

https://github.com/101arrowz/fflate

> fflate (short for fast flate) is the fastest, smallest, and most versatile pure JavaScript compression and decompression library in existence, handily beating pako, tiny-inflate, and UZIP.js in performance benchmarks while being multiple times more lightweight.

Yuyz0112:
Upvote this suggestion!

The only thing I'm thinking about is how to handle compatibility problems since there are already some data compressed by pako.

alexcroox:
I’ve not looked at the data before, do you store a version number/date or similar against the compressed data?

Yuyz0112:
@alexcroox Yes we have a version number in the data, but what's the expected behavior when we provide a new packer via `fflate`, and meet data compressed by `pako`? Should we throw an error to let users use the v1 packer?

101arrowz:
Hi, library author here! As mentioned in the summary, data compressed by `fflate` can be decompressed by `pako` and vice versa. This wouldn't be a breaking change for that reason.

alexcroox:
Thanks for taking the time to let us know @101arrowz!!

101arrowz:
I've looked into the codebase and I think this will work and cut down bundle size while improving performance by a bit.
```patch
 package.json                     |  2 +-
 src/declarations/pako/index.d.ts | 12 ------------
 src/packer/pack.ts               |  4 ++--
 src/packer/unpack.ts             |  4 ++--
 4 files changed, 5 insertions(+), 17 deletions(-)
 delete mode 100644 src/declarations/pako/index.d.ts

diff --git a/package.json b/package.json
index 5e718a2..3c6a682 100644
--- a/package.json
+++ b/package.json
@@ -60,8 +60,8 @@
   "dependencies": {
     "@types/css-font-loading-module": "0.0.4",
     "@xstate/fsm": "^1.4.0",
+    "fflate": "^0.4.2",
     "mitt": "^1.1.3",
-    "pako": "^1.0.11",
     "rrweb-snapshot": "^1.0.1"
   }
 }
diff --git a/src/declarations/pako/index.d.ts b/src/declarations/pako/index.d.ts
deleted file mode 100644
index e92682a..0000000
--- a/src/declarations/pako/index.d.ts
[...]

PR #448, opened 2020-12-21
Update to fflate

101arrowz:
As mentioned in #381, fflate tends to perform better than Pako while adding less to the bundle size. This change will not break existing packed data; I used the same format as previously for full interop with older versions, meaning there is no need for a major version bump. However, it should improve performance significantly by reducing download time and runtime for both the player and the recorder. Tests all still pass without updating the snapshots.

**rrweb-record-pack.min.js**
Before (with Pako): 30089B = 30kB
After (with fflate): 9753B = 9.8kB

**rrweb-replay-unpack.min.js**
Before: 73037B = 73kB
After: 55037B = 55kB


Yuyz0112:
@101arrowz This is so awesome! Thanks a lot and love your work on fflate.

PR #451, opened 2020-12-22
Scroll replayer iframe on firstFullsnapshot

armata007:
Currently when starting recording when page is already scrolled down the replayer iframe content won't automatically go down. It is caused by not using scrollTo when first rebuilding full snapshot. 

This PR should fix it


Yuyz0112:
Thanks!

PR #454, opened 2021-01-04
Protect against generation of no-change viewport resize events

eoghanmurray:
I noticed 8 or 10 of these events being generated in a multi-tab browsing session on Chrome 87.0 on Win10.  I'm speculating they were generated as a side effect of changing tabs but I can't recreate.


Yuyz0112:
Looks reasonable. Thanks!

PR #464, opened 2021-01-18
Compact style mutation

eoghanmurray:
I had one recording in the wild that turned into a monster at 12GB.
One of the sources was continuous DOM Mutations caused by a 3rd party 'snow animation' library that was animating individual snowflakes multiple times per second via JavaScript, causing a constant stream of very large mutations.

Each mutation was of the format
` "style":"color: rgb(255, 255, 255); position: absolute; width: 8px; height: 8px; font-family: arial, verdana; overflow: hidden; font-weight: normal; z-index: 0; display: block; bottom: auto; opacity: 1; padding: 0px; margin: 0px; font-size: 10px; line-height: 10px; text-align: center; vertical-align: baseline; left: 242.807px; top: 85.7332px;"`

Even though just the left or top style attributes had changed.
The new format for this part of the attribute mutation looks like:
` "style": {"left": "242.807px", "top": "85.7332px"}`

~There's a threshold to the attribute length before this change applies, currently set at a style attribute length of 25 characters (this value could be tweaked)~ (the threshold would have caused errors, so I've removed it)

Diffs are generated by creating a dummy node to house the old style and then comparing each style (sub)attribute in turn using [getPropertyValue](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyValue) and [getPropertyPriority](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyPriority). Attributes on the old value are also iterated over to see if any of them have been removed.

At playback time, these sub-style mutations are applied via [setProperty](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty) with either a single value argument, or 2 arguments in the case that the `!important` priority flag is present in the change.

### DIFF
--- a/src/record/mutation.ts
+++ b/src/record/mutation.ts
@@ -449,6 +449,7 @@ export default class MutationBuffer {
         break;
       }
       case 'attributes': {
+        const target = (m.target as HTMLElement);
         let value = (m.target as HTMLElement).getAttribute(m.attributeName!);
         if (m.attributeName === 'value') {
           value = maskInputValue({
@@ -472,13 +473,41 @@ export default class MutationBuffer {
           };
           this.attributes.push(item);
         }
-        // overwrite attribute if the mutations was triggered in same time
-        item.attributes[m.attributeName!] = transformAttribute(
-          this.doc,
-          (m.target as HTMLElement).tagName,
-          m.attributeName!,
-          value!,
-        );
+        if (m.attributeName === 'style') {
+          const old = this.doc.createElement('span');
+          old.setAttribute('style', m.oldValue);
+          if (item.attributes['style'] === undefined) {
+            item.attributes['style'] = {};
+          }
+          for (let i=0; i<target.style.length; i++) {
+            let pname = target.style[i];
+            if (target.style.getPropertyValue(pname) !=
+                old.style.getPropertyValue(pname) ||
+                target.style.getPropertyPriority(pname) !=
+                old.style.getPropertyPriority(pname)) {
+              item.attributes['style'][pname] = [
+                target.style.getPropertyValue(pname),
+                target.style.getPropertyPriority(pname)
+              ];
+            }
+          }
+          for (let i=0; i<old.style.length; i++) {
+            let pname = old.style[i];
+            if (target.style.getPropertyValue(pname) === '' ||
+                !target.style.getPropertyValue(pname)  // covering potential non-standard browsers
+               ) {
+              item.attributes['style'][pname] = false;  // delete
+            }
+          }
+        } else {
+          // overwrite attribute if the mutations was triggered in same time
+          item.attributes[m.attributeName!] = transformAttribute(
+            this.doc,
+            (m.target as HTMLElement).tagName,
+            m.attributeName!,
+            value!,
+          );
+        }
         break;
       }
       case 'childList': {
--- a/src/record/mutation.ts
+++ b/src/record/mutation.ts
@@ -481,14 +481,15 @@ export default class MutationBuffer {
           }
           for (let i=0; i<target.style.length; i++) {
             let pname = target.style[i];
-            if (target.style.getPropertyValue(pname) !=
-                old.style.getPropertyValue(pname) ||
-                target.style.getPropertyPriority(pname) !=
-                old.style.getPropertyPriority(pname)) {
-              item.attributes['style'][pname] = [
-                target.style.getPropertyValue(pname),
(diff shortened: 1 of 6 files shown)


eoghanmurray:
(Sorry still iterating on this as tests are not passing; my dev test environment isn't working so uploading changes here to check)

eoghanmurray:
@Yuyz0112 would you be able to take a look at the failing build e.g. https://travis-ci.com/github/rrweb-io/rrweb/builds/213675796

There are some new style changes showing up after this PR, but as far as I can see, those should be present in the current https://github.com/rrweb-io/rrweb/blob/master/test/__snapshots__/integration.test.ts.snap

i.e. there should be a text attribute change as follows:
`"style": ""left: 8px; width: 46.2109px; top: 115.5px; bottom: auto; display: block;""`

I'm seeing that when I manually explore the select2.html page in Chromium also — that change is in the output.

eoghanmurray:
Okay, I've got all the tests sorted now and have added one extra test for `!important` as this is treated separately.

Aside: during replay, it won't break if a previous string value for the style attribute is present from a prior recording.

src/record/mutation.ts
```diff
@@ -373,12 +390,41 @@ export default class MutationBuffer {
...
+        if (m.attributeName === 'style') {
+          const old = document.createElement('span');
+          old.setAttribute('style', m.oldValue);
+          if (item.attributes['style'] === undefined) {
+            item.attributes['style'] = {};
+          }
+          for (let i=0; i<target.style.length; i++) {
+            let pname = target.style[i];
+	    const new_value = target.style.getPropertyValue(pname);
+	    const new_priority = target.style.getPropertyPriority(pname);
```
Juice10, line 402: For code hygiene it's probably better to camelCase these variable names. 
```suggestion
	    const newValue = target.style.getPropertyValue(pname);
	    const newPriority = target.style.getPropertyPriority(pname);
```

eoghanmurray: Thanks; I take it I cannot just commit this suggestion as-is?  
I'll make a new commit.

eoghanmurray: Updated the original commit with your suggested change as was force-pushing anyhow.

Juice10:
Looks good to me, only thing I can think of would be doing some checking in the replay, if the element has a style attribute. Just in case something went wrong with the recorded ids and we are accidentally referencing a text node. I believe there is a try/catch around the get/setAttributes for that reason on the string type of style mutation.

Yuyz0112:
@eoghanmurray Looks like this approach will not trigger reflow/repaint during diff, because the dummy node is not in the DOM tree, am I right?

And is it worth a little benchmark test to show the time of diff?

eoghanmurray:
Ya not added to the DOM; I felt safe creating the dummy node as we are doing that elsewhere in the code e.g. at https://github.com/rrweb-io/rrweb-snapshot/blob/master/src/snapshot.ts#L204

I made a [quick performance.now() comparison](https://gist.github.com/eoghanmurray/9e2bca1d589644ffb6610ddca7e3716e) and got the following results

Firefox (over 277 events):
Old avg. 0.184ms
New avg 0.234ms
Slower 1.27x

Chrome (over 256 events):
Old avg 0.073ms
New avg 0.224ms
Slower 3.03x 

Looks like the difference is due to the existing transformAttribute code path being just as slow in Firefox. 

This is happening in the DOMMutation callback which is 'off the main loop', right?

Yuyz0112:
LGTM

@eoghanmurray This happens off the main loop, but still executing in sync mode, so it may block following execution. I think we can:
1. support running record logic in web worker
2. do record logic in request idle callback

eoghanmurray:
for #1. I thought web workers didn't have access to the DOM?
#2. looks good if supported

razor-x:
Hello. I'm coming here as a [PostHog ](https://github.com/PostHog) user. It uses this lib to power session recording. My app is able to run a strict CSP, specifically without `'unsafe-inline'` in the policy. However, I noticed a lot of CSP errors generated like the one below:

image

After some digging, I believe this change broke support for users who run a strict CSP. Setting the `style` attribute on the dummy node will be blocked without `'unsafe-inline'`.

Would you prefer I start a new issue or continue discussion here on possible solutions? Best.

Juice10:
@razor-x thanks for bringing that issue up!
It probably makes most sense to start a new issue for that.

PR #466, opened 2021-01-20
Tweaks to timings to get tests passing on my dev laptop

eoghanmurray:
Hopefully this makes tests more deterministic...

This pull request is more to get the conversation started on how variations between test runs should be managed.

In particular, I've added the following comment:

https://github.com/rrweb-io/rrweb/blob/10842b4be98c8bea7c5cf44e042b452069a0b3ae/test/record.test.ts#L143

I'm not sure if the above change is in fact correct; but if not, it would be good to find out what the reason for waiting 500ms here is?

### DIFF
--- a/test/record.test.ts
+++ b/test/record.test.ts
@@ -33,6 +33,8 @@ interface IWindow extends Window {
 }
 
 describe('record', function (this: ISuite) {
+  this.timeout(10_000);
+
   before(async () => {
     this.browser = await launchPuppeteer();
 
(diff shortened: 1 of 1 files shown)


eoghanmurray:
Ok so the upshot of this pull request is that I've shifted the `event.length==33` check back by 200ms to ensure that a slow test environment doesn't result in the checkout already happening at that check.

Are there any other approaches we could take, e.g. instead of sleeping for 500ms, sleep for (500ms - time_already_expended)?

PR #470, opened 2021-01-20
fix for 'suspendMutationsDuringSnapshot' #385

eoghanmurray:
Discovered that the common case of mouse movement or scrolling happening during `takeFullSnapshot` was causing mutations to be immediately emitted, contrary to the goal of https://github.com/rrweb-io/rrweb/pull/385

Instead create two different types of 'lock' for mutations, one to do with freezes initiated by inactivity (which can be unlocked by any new non-mutation activity) and another lock during fullSnapshot, which can't be unlocked until the fullSnapshot is finished.


Yuyz0112:
Thanks

PR #479, opened 2021-02-04
Fix RangeError: Maximum call stack size exceeded

macobo:
Saw this line cause issues in production, causing the following error:

```
RangeError Maximum call stack size exceeded
```

I believe this is caused by javascript engine max argument length - see note from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply#using_apply_and_built-in_functions

> The consequences of applying a function with too many arguments (that is, more than tens of thousands of arguments) varies across engines. (The JavaScriptCore engine has hard-coded argument limit of 65536.

### DIFF
--- a/src/replay/timer.ts
+++ b/src/replay/timer.ts
@@ -30,7 +30,7 @@ export class Timer {
    * @param actions
    */
   public addActions(actions: actionWithDelay[]) {
-    this.actions.push(...actions);
+    this.actions = this.actions.concat(actions);
   }
 
   public start() {


src/replay/timer.ts
```diff
@@ -30,7 +30,7 @@ export class Timer {
    * @param actions
    */
   public addActions(actions: actionWithDelay[]) {
-    this.actions.push(...actions);
+    this.actions = this.actions.concat(actions);
```
Yuyz0112, line 33: So what's the difference between these two ways?

macobo: Say argument `actions = [1,2,3,4,5,6]`.

The new version is classic concat - creates new array consisting of the original + new.

The original version is equivalent to `this.actions.push(1,2,3,4,5,6)` - each element in the array becomes another argument to the function. This can cause `RangeError` when the actions array is very large - I haven't tested it but https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply#using_apply_and_built-in_functions mentions the limit might be 65k, or it might be lower in this case. In any case, unpacking very long arrays as function arguments can cause problems hence the PR.

Yuyz0112: Thanks! Sounds reasonable.

chillyistkult:
Looks like this issue is back in 1.0.1 (or never fixed?).

PR #487, opened 2021-02-13
Add prettier as a dependency

M-Izadmehr:
I noticed that the prettier is not included as a dev dependency, so when you start working with the project, it will not work out of the box.
This small PR adds the prettier as a dependency.


Yuyz0112:
Thanks

PR #490, opened 2021-02-16
Adding prepare npm statement

IMFIL:
This will allow people to pull from latest master without needing to wait for the release in NPM.


Yuyz0112:
Thanks!

PR #492, opened 2021-02-16
Added mousemoveCallback threshold option to sampling config.

provinceinnovation:
(non-breaking) 
This PR lets you configure the throttle threshold for the mousemove callback. This enables you to adjust the frequency of mousemove events emission. Default remains at 500ms.

### DIFF
--- a/src/record/observer.ts
+++ b/src/record/observer.ts
@@ -131,6 +131,8 @@ function initMoveObserver(
 
   const threshold =
     typeof sampling.mousemove === 'number' ? sampling.mousemove : 50;
+  const callbackThreshold =
+    typeof sampling.mousemoveCallback === 'number' ? sampling.mousemoveCallback : 500;
 
   let positions: mousePosition[] = [];
   let timeBaseline: number | null;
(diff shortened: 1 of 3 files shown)


Yuyz0112:
Thanks @provinceinnovation 

Could you tell me the use case of configuring this threshold?

provinceinnovation:
This will allow us to configure the mouse movement second layer to transmit the mouse coordinate set more frequently than once every 500 ms. We are using this to so that we can provide a less laggy mouse movement playback in livemode with no playback 1000ms buffer.

(cc. https://github.com/rrweb-io/rrweb/blob/master/docs/observer.md#mouse-movement)

galesky:
Thanks @provinceinnovation !

Could you also update `docs/observer.md` so that it includes this new parameter ?

I'd also suggest changing the name of the variable `threshold` to something like `mouseMoveThreshold`
and `callbackThreshold` to `mouseTransmitThreshold`

Yuyz0112:
It's reasonable to me @provinceinnovation 

Thanks!

PR #494, opened 2021-02-18
Add ability to mask passwords

Juice10:
In our case we'd really like to know if someone did use the password field, but for security we don't want to know what the contents are. This PR moves passwords from the ignore to the mask list, and allows people to configure what they need.

Corresponding PR in rrweb-snapshot: https://github.com/rrweb-io/rrweb-snapshot/pull/65


Yuyz0112:
LGTM

This will be marked as a breaking change behavior in v1.0.0.

PR #495, opened 2021-02-18
Add `userTriggered`

Juice10:
## What does userTriggered indicate?
`userTriggered` gets added to input events that indicates if this event was triggered directly by the user or not.
Example of `userTriggered` in action: User clicks on radio element (userTriggered: true) which triggers the other radio element to change (userTriggered: false)

## Background:
When analysing RRweb events I was having a hard time trying to figure out if an event was directly generated by a user or not. This PR adds a notation to indicate wherever this was the case or not.

Naming considerations: I was doubting if userTriggered or [isTrusted](https://developer.mozilla.org/en-US/docs/Web/API/Event/isTrusted) was the right name for this attribute. But since isTrusted doesn't completely cover all the cases I thought `userTriggered` might be a better name.


Yuyz0112:
Thanks, @Juice10 !

Things I'm thinking about
1. So we can put more things into `userTriggered` when we have things other than `isTrusted`?
2. Although it's just a boolean flag to some events, it's still something not related to `record and replay`, because it does not have a visual effect. It's totally okay to add this property, but I'm wondering whether we should implement a plugin system so only users who use the `userTriggeredPlugin` will have this data collected. This may help us have a minimal data structure and flexible extensions.

Juice10:
> Thanks, @Juice10 !
> 
> Things I'm thinking about
> 
> 1. So we can put more things into `userTriggered` when we have things other than `isTrusted`?
Because this also covers events for radio buttons that get automatically get unselected when you click on a different radio button. I believe technically ‘isTrusted’ would be true in this case. (But I could be wrong and then ‘isTrusted’ would be a perfect name for it.)
> 2. Although it's just a boolean flag to some events, it's still something not related to `record and replay`, because it does not have a visual effect. It's totally okay to add this property, but I'm wondering whether we should implement a plugin system so only users who use the `userTriggeredPlugin` will have this data collected. This may help us have a minimal data structure and flexible extensions.
Could you elaborate a bit on how you’d like the plug-in system to work? I’m guessing you’d want something more elaborate than a simple config flag to toggle this data collection on and off.

eoghanmurray:
> it's still something not related to record and replay, because it does not have a visual effect

I'd say that in the context of a replayer, there may be cases where you'd want to have a visual effect when you know an event was performed by the user vs. otherwise.

eoghanmurray:
@Yuyz0112 I talked to Justin about this today.
IMO this should be a config switch rather than a plugin.
I think plugins should be reserved for large features which would otherwise bulk out the .min.js payload if they were a config option.
E.g. with #598 would src/record/stringify.ts and `initLogObserver` code be excluded from the packed js if the `replayConsole` plugin is not included?

Yuyz0112:
@eoghanmurray 

Agree! I've found this feature needs much tiny modification in the core code path, so it's not a good idea to be a plugin.

@Juice10 Could you add an option for this feature?

BTW, I think #598 already helps prune the stringify/initLogObserver code from the main package.

Juice10:
@Yuyz0112 I'd be happy to, will let you know when its done!

Juice10:
@eoghanmurray @Yuyz0112 I added userTriggered behind the `userTriggeredOnInput: true` flag in the config and added tests and documentation for it as well.

Yuyz0112:
The code looks good to me. Not sure why there is some code diff from other commits.

Juice10:
@Yuyz0112 ah I think that's because I did a git rebase. I can create a new PR which won't have the excessive diff if you'd like

Juice10:
@Yuyz0112 fixed it, the diff should be correct now on this PR

Yuyz0112:
@Juice10 So we only apply this to input event, but not to other events like mouse interactions?

Juice10:
@Yuyz0112 yes thats correct

Juice10:
@Yuyz0112 Im happy to implement this for mouse as well later on, once I've found some good ways of replicating it.

src/replay/index.ts
```diff
@@ -115,6 +115,7 @@ export class Replayer {
       triggerFocus: true,
       UNSAFE_replayCanvas: false,
       pauseAnimation: true,
+      userTriggeredOnInput: true,
```
Mark-Fenng, line 118: @Juice10 I found this option doesn't perform any work in the replay part. Did I miss something?

Juice10: Good catch @Mark-Fenng! It should only be on the recording end. I'm going to remove it on the replayer today.

PR #497, opened 2021-02-18
Add yarn support for installing unreleased rrweb as a dependency 

Juice10:
Adds support for pointing your package.json to master and installing via yarn.

[Yarn v2 only supports the `prepack` script](https://yarnpkg.com/advanced/rulebook/#packages-should-use-the-prepack-script-to-generate-dist-files-before-publishing) which does pretty much the same thing as `prepare`.
And yarn v1+ needs a `.npmignore` otherwise it will use the `.gitignore` and wipe the bundled code.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,7 @@
+.vscode
+.idea
+node_modules
+package-lock.json
+yarn.lock
+temp
+*.log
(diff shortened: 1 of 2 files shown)


Yuyz0112:
Thanks!

PR #500, opened 2021-02-20
update guide

FeSeason:
Typescript & spellcheck plays an important role, but in order to prevent more people from copying documents directly, it is necessary to modify them.


Mark-Fenng:
Thanks!

PR #515, opened 2021-03-08
Handle event undefined in initMoveObserver

yakkomajuri:
A quick suggestion for handling a case where this expression:

```
isTouchEvent(event)
        ? event.changedTouches[0]
        : event;
```

returns `undefined`, causing an error when trying to access `clientX`.

Appears to be the cause of this error: https://sentry.io/share/issue/f01259a2c753426cbfccad01c30a102b/

### DIFF
--- a/src/record/observer.ts
+++ b/src/record/observer.ts
@@ -202,10 +202,12 @@ function initMouseInteractionObserver(
       if (isBlocked(event.target as Node, blockClass)) {
         return;
       }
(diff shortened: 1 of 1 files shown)


Yuyz0112:
Thanks! @yakkomajuri 

I think safely access the index from an array is always reasonable. And of course, it will be cool if we can find what actually happened to let the event be nullable.

Issue #525, opened 2021-03-29
RRWeb replayer issue with CSSStyleSheetRule insertion replay


VladimirMilenko:
I did a couple of tests with styletron CSS in JS library and it appears, that the replay's snapshot functionality is not handling the CSS stylesheetrule insertions correctly when replayed from a certain point.
This seem to be, that snapshot is not correctly handling CSSStyleSheetRule insertion (original issue: https://github.com/rrweb-io/rrweb/issues/104)

Codesandbox with events and replayer:
https://codesandbox.io/s/trusting-diffie-cwwyq?file=/public/index.html

Replay is available here:
https://user-images.githubusercontent.com/18203255/112888359-8a23ca80-90d4-11eb-832b-3260fd692ff7.mov

VladimirMilenko:
So, I did some dubigging.
It seems that issue is originating in:
```ts
        if (usingVirtualParent && placeholderNode) { // here
          parent.replaceChild(target, placeholderNode);
        }
```
After replacement, the target's sheet is becoming `null`.  And therefore, the style doesn't apply. I tried to simply remove that code and i can see, that node is containing all styles no matter the start position

Yuyz0112:
I think this can be closed now:)

PR #529, opened 2021-04-02
Apply CSSRule insertion/deltetion on Flush event

VladimirMilenko:
Currently, during the node replacement, the `sheet` property is being automatically assigned to null. 
In order to support virtual sheet mutation, we need to store that sheet in another structure.
We also cannot reuse the existing reference to sheet, as the value at this reference would be removed at the `replaceChild` call.

To avoid that, we will create:
```ts
  private virtualStyleRulesMap!: Map<INode, CSSRule[]>;
```
This rule array map would keep the latest CSSRule array for each node during the virtual processing of events and would ensure, that at the Flush event, `style` node has exact same array of css rules, as the one in memory.

I've also added a test which is able to check, whether the stylesheet rule exists in the sheet after insertion.

Fixes https://github.com/rrweb-io/rrweb/issues/525

### DIFF
--- a/test/replayer.test.ts
+++ b/test/replayer.test.ts
@@ -146,6 +146,19 @@ describe('replayer', function (this: ISuite) {
       replayer.play(1500);
       replayer['timer']['actions'].length;
     `);
+
+    const result = await this.page.evaluate(`
+      const rules = Array.from(replayer.iframe.contentDocument.head.children)
+        .filter(x=>x.nodeName === 'STYLE')
+        .reduce((acc, node) => {
+          acc.push(...node.sheet.rules);
+          return acc;
+        }, []);
+
+        rules.some(x=>x.selectorText === ".css-1fbxx79")
+    `);
+
+    expect(result).to.equal(true);
     expect(actionLength).to.equal(
       styleSheetRuleEvents.filter(
         (e) => e.timestamp - styleSheetRuleEvents[0].timestamp >= 1500,
--- a/test/replayer.test.ts
+++ b/test/replayer.test.ts
@@ -183,6 +183,31 @@ describe('replayer', function (this: ISuite) {
     );
   });
 
+  it('can fast forward past StyleSheetRule deletion on virtual elements', async () => {
+    await this.page.evaluate(
+      `events = ${JSON.stringify(styleSheetRuleEvents)}`,
+    );
+    const actionLength = await this.page.evaluate(`
+      const { Replayer } = rrweb;
+      const replayer = new Replayer(events);
+      replayer.play(2500);
+      replayer['timer']['actions'].length;
+    `);
+
+    const result = await this.page.evaluate(`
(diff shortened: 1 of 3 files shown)


VladimirMilenko:
cc @Juice10
This change is removing the check you added. Could you provide some info, whether that might create an issue?

Yuyz0112:
LGTM

@Mark-Fenng PTAL

Mark-Fenng:
A great way to fix the issue.
BTW, is placeholderNode still necessary to exist?

test/replayer.test.ts
```diff
@@ -146,6 +146,19 @@ describe('replayer', function (this: ISuite) {
...
+
+    const result = await this.page.evaluate(`
+      const rules = Array.from(replayer.iframe.contentDocument.head.children)
+        .filter(x=>x.nodeName === 'STYLE')
+        .reduce((acc, node) => {
+          acc.push(...node.sheet.rules);
+          return acc;
+        }, []);
+
+        rules.some(x=>x.selectorText === ".css-1fbxx79")
```
Juice10, line 158: Would be nice if we could simplify this a little
```suggestion
      const rules = [...replayer.iframe.contentDocument.styleSheets].map((sheet)=> sheet.rules);
      rules.some(x=>x.selectorText === ".css-1fbxx79")
```

test/replayer.test.ts
```diff
@@ -170,6 +183,31 @@ describe('replayer', function (this: ISuite) {
...
```
Juice10, line 205: Same goes for here
```suggestion
      const rules = [...replayer.iframe.contentDocument.styleSheets].map((sheet)=> sheet.rules);
      rules.some(x=>x.selectorText === ".css-1fbxx79")
```

Juice10 (changes requested):
Really loving this, great work @VladimirMilenko

Juice10:
> A great way to fix the issue.
> BTW, is placeholderNode still necessary to exist?

As far as I can remember it is necessary the first time the `.sheet` is accessed.
But for performance reasons it would be advisable skip the placeholder/adding-stylesheet-to-dom part when the node already exists in the `virtualStyleRulesMap`.

Juice10:
It looks like constructible stylesheets (`new CSSStyleSheet`) is still in the [proposal phase](https://wicg.github.io/construct-stylesheets/) and is only implemented in Chrome, and [Firefox (behind experimental flag)](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#constructable_stylesheets).
In Safari I couldn't get it working directly but I was able to access it via a workaround: `Object.create(CSSStyleSheet.prototype)`

We might be able to get it working by using this polyfill: https://github.com/calebdwilliams/construct-style-sheets

Juice10:
@VladimirMilenko feel free to close this PR. I ended up incorporating most of it in https://github.com/rrweb-io/rrweb/pull/618 which fixes this issue.

PR #532, opened 2021-04-08
Don't rebuildFullSnapshot if it's the first run

JohnPhamous:
👋  I don't think we need to rebuild the full snapshot during the first play-through since we already do it when we initialize the player.

Before:
- 1 `rebuildFullSnapshot` while initializing the player
- 1 `rebuildFullSnapshot` while processing the `FullSnapshot` event

After:
- 1 `rebuildFullSnapshot` while initializing the player

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -121,7 +121,7 @@ export class Replayer {
   private imageMap: Map<eventWithTime, HTMLImageElement> = new Map();
   
   /** The first time the player is playing. */
-  private firstPlay = false;
+  private firstPlay = true;
                              
   private newDocumentQueue: addedNodeMutation[] = [];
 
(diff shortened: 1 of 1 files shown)


JohnPhamous:
@Yuyz0112 Could you take a look when you can?

src/replay/index.ts
```diff
@@ -459,6 +462,11 @@ export class Replayer {
         break;
       case EventType.FullSnapshot:
         castFn = () => {
+          // Don't build a full snapshot during the first play through since we've already built it when the player was mounted.
+          if (this.firstPlay) {
```
Yuyz0112, line 466: where do we set `firstPlay` to truthy?

JohnPhamous: Whoops, fixed the typo. It's initialized to true at first.

Yuyz0112:
Thanks!

Issue #161, opened 2020-01-07
[Question] Streaming/Paginating events while replaying


praneetnadkar:
If the events data to be replayed is huge, then it will affect the performance. Any way we can stream the events while replay or paginate them?

pnadkar:
Can someone help me on this:?

bugwheels94:
Hi, Any way to achieve this?

Yuyz0112:
Now you can simply use `addEvent` to do this.

```
const replayer = new rrweb.Replayer(events)

for (const newEvent of newEvents) {
  replayer.addEvent(newEvent)
}
```

tamilvendhank-helptap:
@Yuyz0112, how to know I need new set of events? Could you give a full example that includes initializing the player, fetching the data based on some event & adding it to the player?

ShayMalchi:
@Yuyz0112 Hi, using `addEvent` works but the progress bar remains out of date (inactive time in particular) and does not update dynamically like the duration. On hard reload it draws correctly. How can I force it to update as well?

PR #539, opened 2021-04-15
Don't perform newly added actions if player is paused

eoghanmurray:
I've a scenario where newly added events which are coming in after the current play position are affecting the player state even when the player is paused.
I believe the correct thing to do (as per this pull request) is not to execute them until the player is resumed.
I see that starting the `timer.start();` line was added as part of #161 but I can't quiet understand what the idea was there?
I've changed it to instead only queue up the event if the timer is currently executing, which makes more sense to me, but I might be missing something?

### DIFF
--- a/src/replay/machine.ts
+++ b/src/replay/machine.ts
@@ -242,7 +242,7 @@ export function createPlayerService(
             const castFn = getCastFn(event, isSync);
             if (isSync) {
               castFn();
-            } else {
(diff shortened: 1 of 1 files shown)


Yuyz0112:
@eoghanmurray Yes, I believe it's a bug in the original design.

Yuyz0112:
Thanks!

PR #547, opened 2021-04-23
fix: #542 wrong results of splitting log stacks

Mark-Fenng:
fix bug of stack parcer and increase compatibility for different browser vendors


Yuyz0112:
LGTM!

PR #549, opened 2021-04-24
Subscribe to MetaData change

theshakilahmed:
`setMetaDataSubscriber` allows for listening to metadata changes when new events are added using the replayer's `addEvent` method.
It'd be a convenient way to allow use cases such as this issue: rrweb-io/rrweb-player#62

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -30,6 +30,7 @@ import {
   LogReplayConfig,
   logData,
   ReplayLogger,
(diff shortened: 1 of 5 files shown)


eoghanmurray:
Thank you for your pull request and apologies that it was never reviewed and is now sitting near the bottom of the pile.

Do you know if this usecase is still valid and required? 
Is the workaround in the linked issue regarding 'liveMode' sufficient?

Issue #548, opened 2021-04-23
Error: Cannot read property 'implementation' of null


Juice10:
# The Error
The current master version of rrweb triggers a "Cannot read property 'implementation' of null" error in some situations.

# The Bug

I did some digging and it turns out iframes inside of a DocumentFragment don't have a `iframe.contentDocument`.
https://github.com/rrweb-io/rrweb/blob/2b96a68e881695c54bbf6808e7df17c62adb1c21/src/replay/index.ts#L625-L631

Whenever `applyMutation` gets called with `useVirtualParent: true` and a mutation containing an iframe, rrweb ends up calling `buildNodeWithSN` with an empty `doc` property which ends up triggering the error in `buildNode` on: [Source](https://github.com/rrweb-io/rrweb-snapshot/blob/7850a75c835bc37fd42db0e4ddb28c640db958de/src/rebuild.ts#L116)

```ts
      return doc.implementation.createDocument(null, '', null);
//            ^--- null
```

Fixing this line and most other lines in `buildNode` should be quite easy, just replacing `doc` with `document` should do the trick.

For `buildNodeWithSN` however we need a different approach: [Source](https://github.com/rrweb-io/rrweb-snapshot/blob/7850a75c835bc37fd42db0e4ddb28c640db958de/src/rebuild.ts#L262-L268)

```ts
  if (n.type === NodeType.Document) {
    // close before open to make sure document was closed
    doc.close();
//   ^--- null
    doc.open();
    node = doc;
//   ^-- node is extensively used throughout buildNodeWithSN
  }
```

I wasn't really able to go beyond this unfortunately.

# The Test (to reproduce the bug)
Add to `test/replayer.test.ts`:
```ts
import iframeEvents from './events/iframe-events';
```
```ts
  it('can fast forward past iframe changes on virtual elements', async () => {
    await this.page.evaluate(`events = ${JSON.stringify(iframeEvents)}`);
    const actionLength = await this.page.evaluate(`
      const { Replayer } = rrweb;
      const replayer = new Replayer(events);
      replayer.play(3000);
      replayer['timer']['actions'].length;
    `);
    expect(actionLength).to.equal(
      iframeEvents.filter(
        (e) => e.timestamp - iframeEvents[0].timestamp >= 3000,
      ).length,
    );
  });
```
Contents of `test/events/iframe-events.ts`: [iframe-events.ts.zip](https://github.com/rrweb-io/rrweb/files/6366436/iframe-events.ts.zip)

[...]

PR #550, opened 2021-04-26
fix: issue #548

Mark-Fenng:
1. Do not use virtual parent optimization if the mutation targets have iframe elements as children. This will cause some performance regression but will be easy to add and ship.
2. If an iframe element has already been a child of a virtual parent, add the virtual parent back to the dom.

@Juice10 Please take a look at this PR.


Juice10 (approved):
This looks great @Mark-Fenng! Love the pragmatic approach you and @Yuyz0112 came up with. We can always spend some time making it faster again, but at least now we'll have something that works.

PR #558, opened 2021-05-14
Ascending timestamps

eoghanmurray:
This started with the idea that we don't want 'overlapping' events, i.e. a mouse click event followed by a mousemove event whose negative timestamps put the start of the mouse movement before the click.

This is important in the logical processing of a series of events so that we don't have to consider the next event when processing the current event e.g. in building up a graphical timeline.

It could also be important in streaming 'live' events to a player, as although a mousemove can still introduce up to 500ms lag, it now can't result in an out-of-order situation where that click is played back before a mousemove that should precede it.

There seem to be no tests that recreate the mouse movement due to the mentioned difficulty with 'random' mouse movements in the headless browser.  I've added a test (via util.py) to check that timestamps are weakly ascending before we strip them out, but that didn't catch anything prior to these changes.  Any advice on how to increase the test coverage for this commit appreciated!

### DIFF
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -131,9 +131,15 @@ function record<T = eventWithTime>(
         e.data.source === IncrementalSource.Mutation
       )
     ) {
+      let mtimestamp = e.timestamp;
+      if ('positions' in e.data && e.data.positions[0].timeOffset) {
+        // assign the mutation timestamp to the beginning
+        // of mouse/touch movement
+        mtimestamp += e.data.positions[0].timeOffset;
+      }
       // we've got a user initiated event so first we need to apply
       // all DOM changes that have been buffering during paused state
-      mutationBuffers.forEach((buf) => buf.unfreeze(e.timestamp));
+      mutationBuffers.forEach((buf) => buf.unfreeze(mtimestamp));
     }
     if (ongoingMove) {
       // emit any ongoing (but throttled) mouse or touch move;
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -1,5 +1,5 @@
 import { snapshot, MaskInputOptions, SlimDOMOptions } from 'rrweb-snapshot';
-import { initObservers, mutationBuffers } from './observer';
+import { initObservers, mutationBuffers, ongoingMove } from './observer';
 import {
   on,
   getWindowWidth,
@@ -131,6 +131,12 @@ function record<T = eventWithTime>(
       // we've got a user initiated event so first we need to apply
       // all DOM changes that have been buffering during paused state
       mutationBuffers.forEach((buf) => buf.unfreeze());
+    }
+    if (ongoingMove) {
+      // emit any ongoing (but throttled) mouse or touch move;
+      // emitting now creates more events, but ensures events are emitted in
+      // sequence without any overlap from the negative Move timeOffset
+      ongoingMove();
     }
 
     emit(((packFn ? packFn(e) : e) as unknown) as T, isCheckout);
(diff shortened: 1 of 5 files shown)


eoghanmurray:
The downside of this pull request is that now long mouse movements can potentially be split up into many shorter events.

A danger would be on a page which has many ongoing background mutations ... let me know if that is a concern.

packages/rrweb/src/record/index.ts
```diff
@@ -22,7 +22,10 @@ import {
 import { IframeManager } from './iframe-manager';
 import { ShadowDomManager } from './shadow-dom-manager';
 
-function wrapEvent(e: event): eventWithTime {
+function wrapEvent(e: event | eventWithTime): eventWithTime {
+  if ((e as eventWithTime).timestamp) {
```
Yuyz0112, line 26: Does this get rid of type assertion?

```ts
if ('timestamp' in e) {}
```

packages/rrweb/src/record/index.ts
```diff
@@ -158,14 +175,18 @@ function record<T = eventWithTime>(
     }
   };
 
-  const wrappedMutationEmit = (m: mutationCallbackParam) => {
+  const wrappedMutationEmit = (m: mutationCallbackParam, timestamp?: number) => {
+    if (typeof timestamp === 'undefined') {
+      timestamp = Date.now()
```
Yuyz0112, line 180: why mutation emit need to create timestamp by itself while other callback still using the `wrappedEmit` method?

packages/rrweb/src/record/index.ts
eoghanmurray: Yes, thanks that is better

packages/rrweb/src/record/index.ts
eoghanmurray: This should be written without a `timestamp = Date.now();` in this function; I will rewrite it now and force push this and prior change.

eoghanmurray: I've made that change now, but maybe you meant something else?

eoghanmurray:
I definitely need to create some proper tests for this PR as it involves intricate interactions between ongoing mouse movement and interrupting mutations.  Actually I haven't fully thought about whether splitting a single mouse movement event in two will have any differences in replay rendering?  

Even with this PR, I still think it's a good idea to emit an event at the start of mousemovent for the use of live replayers and for cleaner processing (even though that starting timestamp info is also contained in the eventual throttled event).

packages/rrweb/src/record/index.ts
```diff
@@ -128,9 +139,15 @@ function record<T = eventWithTime>(
         e.data.source === IncrementalSource.Mutation
       )
     ) {
+      let mtimestamp = e.timestamp;
+      if ('positions' in e.data && e.data.positions[0].timeOffset) {
+        // assign the mutation timestamp to the beginning
+        // of mouse/touch movement
+        mtimestamp += e.data.positions[0].timeOffset;
```
Yuyz0112, line 146: @eoghanmurray What I get confused about is this part.

Why the position data need get timestamp from a mutation event, instead of recording by itself as the emitted?

eoghanmurray: The scenario here is some frozen mutations being interrupted by a mouse move event.
So the mutations happened at some prior time and continued to be frozen when the mouse started moving; but were only unfrozen after the buffered mouse move event is emitted.  So we are ensuring that the 'unfreeze' time is prior to the start of the first mouse movement.

I've just realized that this wouldn't be necessary if we added that placeholder event for 'mouse movement started' as briefly discussed previously.

Yuyz0112: Yes, that will be a little neater.

Would like to do it in this PR or in the future?

eoghanmurray:
Can retry (rebase on master) after #1475 is merged as I incorporated some of this PR into that PR

eoghanmurray:
So I did some quantitative analysis on this on a corpus of rrweb events with this PR deployed, and got the AI to reconstruct how many batched events had been split up that would not have been before (based on timestamp adjacency).

Bottom line: this PR generated an estimated 54% more mouse move events (as anticipated) by sometimes preventing the full 500ms batching that has been there since `initMousemoveObserver` was introduced.  We get the same `positions` data output, so no effect on replay, just that these events are more often emitted separately with smaller numbers of positions.  The per-event overhead is 70 bytes and looking at this extra bytes vs. other event types, the effect is negligible.

Issue #567, opened 2021-05-26
Bug: Iframe's contents (IncrementalSnapshot) recorded before iframe (FullSnapshot)


Juice10:
Problem:
rrweb creates a IncrementalSnapshot event to add content of the iframe ___before___ it creates a FullSnapshot event that includes the (empty) iframe.

Expected behaviour: 
First it should create the FullSnapshot (with the iframe), then an IncrementalSnapshot (with the contents of the iframe).

Example of a failing test that should work: https://github.com/rrweb-io/rrweb/compare/master...Juice10:iframe-event-order?expand=1#diff-27a64575524d0bae582e86c978c4e655648fc1fa3296bbe01dbe129e285ea8e1
Results:
```
AssertionError: expected [ 3, 2 ] to have the same ordered members as [ 2, 3 ]
    at Suite.<anonymous> (test/record.test.ts:307:40)
    at step (test/record.test.ts:33:23)
    at Object.next (test/record.test.ts:14:53)
    at fulfilled (test/record.test.ts:5:58)

+ expected - actual

 [
+  2
   3
-  2
 ]
```

Example of events recorded (comments inline):
```jsonc
[
  {
    "type": 4,
    "data": { "href": "about:blank", "width": 1920, "height": 1080 },
    "timestamp": 1622026127159
  },
  {
    // IncrementalSnapshot (type: 3) before FullSnapshot (type: 2)!
    "type": 3,
    "data": {
      "source": 0,
      "adds": [
        {
          // node with id 6 doesn't exist yet
          "parentId": 6,
          "nextId": null,
          "node": {
            "type": 0,
            "childNodes": [
              {
                "type": 2,
                "tagName": "html",
                "attributes": {},
                "childNodes": [
                  {
                    "type": 2,
                    "tagName": "head",
                    "attributes": {},
                    "childNodes": [],
                    "rootId": 8,
                    "id": 10
                  },
                  {
                    "type": 2,
                    "tagName": "body",
                    "attributes": {},
                    "childNodes": [
                      {
                        "type": 2,
                        "tagName": "button",
                        "attributes": {},
                        "childNodes": [
                          { "type": 3, "textContent": "Mysterious Button", "rootId": 8, "id": 13 }
                        ],
                        "rootId": 8,
                        "id": 12
                      }
                    ],
                    "rootId": 8,
                    "id": 11
                  }
                ],
                "rootId": 8,
                "id": 9
              }
            ],
            "id": 8
          }
        }
      ],
[...]

PR #568, opened 2021-05-26
Tests: Iframe event order

Juice10:
Adds tests ~and fixes~ _for_ https://github.com/rrweb-io/rrweb/issues/567

_These tests will pass once the changes are accepted in https://github.com/rrweb-io/rrweb-snapshot/pull/78 and a new version of rrweb-snapshot is added in package.json_


Juice10:
I'm not sure if using `setTimeout` is the right approach here. It works but if we could do it without `setTimeout` I would prefer that.

@eoghanmurray do you think it would be possible to use the MutationBuffers to keep these mutations from getting added until the FullSnapshot has been built?

Yuyz0112:
Hi @Juice10, thanks for the PR.

I think https://github.com/rrweb-io/rrweb-snapshot/pull/78 is a better approach.

But I'm curious because I think I already handle the order issue with this [queue](https://github.com/rrweb-io/rrweb/blob/master/src/replay/index.ts#L1138). Are you experiencing any issues in the replay?

Juice10:
> Hi @Juice10, thanks for the PR.
My pleasure!
> I think [rrweb-io/rrweb-snapshot#78](https://github.com/rrweb-io/rrweb-snapshot/pull/78) is a better approach.
I totally agree, I combed through the snapshot output and PR 78 seems consistent and in the correct order.
> But I'm curious because I think I already handle the order issue with this [queue](https://github.com/rrweb-io/rrweb/blob/master/src/replay/index.ts#L1138). Are you experiencing any issues in the replay?
Yes, the new document queue only works if the parentId’s id is not in use and the parent can’t be found, I my case I tend to come across a lot of ids that are reused so the parent is found and the new document queue isn’t triggered.

eoghanmurray:
> @eoghanmurray do you think it would be possible to use the MutationBuffers to keep these mutations from getting added until the FullSnapshot has been built?

Maybe! (I've only recently had a chance to glance at the changes that resulted in a mutationbuffer per iframe)

Yuyz0112:
@Juice10 I've published rrweb-snapshot v1.1.4

Yuyz0112:
looks like CI is failing https://travis-ci.com/github/rrweb-io/rrweb/builds/228652215

Juice10:
There seem to be some inconsistencies between my development environment and CI. Solved most of them by upgrading puppeteer. I ran out of time yesterday but I'll try to fix the last one today.

Juice10:
@Yuyz0112 looks like this PR is good to go now!

Issue #573, opened 2021-06-03
Media playback doesn't work 


bachmanity1:
Video crashes during the playback of events recorded from `iframe` containing `media` data. 

After the crash I get error message shown below:
`Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause(). https://goo.gl/LdLk22`

https://github.com/rrweb-io/rrweb/blob/cbcd46a5b1ccb30db9d87b2393ac5bed9a162d3f/src/replay/index.ts#L901-L927

I think the problem is that you ignore the return value of `play()` function, which is actually asynchronous and returns a promise. [This link]( https://goo.gl/LdLk22) explains this bug.

PR #574, opened 2021-06-05
Disable full snapshot rebuild after first event

bachmanity1:
Currently full snapshot is built even after first event has already been played. 
This leads to bugs such as media playback crash when media has already started playing. #573

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -462,9 +462,10 @@ export class Replayer {
       case EventType.FullSnapshot:
         castFn = () => {
           // Don't build a full snapshot during the first play through since we've already built it when the player was mounted.
-          if (this.firstPlayedEvent && this.firstPlayedEvent === event) {
+          if (this.firstPlayedEvent) {
             return;
           }
+          this.firstPlayedEvent = event
           this.rebuildFullSnapshot(event, isSync);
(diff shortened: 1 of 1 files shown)


bachmanity1:
@Yuyz0112 can you please check out this?

Yuyz0112:
@bachmanity1 A session may have more than one full snapshot, so I think we need to check whether the latest played full snapshot event is the one we will rebuild now.

bachmanity1:
I don't understand why you would want to rebuild full snapshot after you've done it once. You even have this comment in your code `Don't build a full snapshot during the first play through since we've already built it when the player was mounted.`
And I don't understand how we can compare two full snapshots, do we need to compare them node by node? In your code you are comparing them by reference like `this.firstPlayedEvent === event` but I think this condition is never going to hold true.

bachmanity1:
@Yuyz0112 you may ignore my previous comment. The PR that I submitted in `rrweb-snapshot` resolves #573. There is one more thing though. Rebuilding page that contains `HTMLMediaElement` makes video appear glitchy, especially when checkout period is short. How about an option to disable full snapshot rebuild after the first build?

bachmanity1:
I just realized that it's possible to disable full snapshot rebuild in the application code without changing anything in `rrweb` itself. So I'm closing this issue as it's resolved now.

PR #586, opened 2021-06-16
Fix order of ternary expression for mouse move source type

yashkumar18:
DragEvent is inherited from MouseEvent so the first expression evaluates to true for both MouseEvent and DragEvent which fills in incorrect source for the event.

Doesn't affect the re-player since everything except the source is same and it's played correctly as mouse moves as expected.

Due the inheritance between the events, we should first check the instance type for the sub event.


Yuyz0112:
Thanks!

PR #587, opened 2021-06-17
Handle Drag events in replayer

yashkumar18:
Drag events are not handled in replayer.
This doesn't break the replay but the mouse movement during a 'drag' is skipped directly to the event.

Drag and MouseMove for the replayer is identical, so this source type can fallback to MouseMove.

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -783,6 +783,7 @@ export class Replayer {
         this.applyMutation(d, isSync);
         break;
       }
+      case IncrementalSource.Drag:
       case IncrementalSource.MouseMove:
         if (isSync) {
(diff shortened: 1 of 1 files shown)


src/replay/index.ts
```diff
@@ -783,6 +783,7 @@ export class Replayer {
         this.applyMutation(d, isSync);
         break;
       }
+      case IncrementalSource.Drag:
```
Yuyz0112, line 786: maybe add TouchMove either?

Yuyz0112 (changes requested):

src/replay/index.ts
yashkumar18: Sound good

Yuyz0112 (approved):

PR #588, opened 2021-06-18
Test that the observer is monitoring changes on the <html> element itself

eoghanmurray:
I'm running into this problem where mutations are missed on the <html> element, e.g. important ones where a class is added to the html element to blank out the screen, and then subsequently removed when content has loaded... recording just stays blank.

Creating this pull request to confirm problem is in rrweb (my local test environment hasn't been working)


eoghanmurray:
From the test result this actually appears to be working fine, so I think the problem I'm seeing is elsewhere.

PR #593, opened 2021-06-23
implement video seek feature

bachmanity1:
Listening for `seeked` event helps to synchronize media content between recorded and replayed page.

### DIFF
--- a/src/record/observer.ts
+++ b/src/record/observer.ts
@@ -483,7 +483,7 @@ function initMediaInteractionObserver(
   mirror: Mirror,
 ): listenerHandler {
   const handler = (type: MediaInteractions ) => (event: Event) => {
-    const { target } = event;
+    const target = getEventTarget(event);
     if (!target || isBlocked(target as Node, blockClass)) {
       return;
     }
--- a/src/record/observer.ts
+++ b/src/record/observer.ts
@@ -482,17 +482,24 @@ function initMediaInteractionObserver(
   blockClass: blockClass,
   mirror: Mirror,
 ): listenerHandler {
-  const handler = (type: 'play' | 'pause') => (event: Event) => {
-    const target = getEventTarget(event);
+  const handler = (type: MediaInteractions ) => (event: Event) => {
+    const { target } = event;
     if (!target || isBlocked(target as Node, blockClass)) {
       return;
     }
     mediaInteractionCb({
-      type: type === 'play' ? MediaInteractions.Play : MediaInteractions.Pause,
+      type,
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -474,11 +474,15 @@ export type inputCallback = (v: inputValue & { id: number }) => void;
 export const enum MediaInteractions {
   Play,
   Pause,
+  Seeked,
 }
 
 export type mediaInteractionParam = {
   type: MediaInteractions;
   id: number;
+  attributes: {
```
Yuyz0112, line 483: remove the `attributes` namespace looks similar to other param structure

Yuyz0112: BTW, since we do not have `currentTime` property before this patch, we should mark this property as optional `currentTime?: number`, so the replayer knows it will handle legacy events without this property.

Yuyz0112 (changes requested):

bachmanity1:
@Yuyz0112 I reflected your comments.

src/replay/index.ts
```diff
@@ -910,14 +910,14 @@ export class Replayer {
             mediaEl.pause();
           }
           if (d.type === MediaInteractions.Play) {
-            mediaEl.currentTime = d.attributes.currentTime;
+            mediaEl.currentTime = d.currentTime as number;
```
Yuyz0112, line 913: maybe this?

```ts
if (d. currentTIme) {
  mediaEl.currentTime = d.currentTime
}
```

bachmanity1: changed the code.

src/record/observer.ts
```diff
@@ -482,17 +482,22 @@ function initMediaInteractionObserver(
   blockClass: blockClass,
   mirror: Mirror,
 ): listenerHandler {
-  const handler = (type: 'play' | 'pause') => (event: Event) => {
-    const target = getEventTarget(event);
+  const handler = (type: MediaInteractions ) => (event: Event) => {
+    const { target } = event;
```
Yuyz0112, line 486: @bachmanity1 Sorry I missed this line in the previous review.

`getEventTarget` is a util to get event target from normal DOM scope and shadow DOM, so I think it's needed here.

Yuyz0112 (changes requested):

src/record/observer.ts
bachmanity1: fixed.

Yuyz0112:
@bachmanity1 Thanks!

Yuyz0112 (approved):

PR #602, opened 2021-06-28
Mask value attribute changes for elements in maskInputOptions

Juice10:
I noticed that [some frameworks](https://reactjs.org/docs/forms.html#controlled-components) keep the contents of their form fields in sync with their value attributes.
Since we were not filtering mutations to value attributes for sensitive fields, that means that passwords and other masked information could leak into rrweb events.

This PR masks attribute mutations on value attributes for masked input fields.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -55,7 +55,7 @@ export function createMirror(): Mirror {
       delete this.map[id];
       if (n.childNodes) {
         n.childNodes.forEach((child) =>
-          this.removeNodeFromMap(child as Node as INode),
+          this.removeNodeFromMap((child as Node) as INode),
         );
       }
     },
@@ -277,7 +277,7 @@ export function isAncestorRemoved(target: INode, mirror: Mirror): boolean {
   if (!target.parentNode) {
     return true;
   }
-  return isAncestorRemoved(target.parentNode as unknown as INode, mirror);
+  return isAncestorRemoved((target.parentNode as unknown) as INode, mirror);
 }
 
 export function isTouchEvent(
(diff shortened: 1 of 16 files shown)


src/utils.ts
```diff
@@ -650,5 +656,34 @@ export function getBaseDimension(
 export function hasShadowRoot<T extends Node>(
   n: T,
 ): n is T & { shadowRoot: ShadowRoot } {
-  return Boolean((n as unknown as Element)?.shadowRoot);
+  return Boolean(((n as unknown) as Element)?.shadowRoot);
+}
+
+// TODO: move me to rrweb-snapshot
+// we are doing similar things there as well, would be best to keep things consistent
+export function maskInputValue({
```
Yuyz0112, line 664: is it possible to move this code now? I can release a new version of rrweb-snapshot when this gets merged.

Juice10: I'm out of office all day today, earliest I can do this would be about 24 hours from now

Juice10: Maybe we can merge this PR and I can follow up with a cleanup PR that moves this to rrweb-snapshot tomorrow?

Yuyz0112: sure

Yuyz0112:
@Juice10 Let me understand this better.

If someone use React to build an input component and sync the value to DOM attribute:
1. the replay of input typing will be masked correctly
2. users can inspect the replayer and get the value from the DOM attribute

Am I right?

src/utils.ts
Juice10: PR: https://github.com/rrweb-io/rrweb-snapshot/pull/83

Juice10:
@Yuyz0112 correct!

Yuyz0112:
@Juice10 Please upgrade to rrweb-snapshot@1.1.6, thanks!

Yuyz0112:
LGTM

PR #607, opened 2021-06-30
Encountered a bug where firstFullSnapshot was played twice

eoghanmurray:
I believe the Timer was immediately started and reached the snapshot before the setTimeout returned. I feel this implementation is more robust compared rather than relying on an external firstPlayedEvent that needs to be populated.

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -89,8 +89,6 @@ export class Replayer {
   private imageMap: Map<eventWithTime, HTMLImageElement> = new Map();
 
   private mirror: Mirror = createMirror();
-  /** The first time the player is playing. */
-  private firstPlayedEvent: eventWithTime | null = null;
 
   private newDocumentQueue: addedNodeMutation[] = [];
 
@@ -145,7 +143,6 @@ export class Replayer {
       }
     });
     this.emitter.on(ReplayerEvents.PlayBack, () => {
-      this.firstPlayedEvent = null;
       this.mirror.reset();
     });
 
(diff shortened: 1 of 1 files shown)


eoghanmurray:
Here's an alternative fix that I looked at first:
https://github.com/statcounter/rrweb/commit/0bc5bc3d634b083a97b3570cc8fae57f53b3ff1d

src/replay/index.ts
```diff
@@ -206,11 +203,12 @@ export class Replayer {
...
       setTimeout(() => {
-        // when something has been played, there is no need to rebuild poster
-        if (this.firstPlayedEvent) {
+        const fIndex = this.service.state.context.events.indexOf(firstFullsnapshot);
+        if (fIndex === -1) {
+          // already been played via Timer
           return;
         }
-        this.firstPlayedEvent = firstFullsnapshot;
+        this.service.state.context.events.splice(fIndex, 1);  // remove from Timer queue
```
Yuyz0112, line 211: I think this mutates the events array. What will happen if someone plays the session again?

eoghanmurray: Sorry, need to withdraw this, I thought I was mutating the `timer.actions`

eoghanmurray: I'm also fixing a more serious problem, where the FirstFullSnapshot is still building (probably via the firstFullsnapshot setTimeout), but a mutation gets played by the Timer which tries to mutate an element that isn't there yet.

eoghanmurray:
Ok have gone with https://github.com/rrweb-io/rrweb/pull/608 as I can't see any way to reach into the timer.

PR #608, opened 2021-06-30
Ignore firstFullSnapshot once only after initial 'poster' build

eoghanmurray:
~I believe the Timer was immediately started and reached the snapshot before the setTimeout returned.~

Actually that probably wasn't the case, but this includes a guard against that which prevents the 'poster' build if *any* FullSnapshot has already been built.

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -90,7 +90,7 @@ export class Replayer {
 
   private mirror: Mirror = createMirror();
 
-  private firstFullSnapshot: eventWithTime | null = null;
+  private firstFullSnapshot: eventWithTime | true | null = null;
 
   private newDocumentQueue: addedNodeMutation[] = [];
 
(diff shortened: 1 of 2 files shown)


eoghanmurray:
Ok, now I've got the real fix to my issue included in the second commit.

Scenario was:

 - player built the first FullSnapshot 'poster' before starting the Timer
 - timer starts, correctly doesn't replay first FullSnapshot, but applies some destructive mutations (remove a node)
 - premature finish: playback was FINISHed as we are lazy loading events over the network
 - new events came in so we restarted the player using `.play(current_time)`
 - player requeued up all the already replayed events to get state up to current_time (`neededEvents`); however the first FullSnapshot was incorrectly ignored this time round
 - during this process, the destructive mutations were replayed, however this time they failed, as the node was no longer there to be removed.

Juice10 (approved):
I've been bumping in to this issue as well. I think this would solve it indeed

Juice10:
I tried this PR out and it seems to solve my issues. Looks good to me!

PR #611, opened 2021-07-01
Fix that timer.addAction reverses order when splicing multiple events at same timestamp

eoghanmurray:
Found an error where two mutation events had the same timestamp, but one removed a node added in the other. The `actions.splice` method was reversing their order and triggering a 'Node with id [...] not found' error


Juice10 (approved):

Yuyz0112:
Thanks

PR #618, opened 2021-07-06
Buffer modifications to virtual stylesheets

### LINKED ISSUE #563: CSS stylesheet `insertRule` only works on first snapshot for `IncrementalSource.StylesheetRule`
Hi there. We've been having issues on our recordings where on the first load/playback of a set of events, the `insertRule` method on the `<style/>` element is reflected on the DOM, but when the events up to a certain timestamp are retraced (for a `pause()` action, e.g.) the same changes aren't reflected. 

Specifically, this seems to be happening on `IncrementalSnapshot` events with `IncrementalSource.StyleSheetRule`.

For example, for [these](https://github.com/rrweb-io/rrweb/files/6528739/events.json.zip) events, if you play the recording for the first time, the styles for this input box
 image
  look fine, but if you click back to around 3:50 after playing it through, the styles are no longer there:

image

I'm not sure whether this is the correct way to fix it, but by simply appending rules to the `textContent` of the `<style/>` rule, rather than using `insertRule`, it seems to work.  See here:
https://github.com/jay-khatri/rrweb/pull/1/files

@eoghanmurray @Yuyz0112 any thoughts?

### LINKED ISSUE #614: replayer modal rendering inconsistencies
I'm facing this issue where the same set of events when replayed, sometimes render the modals perfectly, and sometimes completely skips or renders it weirdly. The same events played twice and one time it renders everything perfectly and then when I replayed it again it played the same events with different result where modal isn't visible at all

Juice10:
Fixes #563 #525 and (probably) fixes #614
Builds on https://github.com/rrweb-io/rrweb/pull/529. Special thanks to @VladimirMilenko for doing the initial work on this.

Moves all mutations to virtual stylesheets that aren't part of the dom to `virtualStyleRulesMap`.
When stylesheets get moved to the virtual dom their styles are also backed up to the `virtualStyleRulesMap`.

This PR also adds replayer snapshot testing, and some replay unit tests for restoring virtual styles.


Yuyz0112:
Although it will be a little easier if we can have two PRs for the stylesheet buffer and the new testing infra, I finally finish the review:)

@Juice10 The testing part looks good to me! And huge thanks for picking @VladimirMilenko's work. Nice teamwork.

test/utils.ts
```diff
@@ -78,7 +79,10 @@ function stringifySnapshots(snapshots: eventWithTime[]): string {
               'style' in a.attributes &&
               coordinatesReg.test(a.attributes.style!)
             ) {
-              a.attributes.style = a.attributes.style!.replace(coordinatesReg, '$1: Npx');
+              a.attributes.style = a.attributes.style!.replace(
+                coordinatesReg,
+                '$1: Npx',
+              );
```
eoghanmurray, line 85: @Juice10 I appreciate the wish to apply standardised formatting to things, and I'm going to be doing that now with my own commits whenever I can (now that you've shown me how with prettify!)
however these type of non-functional changes do have a cost in terms of conflict resolution, e.g. when there's a parallel change in an open pull request (or in a custom commit which is to be maintained on top of the master)

Juice10: You are correct, we should include it in a CI step so it gets applied to all

eoghanmurray: Do you mean like a failing test, so that when you create a pull request you know to make the changes yourself?

Juice10: @eoghanmurray Correct!

Juice10: There might be certain hooks we could use to do this automatically as well, but I'm not sure what would be a good one for that.

PR #620, opened 2021-07-07
Node not found warnings

eoghanmurray:
Add a try/catch around all mutations and show a warning rather than stopping playback because of the exception

Change how 'node not found' warnings are reported:
 - ignore ones where the node in question has been removed in the same mutation
 - different message for when a node was removed in a prior mutation vs. trying to alter a node that didn't exist in the last fullsnapshot

### DIFF
--- a/src/replay/index.ts
+++ b/src/replay/index.ts
@@ -1569,7 +1569,11 @@ export class Replayer {
   }
 
   private warnNodeNotFound(d: incrementalData, id: number) {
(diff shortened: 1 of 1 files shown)


Juice10:
I love this, this is going to save many breaking replays in the case of minor hiccups.

It would be nice to see some tests around this though. I'm happy to help you out with them.

Juice10 (approved):
LGTM

PR #626, opened 2021-07-13
Add type checking to ci

Juice10:
I tried to build master and realised code had crept in that doesn't pass typescript's type checking.
This pr adds typescript type checking as one of the steps on CI so we don't accidentally merge PRs that don't pass the type checker.


Yuyz0112:
Thanks!

PR #630, opened 2021-07-13
DragEvent handling, null check fixes

dbseel:
* fix the thrown error caused by DragEvent variable not being defined on some devices/browsers
* add additional node and shadowHost null checks to fix thrown errors

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -271,7 +271,10 @@ export default class MutationBuffer {
       const shadowHost: Element | null = n.getRootNode
         ? (n.getRootNode() as ShadowRoot)?.host
         : null;
-      const notInDoc = !this.doc.contains(n) && !this.doc.contains(shadowHost);
+      // ensure shadowHost is a Node, or doc.contains will throw an error
+      const notInDoc =
+        !this.doc.contains(n) &&
+        (!(shadowHost instanceof Node) || !this.doc.contains(shadowHost));
       if (!n.parentNode || notInDoc) {
         return;
       }
@@ -359,13 +362,16 @@ export default class MutationBuffer {
       if (!node) {
         for (let index = addList.length - 1; index >= 0; index--) {
           const _node = addList.get(index)!;
-          const parentId = this.mirror.getId(
-            (_node.value.parentNode as Node) as INode,
-          );
-          const nextId = getNextId(_node.value);
-          if (parentId !== -1 && nextId !== -1) {
-            node = _node;
-            break;
+          // ensure _node is defined before attempting to find value
+          if (_node) {
+            const parentId = this.mirror.getId(
+              (_node.value.parentNode as Node) as INode,
+            );
+            const nextId = getNextId(_node.value);
+            if (parentId !== -1 && nextId !== -1) {
+              node = _node;
+              break;
+            }
           }
         }
       }
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -194,21 +194,43 @@ function initMoveObserver(
     },
     callbackThreshold,
   );
-  const updatePosition = throttle<MouseEvent | TouchEvent | DragEvent>(
+
+  // update position for mouse, touch, and drag events (drag event extends mouse event)
+  function handleUpdatePositionEvent(evt: MouseEvent | TouchEvent) {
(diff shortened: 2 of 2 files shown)


IMFIL:
@Yuyz0112 We use a fork of rrweb with this change and this PR fixed bugs that we saw happen to millions of our users.

src/record/mutation.ts
```diff
@@ -359,13 +362,16 @@ export default class MutationBuffer {
       if (!node) {
         for (let index = addList.length - 1; index >= 0; index--) {
           const _node = addList.get(index)!;
```
Juice10, line 364: Since we are no longer expecting `_node` to always have a value, let's remove the `!` which marks it as non-null.
```suggestion
          const _node = addList.get(index);
```

Juice10 (changes requested):

Juice10 (approved):

src/record/mutation.ts
```diff
@@ -271,7 +271,10 @@ export default class MutationBuffer {
       const shadowHost: Element | null = n.getRootNode
         ? (n.getRootNode() as ShadowRoot)?.host
         : null;
-      const notInDoc = !this.doc.contains(n) && !this.doc.contains(shadowHost);
+      // ensure shadowHost is a Node, or doc.contains will throw an error
+      const notInDoc =
+        !this.doc.contains(n) &&
+        (!(shadowHost instanceof Node) || !this.doc.contains(shadowHost));
```
Yuyz0112, line 277: just curious, what shadow host will be if it's not a Node?

src/record/mutation.ts
```diff
@@ -358,14 +361,17 @@ export default class MutationBuffer {
...
-          const _node = addList.get(index)!;
-          const parentId = this.mirror.getId(
-            (_node.value.parentNode as Node) as INode,
-          );
-          const nextId = getNextId(_node.value);
-          if (parentId !== -1 && nextId !== -1) {
-            node = _node;
-            break;
+          const _node = addList.get(index);
+          // ensure _node is defined before attempting to find value
```
Yuyz0112, line 365: I'm not sure when do you facing this problem, because I think the root cause has been fixed in https://github.com/rrweb-io/rrweb/commit/2c6be72f35f68513b0d511a37037018663cbc704

So we should not get a node with value null in this linked list anymore?

src/record/mutation.ts
dbseel: It can be null, see line 271

src/record/mutation.ts
dbseel: We encountered this problem using a version or rrweb that included that commit.

Yuyz0112: if so, I believe we should debugging into the issue because skip one node of the linked list can cause potential data loss.

eoghanmurray:
Why is the `typeof DragEvent` check done outside of the `updatePosition` function?  Why not do something like

        typeof DragEvent !== 'undefined' && evt instanceof DragEvent
          ? IncrementalSource.Drag
          : evt instanceof MouseEvent
          ? IncrementalSource.MouseMove
          : IncrementalSource.TouchMove;

?

I'm trying to reconcile this commit with https://github.com/rrweb-io/rrweb/pull/558/commits/97c36393cfc2e92bb8e86851cce5d142d65fdbc5 and the separation into two functions is leading to a lot of further code duplication.  The `evt instanceof DragEvent` clause seems to be the only bit that is different between `updateDragPosition` and `updatePosition`

dbseel:
> Why is the `typeof DragEvent` check done outside of the `updatePosition` function? Why not do something like
> 
> ```
>     typeof DragEvent !== 'undefined' && evt instanceof DragEvent
>       ? IncrementalSource.Drag
>       : evt instanceof MouseEvent
>       ? IncrementalSource.MouseMove
>       : IncrementalSource.TouchMove;
> ```
> 
> ?
> 
> I'm trying to reconcile this commit with [97c3639](https://github.com/rrweb-io/rrweb/commit/97c36393cfc2e92bb8e86851cce5d142d65fdbc5) and the separation into two functions is leading to a lot of further code duplication. The `evt instanceof DragEvent` clause seems to be the only bit that is different between `updateDragPosition` and `updatePosition`
> 
> [edit I've added #652 to show this idea]

They prototype of the call itself was causing problems when DragEvent wasn't defined. That is why the check is done ahead of the function call.

PR #638, opened 2021-07-17
use prettier instead of node-libtidy

Yuyz0112:
`node-libtidy` depends on node-pre-gyp to install binary, which seems [buggy](https://github.com/mapbox/node-pre-gyp/issues/367).

Since we already has prettier as our dependency, I think we can use prettier to format HTML in snapshot testing?


Yuyz0112:
@Juice10 Is this ok for you?

Juice10:
Much better, thanks for cleaning that up @Yuyz0112!

PR #643, opened 2021-07-22
Cache addHoverClass as it is quite expensive

Juice10:
Problem:
addHoverClass is one of the most CPU intensive things we do when rebuilding a snapshot. And especially when scrubbing backwards or browsing a recording by clicking around on the timeline addHoverClass gets repeatedly called.

Solution:
This PR caches the results of addHoverClass so they only have to be calculated once and after that they are fast.

### DIFF
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -123,7 +123,7 @@ function buildNode(
   options: {
     doc: Document;
     hackCss: boolean;
-    cache?: BuildCache;
+    cache: BuildCache;
   },
 ): Node | null {
   const { doc, hackCss, cache } = options;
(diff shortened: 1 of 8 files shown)


packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -99,20 +103,30 @@ export function addHoverClass(cssText: string): string {
...
+    stylesWithHoverClass,
+  };
 }
 
 function buildNode(
   n: serializedNodeWithId,
   options: {
     doc: Document;
     hackCss: boolean;
+    cache?: BuildCache;
```
Yuyz0112, line 126: shall we change this and the following code to non-optional params? This helps people(us) do not forget to pass the params.

Juice10: Agreed! I've updated the PR to make cache a non-optional parameter

Yuyz0112:
Let's go!

PR #651, opened 2021-08-03
Don't trust the html 'value' attribute on <option>

eoghanmurray:
The DOM `.selected` is the one we want I think.

 - encountered inconsistent html with multiple <option> elements with the same value attribute
 - due to check `attributes.selected = (n as HTMLOptionElement).selected;` thee extra ones were being stored as \'selected: false`; furthermore, the `false` value was being ignored upon replay, and so the last of the extra ones was being chosen as the 'selected' option


Yuyz0112:
good catch

PR #656, opened 2021-08-05
fix: fix console plugin's OOM problem

PeterChen1997:


Yuyz0112:
LGTM

maybe @Mark-Fenng can do a review either?

packages/rrweb/src/plugins/console/record/stringify.ts
```diff
@@ -48,6 +48,26 @@ function pathToSelector(node: HTMLElement): string | '' {
   return path;
 }
 
+/**
+ * judge the object's depth
+ */
+function isObjTooDeep(obj: any, limit: number): boolean {
```
Mark-Fenng, line 54: This solution can solve the issue well but I prefer the way to judge depth in the stringify function.

PeterChen1997:
@Mark-Fenng @Yuyz0112 fixed~

Mark-Fenng (approved):
This is really great now!

Issue #503, opened 2021-02-25
distinguish 'consequental' clicks (links etc.) vs. clicks/touches on blank space


eoghanmurray:
At the moment, the rrweb MouseInteraction events record a Click with a x/y co-ordinate but without reference to the underlying event that was clicked upon.
This could be retrieved during playback by calling [elementFromPoint](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementFromPoint) after the event is executed, but for my usecase — labelling/describing the clicks on the timeline — this is too late.  (I haven't tested whether elementFromPoint works or is accurate during playback — the `pointer-events:none` style on the rrweb container may hamper it).

Identifying the element would be great; it would also be great if we could identify whether the click/touch results in a page load (click on a link) or triggers some other javascript action (e.g. a modal popup)
There's also situations where e.g. a click on a link opens the link in a new background tab, so recording which meta keys are depressed would also be useful.
I'm not sure if these could be teased out, so I guess identifying the element and including it in the event would be a great first step.

I should be able to get around to implementing this but it's not high priority at the moment so just raising here in case there's prior work or other things to be concerned about.

Aside: some other commercial offerings label multiple clicks on blank space 'rage clicks', but personally I don't find this useful or accurate.

Yuyz0112:
> This could be retrieved during playback by calling elementFromPoint after the event is executed

```
type mouseInteractionParam = {
  type: MouseInteractions;
  id: number;
  x: number;
  y: number;
};
```

I think we are recording element id with mouse interactions?

eoghanmurray:
Ah sorry I should have read the code (or the output) more closely before posting, apologies!

I also see that it's impossible to always check if there's been a click event listener set on an element:
https://stackoverflow.com/questions/11455515/how-to-check-whether-dynamically-attached-event-listener-exists-or-not
so I guess I'll have to drop the detection of 'consequential' clicks.

So other than that, would it be okay to record the 'href' attribute if it exists on the target?  This would at least identify a traditional click on a link.

meatflavourdev:
> I also see that it's impossible to always check if there's been a click event listener set on an element:
> https://stackoverflow.com/questions/11455515/how-to-check-whether-dynamically-attached-event-listener-exists-or-not
> so I guess I'll have to drop the detection of 'consequential' clicks.

Can you explain this to a 5th grader? I am trying to understand your reasoning here.

eoghanmurray:
[...]

PR #659, opened 2021-08-09
Record href with click

eoghanmurray:
See #503 for my proposal on this.

The idea is that the actual href of an anchor is a useful distinguishing feature of a click.

I'm going for the low hanging fruit here, but if it's deemed to be a good idea, I could extend this PR with detecting `<button type="submit">` clicks and interrogating the associated `<form target="">` (we might need to also record POST/GET for that).

### DIFF
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -23,6 +23,8 @@ import {
   mousemoveCallBack,
   mousePosition,
   mouseInteractionCallBack,
+  mouseInteractionParam,
+  clickParam,
   MouseInteractions,
   listenerHandler,
   scrollCallback,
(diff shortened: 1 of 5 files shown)


meatflavourdev:
Right, the "consequence" of a click can be determined by the metadata of the click event is sort of a generalization of the idea here. Is there other data that might be relevant that should be recorded here? You mentioned buttons, etc.

Yuyz0112:
This code looks too 'analytics tool specific' for me. Obviously, we can have more of these data along with the core data structure for analytics usage, but other users do not care.

Not quite sure how users will consume these data but do a separate event can do the same thing?
```
[
  click event,
  custom click event(with href, button type, etc.),
]
```

If this works for the backend and replayer, I think it's a better implementation because we can move it to a plugin.

eoghanmurray:
The application on our end is to show a tooltip on the timeline so that users can inspect clicks.  It's also useful in the context of a multi-pageview session to see whether a click on one page resulted in a corresponding Meta event with the same url for the next page load. This can be useful if there are multiple tabs open at once in terms of joining up a session correctly.

> I think it's a better implementation because we can move it to a plugin

That's fine by me!  I'm also planning to add CSS selectors to click events (and maybe mutations also for debugging purposes) e.g. via https://github.com/antonmedv/finder — given the extra js include, it's definitely something that should be done via a plugin.   
I thought this PR might be more generally useful in the trunk given that it doesn't require extra javascript.  Also click events (on links) aren't all that common, so shouldn't add much to an average recording, but I think are just as significant as pageloads e.g. if the user is exiting the website.

Aside: I've also considered adding a slimDOM option to replace all href values with e.g. an 'X', as hrefs can contain long urls with bloated query strings which don't get rendered during replay (with the possible exception of a tailored CSS rule).  This PR would be necessary to preserve only the relevant href info in that case.

eoghanmurray:
> Is there other data that might be relevant?

I've just thought of whether SHIFT/META keys are held down during the click; that is significant in terms of opening the link in a new tab.

Yuyz0112:
Yes, the main concern is even we provide some options to not collecting these data, the code of these implementations still makes the rrweb's core code getting a bit more complicated.

So a standalone plugin may be helpful because it moves the code to a single place and decoupled with the core.

eoghanmurray:
I'll try to get time to implement this as a plugin, but am not exploring this at the moment so if anyone wants to take up the mantle, go ahead.

@Yuyz0112 00fd6eb0037642bcb88c33199b539a898a6f5435 might otherwise be useful as it is independent of this PR

Yuyz0112:
@eoghanmurray , yes, we can pick the test commit.

eoghanmurray:
https://github.com/rrweb-io/rrweb/commit/00fd6eb0037642bcb88c33199b539a898a6f5435 has been merged now in https://github.com/rrweb-io/rrweb/commit/3ce6903122b039f9fb200c9a6130321772cac117 as it was useful in #1201 (which fixes up #1129 )

PR #661, opened 2021-08-10
Rename `packages/rrweb-player` to `packages/rrweb-playback-ui`

eoghanmurray:
- avoid confusion with `packages/rrweb/src/replayer` which is the generic player

I don't know what further work would be needed on the publishing side of things to get e.g. new url after this change https://cdn.jsdelivr.net/npm/rrweb-playback-ui@latest/dist/index.js working?


Yuyz0112:
> quote from the core team discussion, so others can get more information:

I agree with the naming of `rrweb/src/replayer` and `rrweb-player` is confusing sometimes.

From the use case, I think:
- `rrweb/src/replayer` is the minimum core with most of the replay features.
- `rrweb-player` is the player with a user-friendly UI. So maybe `rrweb-ui` is a better name?

And I think `rrweb-player` will be the only UI package maintained by us, this means:
- we may have UI packages with frameworks like React, Vue, Angular, but is a wrapper of `rrweb-player`, like [this](https://codesandbox.io/s/competent-keldysh-m65dt?file=/src/App.js)
- community can maintain UI packages implemented on top of `rrweb/src/replayer` instead of `rrweb-player`, like [this](https://github.com/Preflight-HQ/rrweb-player-vue)
- the way of customizing UI can be [various](https://github.com/rrweb-io/rrweb/blob/master/docs/recipes/customize-replayer.md).
- the tech stack of rrweb-player is under the water, so maybe we can switch to other frameworks in the future version of `rrweb-player`, without breaking changes in naming and version.

And we choose svelte because it has a slim runtime which reduces the overhead when using in other frameworks. Users can just use `rrweb-player` as a vanilla JS UI without caring it's implemented in svelte or plain JS.

The reason we are not using plain JS is that we need a reactive data model and component system.

eoghanmurray:
@Yuyz0112 wondering if you prefer `rrweb-ui` over `rrweb-playback-ui` as I don't think I had updated this change to the latter when you made your comment above?

eoghanmurray:
Parking this as we are going to tie the rename of this package with a new release with improve features and greater ability to customize.

eoghanmurray:
This is a 'nice to have' if we ever rearrange all the packages.

Juice10:
Since we've moved a lot of the packages, I think we should live in `packages/playback-ui`, and `@rrweb/playback-ui` as the package name.
It better reflects the new naming of the packages

PR #662, opened 2021-08-11
The desktop pointer cursor is not representative of mobile

eoghanmurray:
The desktop pointer cursor is not representative of what is happening on a mobile device.

Instead, check a recording for any presence of a Touch event, and switch to a touch visualisation mode for the entire recording.
(for now, we use this mode even for mixed touch/mouse devices - this could be improved upon in future)

Show a round circle representing the users' finger which is visible only between TouchStart and TouchEnd events
Again this can be evolved upon, but this change should be a good start in the right direction.

### DIFF
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -142,6 +142,7 @@ export class Replayer {
 
     this.handleResize = this.handleResize.bind(this);
     this.getCastFn = this.getCastFn.bind(this);
+    this.applyEventsSynchronously = this.applyEventsSynchronously.bind(this);
     this.emitter.on(ReplayerEvents.Resize, this.handleResize as Handler);
 
     this.setupDom();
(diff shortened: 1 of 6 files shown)


Juice10 (approved):

packages/rrweb/src/replay/index.ts
```diff
@@ -115,6 +123,8 @@ export class Replayer {
 
   private newDocumentQueue: addedNodeMutation[] = [];
 
+  private mouseState: mouseState | null = null;
```
Yuyz0112, line 126: The code to maintain this state looks a bit complex. Is it possible to use a getter to compute whenever need?

eoghanmurray: It's supposed to be a lightweight object to hold mouse x/y and whether touch is active or not. The extra `id` (and debugData) is for `moveAndHover`.
It's only relevant during application of 'sync' events, to maintain a record of the last change to mouse state, and is cleared after each batch.
I'm not sure what a getter would achieve?

eoghanmurray: Ok actually some of the complexity comes from me trying to wrap two separate things into the same state object.  I've separated them out in 5606987242251b40ea3e758cf93b21fe3f096d24 and this should make things clearer.

Yuyz0112:
Good Job!

PR #664, opened 2021-08-12
Batch apply synchronous events

eoghanmurray:
- this is a move towards a new approach of how we 'fast forward' through events, i.e. the behaviour with `sync = true`
 - the new `applyEventsSynchronously` function could be the place we'd initialize an rrdom object and apply mutation updates to it
 - there may be further optimizations enabled by `applyEventsSynchronously` e.g. executing `applyScroll` only once per element for a batch of events

### DIFF
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -142,6 +142,7 @@ export class Replayer {
 
     this.handleResize = this.handleResize.bind(this);
     this.getCastFn = this.getCastFn.bind(this);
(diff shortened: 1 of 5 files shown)


Juice10:
LGTM! Seems like the tests are failing due to a flaky network connection. You can re-run them by adding an extra commit (tweaking some whitespace, adding more documentation, whatever you'd like)

Juice10 (approved):

eoghanmurray:
This has been merged now as it was also part of #662

PR #665, opened 2021-08-13
Record and replay nested stylesheet rules

Juice10:
CSS can have nested rules thanks to @media queries and other [at rules](https://developer.mozilla.org/en-US/docs/Web/API/CSSGroupingRule). 

It is even possible to nest media queries in media queries, I tried it a couple levels and haven't found a limit.

It seems like https://stitches.dev uses mutations to @media rules quite heavily which means that currently recordings of sites built with Stitches feel very broken.

This pr add support for mutations to those nested rules.

### DIFF
--- a/docs/observer.md
+++ b/docs/observer.md
@@ -1,4 +1,5 @@
 # Incremental snapshots
+
 After completing a full snapshot, we need to record events that change the state.
 
 Right now, rrweb records the following events (we will expand upon this):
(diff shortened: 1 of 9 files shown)


Yuyz0112:
@Juice10 Nice work! There are some small TS types that need an update to pass the CI.

Yuyz0112:
picked in https://github.com/rrweb-io/rrweb/pull/666

PR #693, opened 2021-09-09
Avoid runtime error when CSSGroupingRule is not supported

### LINKED ISSUE #690: CSSGroupingRule not supported on Safari
Safari on desktop and mobile [doesn't support CSSGroupingRule](https://caniuse.com/?search=CSSGroupingRule). This is causing errors when recording on Safari.

```
ReferenceError: Can't find variable: CSSGroupingRule
```

The error is happening [on this line](https://github.com/rrweb-io/rrweb/blob/5dcbfa530f3a1c6d7cadaba076be25e4b263de39/packages/rrweb/src/record/observer.ts#L524).

Not sure what the pattern for handling browser API support. The obvious fix is to check if `CSSGroupingRule` is defined before monkey patching the methods. This would silence the error but would cause the recording to be inaccurate.

Any suggestions @Yuyz0112 and @Juice10 ?

Monchi:
ad-hoc fix for #690.

### DIFF
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -60,6 +60,8 @@ type WindowWithAngularZone = Window & {
 
 export const mutationBuffers: MutationBuffer[] = [];
 
+const isCSSGroupingRuleSupported = typeof CSSGroupingRule !== "undefined"
+
 function getEventTarget(event: Event): EventTarget | null {
   try {
     if ('composedPath' in event) {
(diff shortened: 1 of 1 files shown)


Juice10 (approved):
This PR fixes the safari error for #690 but doesn’t re-add support for nested styles to safari, so we’ll have to add that in a separate PR. More info: https://github.com/rrweb-io/rrweb/issues/690#issuecomment-912940633

Monchi:
@Juice10 Yes, this is just an ad-hoc patch to avoid runtime errors!

Yuyz0112:
Thanks!

PR #696, opened 2021-09-10
Keep blocked root elements as placeholders

Juice10:
According to [our documentation](https://github.com/rrweb-io/rrweb/blame/33010388695cc82942557a0360692ef8f966109f/guide.md#L166):

> An element with the class name .rr-block will not be recorded. Instead, it will replay **as a placeholder with the same dimension**.

Currently incremental snapshots are completely removing these blocked elements from being recorded. 
And in full snapshots these blocked elements are stripped of their children and turned into placeholders.

This PR makes that behaviour consistent between different types of snapshots and mutations.

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -262,9 +262,6 @@ export default class MutationBuffer {
         ns = ns && ns.nextSibling;
         nextId = ns && this.mirror.getId((ns as unknown) as INode);
       }
-      if (nextId === -1 && isBlocked(n.nextSibling, this.blockClass)) {
-        nextId = null;
-      }
       return nextId;
     };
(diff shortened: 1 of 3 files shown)


packages/rrweb/src/record/mutation.ts
```diff
@@ -604,7 +595,11 @@ export default class MutationBuffer {
       this.addedSet.add(n);
       this.droppedSet.delete(n);
     }
-    n.childNodes.forEach((childN) => this.genAdds(childN));
+
+    // if this node is blocked `serializeNode` will turn it into a placeholder element
+    // but we have to remove it's children
```
Juice10, line 600:
```suggestion
    // but we have to remove it's children otherwise they will be added as placeholders too
```

Yuyz0112:
Thanks and LGTM!~

PR #697, opened 2021-09-13
Record when a doc is in `compatMode` and trigger this mode upon replay

eoghanmurray:
https://developer.mozilla.org/en-US/docs/Web/API/Document/compatMode

 the included DOCTYPE was picked up from https://stackoverflow.com/questions/18976213/ - there may be better ways of triggering compatMode.


Yuyz0112:
Thanks! Going to merge this.

PR #704, opened 2021-09-15
improve config definition in the console plugin

Mark-Fenng:
One user encountered a type problem(https://rrweb.slack.com/archives/C01BYDC5C93/p1631683830050900). It turns out that definition of LogReplayConfig isn't perfect


Yuyz0112:
LGTM!

PR #707, opened 2021-09-23
Set the blockClass element's background color to the inherited color

JohnPhamous:
Changes an element that has the `blockClass` from having a background color of `#ccc` to `currentColor`. Setting it to `currentColor` will set the background color to the same value as `color`.

I think this is a nicer default because visually, this will work across all color themes.


Yuyz0112 (approved):

Yuyz0112:
Thanks!

PR #717, opened 2021-10-05
[RRDom] tagNames to Upper, tests and tooling

Juice10:
- Set tagNames to upper case by default
- Add vscode workspace file: https://code.visualstudio.com/docs/editor/workspaces
- VSCode settings for Jest and rrdom's tests
- Add basic test for RRDocument.buildFromDom


Mark-Fenng:
great unit test and workplace settings

PR #718, opened 2021-10-05
VSCode Workspace

Juice10:
Adds vscode workspace https://code.visualstudio.com/docs/editor/workspaces
In vscode if you go to File > Open Workspace and open the .vscode/monorepo.code-workspace your files will be split into different sections based on what package they fall under:

image

This also include rrdom (which will be empty for most people until we merge the rrdom branch)

### DIFF
--- a/.vscode/monorepo.code-workspace
+++ b/.vscode/monorepo.code-workspace
@@ -17,5 +17,12 @@
       "path": "../packages/rrweb-snapshot"
     }
   ],
-  "settings": {}
+  "settings": {
+    "jest.disabledWorkspaceFolders": [
+      "Monorepo",
+      "rrweb",
+      "rrweb-snapshot",
+      "rrweb-player"
+    ]
+  }
 }
--- a/.vscode/monorepo.code-workspace
+++ b/.vscode/monorepo.code-workspace
@@ -1,0 +1,21 @@
+{
+  "folders": [
+    {
+      "name": "Monorepo",
+      "path": ".."
+    },
+    {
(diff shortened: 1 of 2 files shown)


.vscode/monorepo.code-workspace
```diff
@@ -0,0 +1,28 @@
+{
+  "folders": [
+    {
+      "name": "Monorepo",
+      "path": ".."
+    },
+    {
+      "path": "../packages/rrdom"
```
Yuyz0112, line 8: will this cause a problem since rrdom has not been merged yet? or the plugin just ignore it correctly.

Juice10: It puts an `!` behind rrdom to indicate the folder doesn't exist.
image

I could have left it out but otherwise if you switch to the rrdom branch you have to close the workspace and re-open it again to pick up rrdom, which is annoying.

.vscode/monorepo.code-workspace
```diff
@@ -0,0 +1,28 @@
+{
+  "folders": [
+    {
+      "name": "Monorepo",
```
Juice10, line 4: It might make more sense to call this something other than "Monorepo", but it needs to be distinct enough from "rrweb" so people don't mix them up.

Alternatively we can go for explicit names:
rrweb monorepo `/`
rrweb package `/packages/rrweb`
rrweb-player package `/packages/rrweb-player`
rrweb-snapshot package `/packages/rrweb-snapshot`

Yuyz0112:
Thanks!

PR #724, opened 2021-10-07
Preserve iframe node id through loading callback

eoghanmurray:
Ensure the iframe id survives the onceIframeLoaded process; for some reason I'm seeing the `__sn` attribute on the iframe DOM element disappearing with the callback

### DIFF
--- a/packages/rrweb/src/record/iframe-manager.ts
+++ b/packages/rrweb/src/record/iframe-manager.ts
@@ -55,12 +55,13 @@ export class IframeManager {
 
   public attachIframe(
     iframeEl: HTMLIFrameElement,
+    parentId: number,
     childSn: serializedNodeWithId,
   ) {
     this.mutationCb({
       adds: [
         {
-          parentId: this.mirror.getId(iframeEl),
+          parentId: parentId,
           nextId: null,
           node: childSn,
         },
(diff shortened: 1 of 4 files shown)


packages/rrweb/src/record/iframe-manager.ts
```diff
@@ -18,11 +18,11 @@ export class IframeManager {
     this.loadListener = cb;
   }
 
-  public attachIframe(iframeEl: INode, childSn: serializedNodeWithId) {
+  public attachIframe(iframeEl: INode, parentId: number, childSn: serializedNodeWithId) {
     this.mutationCb({
       adds: [
         {
-          parentId: iframeEl.__sn.id,
+          parentId: parentId,
```
Juice10, line 25: This is slightly neater
```suggestion
          parentId,
```

Juice10 (dismissed):
LGTM!

eoghanmurray:
So the circumstances of this PR were that the iframe element was 'somehow' losing it's `__sn` DOM attribute.

We no longer store the `__sn` in the DOM but rather retrieve the id from the mirror.

So now the question would be whether the identity of the iframe has changed such that in that scenario, the mirror wouldn't be able to retrieve it.   I don't know as I don't have the scenario available anymore.

So we could merge as a precaution; what do you think @Juice10  ?

Juice10:
@eoghanmurray we don't have a test case for this. And the underlying code has changed pretty heavily. I'm thinking we close this till it becomes a problem again and resurrect it then (with test case). But I'm not 100% on this one

PR #725, opened 2021-10-07
don't reference old rrweb-snapshot 

eoghanmurray:
Ensure an `npm update` in this folder doesn't cause an old version of rrweb-snapshot package to be installed, which will take precedence over the monorepo one


eoghanmurray:
Thanks for the input, I'm totally at sea on this.
How else should I have installed `jest` except using npm?
Is there a way to prevent npm running then if it's incompatible?

Juice10:
Just changed the title to reflect the new reason for this PR based on the conversation between @eoghanmurray and @Yuyz0112

PR #748, opened 2021-11-03
test(rrdom): add snapshot for document.test.ts

Mark-Fenng:
Add a html file for testing RRDocument and make a snapshot of printed rrdom.
I'm not sure if this is a good way, so it may need some review.


Juice10 (approved):
I love it @Mark-Fenng, great work!

PR #760, opened 2021-11-26
Add option to block animation on <title> tag

eoghanmurray:
These can generate massive recordings on some websites (think scrolling title tag)

### DIFF
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -174,7 +174,7 @@ function record<T = eventWithTime>(
           // as they destroy some (hidden) info:
           headMetaAuthorship: _slimDOMOptions === 'all',
           headMetaDescKeywords: _slimDOMOptions === 'all',
-          headTitleMutations: _slimDOMOptions === 'all', // block title tag 'animation' which can generate a lot of mutations that aren't usually displayed in replayers
+          headTitleMutations: _slimDOMOptions === 'all',
         }
       : _slimDOMOptions
       ? _slimDOMOptions
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -174,6 +174,7 @@ function record<T = eventWithTime>(
           // as they destroy some (hidden) info:
           headMetaAuthorship: _slimDOMOptions === 'all',
           headMetaDescKeywords: _slimDOMOptions === 'all',
+          headTitleMutations: _slimDOMOptions === 'all', // block title tag 'animation' which can generate a lot of mutations that aren't usually displayed in replayers
         }
       : _slimDOMOptions
       ? _slimDOMOptions
--- a/.changeset/title-deanimate-option.md
+++ b/.changeset/title-deanimate-option.md
@@ -1,0 +1,6 @@
+---
+"rrweb-snapshot": patch
+"rrweb": patch
+---
+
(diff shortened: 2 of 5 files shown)


jlalmes:
I am making use of the `<title>` in recordings. I think it's fine to offer the option to ignore changes, but maybe removing changes from recordings by default when `slimDOMOptions` is set to `all` is an unintended breaking change?

Use case - when recording SPAs I would like the `<title>` value to update when navigating from page to page.

Yuyz0112:
good point @jlalmes

eoghanmurray:
Great spot, thanks @jlalmes  ... in light of that, maybe it should be a throttling thing?
In the mean time I'll update so that you have to explicitly turn this on, rather than including it with `all` as you suggested.

eoghanmurray:
Actually @jlalmes, the intention of the `'all'` is to enable all slim dom options — there is also a `true` option which is there to enable all 'safe' options (for some definition of 'safe').
Do you think you should change to using `true` or is that difference non-obvious?

This also reminds me that I've got some code external to rrweb which detects Single Page App url changes, but only if they happen in tandem with a <title> tag change.  If I were to integrate that in rrweb it would introduce a new event, analogous to the `Meta` event to record SPA navigation.  (this wouldn't get recorded if just the URL changed).  Would that be useful?

jlalmes:
> Do you think you should change to using true or is that difference non-obvious?

Good idea! I will look into that.

> This also reminds me that I've got some code external to rrweb which detects Single Page App url changes, but only if they happen in tandem with a <title> tag change. If I were to integrate that in rrweb it would introduce a new event, analogous to the Meta event to record SPA navigation. (this wouldn't get recorded if just the URL changed). Would that be useful?

Yes, a `PageNavigation` event would be really useful I think. I imagine we would want this event to fire even if the `<title>` tag doesn't change, however, I'd be interested to take a look at your implementation of this & collaborate on a PR.

jlalmes (approved):

packages/rrweb/src/record/index.ts
```diff
@@ -110,6 +110,7 @@ function record<T = eventWithTime>(
           // as they destroy some (hidden) info:
           headMetaAuthorship: _slimDOMOptions === 'all',
           headMetaDescKeywords: _slimDOMOptions === 'all',
+          headTitleMutations: _slimDOMOptions === 'all',  // block title tag 'animation' which can generate a lot of mutations that aren't usually displayed in replayers
```
Juice10, line 113: This key/value should also get added to the SlimDOMOptions type.
The comment you've added here would make a perfect JS/TSDoc for that type.

eg.
```ts
/**
 * blocks title tag 'animations' which can generate a lot of mutations that aren't usually displayed in replayers
 **/
headTitleMutations: boolean;
```

eoghanmurray: thanks, I've made that change and rebased upon master ... function sig of `isIgnored` has changed since I authored this PR

eoghanmurray (commented):
I thought I satisfied all the requested changes, not sure how to indicate that!

eoghanmurray:
OK I guess my force push messed things up, as it can't find the commits the requested changes were attached to!

Juice10:
This will fix the issues with CI not passing on this PR: https://github.com/rrweb-io/rrweb/pull/1500

Juice10 (approved):

PR #762, opened 2021-12-02
Small regex fix for older browsers

croqaz:
Hi guys!

I made a tiny change to the regexes to allow rrWeb to run on older browsers.
The issue is: `"can't supply flags when constructing one RegExp from another"`, so I'm fixing it here.
I hope you'll accept it, I'm already using it in a custom browser.

Thanks!


Yuyz0112:
Thanks!

PR #768, opened 2021-12-13
fix: The URL involved in the dynamic modification element inline style is not converted to an absolute path

champkeh:


Yuyz0112:
Maybe we need transform all the style attribute here?

champkeh:
Yes, indeed. I'll update it later.

eoghanmurray:
Is this to fix that there are URLs in the `background-image` attribute?

- does `transformAttribute` handle css shorthand e.g. `background: black url(/xyz.jpg) no-repeat;`
- should (can) we do this on the replayer side?
- is it possible to write a test for this?

champkeh:
@eoghanmurray I write some test case for this.

eoghanmurray:
Thanks for the test case!

Is this going down the `if (name === 'style' && value)` code path in `transformAttribute` ?
I think in that case, calling `absoluteToStylesheet` directly is preferable as we are not actually dealing with an entire attribute here.  

Is it necessary to update your test case with a shorthand one; `background: black url(/xyz.jpg) no-repeat;` or can we trust that `absoluteToStylesheet` will do the right thing?

packages/rrweb/src/record/mutation.ts
```diff
@@ -490,12 +490,18 @@ export default class MutationBuffer {
...
           const styleObj = item.attributes.style as styleAttributeValue;
           for (const pname of Array.from(target.style)) {
-            const newValue = target.style.getPropertyValue(pname);
+            let newValue = target.style.getPropertyValue(pname);
             const newPriority = target.style.getPropertyPriority(pname);
             if (
               newValue !== old.style.getPropertyValue(pname) ||
               newPriority !== old.style.getPropertyPriority(pname)
             ) {
+              newValue = transformAttribute(
```
eoghanmurray, line 499: could call `absoluteToStylesheet` directly here on `newValue`

eoghanmurray: like `newValue = attributeToStylesheet(newValue);`

eoghanmurray: (would need to import attributeToStylesheet and also maybe change it's href attribute to default to getHref() rather than importing getHref also)

packages/rrweb/src/record/mutation.ts
```diff
@@ -490,12 +490,18 @@ export default class MutationBuffer {
           }
           const styleObj = item.attributes.style as styleAttributeValue;
           for (const pname of Array.from(target.style)) {
-            const newValue = target.style.getPropertyValue(pname);
+            let newValue = target.style.getPropertyValue(pname);
```
eoghanmurray, line 493: Thought we might need to do the transformation here, but you are right to wait til after the comparison.

packages/rrweb/test/html/absolute-image-url.html
```diff
@@ -0,0 +1,20 @@
...
+  <title>absolute image url</title>
+  <script src="../../../rrweb-snapshot/dist/rrweb-snapshot.js"></script>
+</head>
+<body>
+<div id='div1' style='height: 200px; background-image: url("static.png")'></div>
+<div id='div2' style='height: 200px;'></div>
+<button onclick='addDynamicBackgroundImage()'>add dynamic background image</button>
+<script>
+  function addDynamicBackgroundImage() {
+    document.querySelector('#div2').style.setProperty('background-image', 'url(dynamic.png)')
```
eoghanmurray, line 16: The images are cute!  However they do need to be downloaded by everyone who downloads rrweb ... so they can be removed without affecting the test case (it shouldn't matter for the purposes of the test whether they actually exist)

eoghanmurray (changes requested):

Issue #737, opened 2021-10-22
Snapshot images as base64


croqaz:
Hi!

First of all awesome project, thanks for making it public!

I have a question regarding saving images offline as base64, but also keeping the original URL link.
Right now an image is stored something like:

```js
{'type': 2,
 'tagName': 'img',
 'attributes': {'id': 'mainImage', 'src': 'https://whatever.com/something.gif'},
 'childNodes': [],
 'id': 29},
```

Ideally the image content would be stored somewhere in there like: `src="data:image/gif;base64,deadbeef ...`

There's an option to save the canvas like that.

Is it possible for the images as well?

champkeh:
This could be cause result storage very big. 
My current method is to store images separately in Minio based on their content hash at project build time, then in the playback phase, dynamically proxy the address of the image to Minio's url.

croqaz:
Thanks for your response!

I think storage is not such a big issue if you really want a complete, portable and self-container snapshot with everything.
Saving the image is not that complicated, the problem is I don't know any way of doing it in a sync mode. So this means all the functions will have to be converted to async.
Eg: https://stackoverflow.com/a/42508185

After saving the base64 in the `rr_dataURL'`, you only need to update the rebuild.ts here:

```js
// from
if (tagName === 'canvas' && name === 'rr_dataURL') {
// to
if ((tagName === 'canvas' || tagName === 'img') && name === 'rr_dataURL') {
```

champkeh:
Oh, I see what you mean，you can create a global canvas element as a render service, then in `serializeNode` function add follow code:

```js
const canvasService: HTMLCanvasElement = document.createElement('canvas');
const canvasCtx: CanvasRenderingContext2D | null = canvasService.getContext('2d');

function serializeNode(n, options) {

  // ...

  // convert image to base64
  if (tagName === 'img' && canvasCtx) {
    canvasService.width = (n as HTMLImageElement).naturalWidth;
    canvasService.height = (n as HTMLImageElement).naturalHeight;
    canvasCtx.drawImage((n as HTMLImageElement), 0, 0);
    attributes.rr_dataURL = canvasProvider.toDataURL();
  }

  // ...
}
```

and i test that in local environment, result is follow, it seems feasible:

![image](https://user-images.githubusercontent.com/7316006/142144273-27143099-f10b-4a9b-b5d7-62a09390448d.png)

croqaz:
Your solution is really cool and it should work! Thanks for sharing!!

[...]

PR #770, opened 2021-12-14
Save images offline, in the snapshot

croqaz:
Hi guys,

I made this feature that I already discussed here: https://github.com/rrweb-io/rrweb/issues/737
It's about saving images inside the snapshot as base64, just like we have for canvas already.

I tried to use the same code style and not change things too much.
I added a test, but I'm not sure if it's in the right place. In the integration, it seemed the best place, because you're using Puppeteer and we need to simulate a canvas to make it work.

Please let me know if you think something is wrong, or you want more tests, or I'm missing something.


packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -226,18 +226,24 @@ function buildNode(
                 ctx.drawImage(image, 0, 0, image.width, image.height);
               }
             };
+          } else if (tagName === 'img' && name === 'rr_dataURL') {
+            const image = (node as HTMLImageElement);
+            if (!image.currentSrc.startsWith('data:')) {
+              // backup original img src
+              image.setAttribute('data-src', image.currentSrc);
```
croqaz, line 233: Here I used the `data-src` attribute for keeping the original URL.
Maybe we can use something else, less likely to conflict, maybe `data-rr-src` ? or `data-rrweb-src` ?
What do you think?

croqaz: OK I just went ahead and renamed this to `data-rrweb-src`.

guide.md
```diff
@@ -155,6 +155,7 @@ The parameter of `rrweb.record` accepts the following options.
 | packFn               | -                  | refer to the [storage optimization recipe](./docs/recipes/optimize-storage.md)                                                                                                                |
 | sampling             | -                  | refer to the [storage optimization recipe](./docs/recipes/optimize-storage.md)                                                                                                                |
 | recordCanvas         | false              | whether to record the canvas element                                                                                                                                                          |
+| recordImages         | false              | whether to record the image content                                                                                                                                                          |
```
Yuyz0112, line 158: maybe inlineImages or something more concrete?

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -496,6 +512,15 @@ function serializeNode(
       if (tagName === 'canvas' && recordCanvas) {
         attributes.rr_dataURL = (n as HTMLCanvasElement).toDataURL();
       }
+      // save image offline
+      if (tagName === 'img' && recordImages && canvasService && canvasCtx) {
```
Yuyz0112, line 516: To avoid some corner cases in the browser, especially some security options with cross-origin domains, how about adding a try-catch to this block? So even though corner cases happen, it will not block the recording.

Yuyz0112 (changes requested):
Very cool feature! Thanks!

guide.md
croqaz: Sure! Done!

packages/rrweb-snapshot/src/snapshot.ts
croqaz: Not sure if I understood where you want the try catch, I added it a little below.
See https://github.com/rrweb-io/rrweb/pull/770/commits/cec7fdae383e5cd8ddb9686d7c4fb264a519e217#diff-6434d77e679a3d1e4217dcad03ee6f02551e41dc0b077a06a45020e68e1b4ee3R519-R526

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -513,13 +513,17 @@ function serializeNode(
...
-        canvasService.height = image.naturalHeight;
-        canvasCtx.drawImage(image, 0, 0);
-        attributes.rr_dataURL = canvasService.toDataURL();
+        try {
+          canvasService.width = image.naturalWidth;
+          canvasService.height = image.naturalHeight;
+          canvasCtx.drawImage(image, 0, 0);
+          attributes.rr_dataURL = canvasService.toDataURL();
+        } catch {
+          // ignore error
```
lopuhin, line 525: shall we log something here, e.g. with ``console.warn``?

Yuyz0112:
Sorry for the late, let's go!

eoghanmurray:
@croqaz could you check out #1468 for me?  This builds on top of subsequent work in #822

PR #776, opened 2021-12-18
Fix animations on pseudo-elements not pausing when replayer is paused

### LINKED ISSUE #775: CSS animations applied to pseudo-elements are not paused when player is paused
Currently we have this rule injected into the player:

```css
 'html.rrweb-paused * { animation-play-state: paused !important; }',
```

This is working as expected. There's an edge case for when an animation is applied to a pseudo-element (`::before` and `::after`).

JohnPhamous:
Fixes #775


Yuyz0112:
Thanks!

PR #784, opened 2021-12-22
Update css-font-loading-module

gabrielstellini:
Updates css-font-loading-module from `0.0.4` to `0.0.7`.

### DIFF
--- a/packages/rrweb/src/types.ts
+++ b/packages/rrweb/src/types.ts
@@ -8,7 +8,6 @@ import {
   MaskTextFn,
 } from 'rrweb-snapshot';
 import { PackFn, UnpackFn } from './packer/base';
-import { FontFaceDescriptors } from 'css-font-loading-module';
 import { IframeManager } from './record/iframe-manager';
 import { ShadowDomManager } from './record/shadow-dom-manager';
 import type { Replayer } from './replay';
--- a/packages/rrweb/package.json
+++ b/packages/rrweb/package.json
@@ -71,7 +71,7 @@
     "typescript": "^3.9.5"
   },
   "dependencies": {
-    "@types/css-font-loading-module": "0.0.4",
+    "@types/css-font-loading-module": "0.0.7",
     "@xstate/fsm": "^1.4.0",
     "fflate": "^0.4.4",
     "mitt": "^1.1.3",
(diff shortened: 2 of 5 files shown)


packages/rrweb/src/types.ts
```diff
@@ -8,7 +8,6 @@ import {
   MaskTextFn,
 } from 'rrweb-snapshot';
 import { PackFn, UnpackFn } from './packer/base';
-import { FontFaceDescriptors } from 'css-font-loading-module';
```
Juice10, line 11: Why are you removing the `FontFaceDescriptors` import? 

Looks like it would break the code since `FontFaceDescriptors` is still being referenced later on in the file: https://github.com/rrweb-io/rrweb/blob/c70b32870e597b99b8db6d20fc012f977dea8c21/packages/rrweb/src/types.ts#L445

Juice10 (changes requested):

packages/rrweb/src/types.ts
jtheberge: Can we use typescript's DOM lib definition here instead to avoid interface duplicate naming conflicts?
https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L491
Assuming that's why it can be safely deleted here.
Relevant issue: #813

gabrielstellini: @jtheberge If it's urgent, I recommend using `patch-package` to fix this issue (temporarily). Here's the contents of my  `patches/rrweb+1.0.7.patch` file:

```
diff --git a/node_modules/rrweb/typings/types.d.ts b/node_modules/rrweb/typings/types.d.ts
index 7d9c049..4bfd2ea 100644
--- a/node_modules/rrweb/typings/types.d.ts
+++ b/node_modules/rrweb/typings/types.d.ts
@@ -1,6 +1,5 @@
 import { serializedNodeWithId, idNodeMap, INode, MaskInputOptions, SlimDOMOptions, MaskInputFn, MaskTextFn } from 'rrweb-snapshot';
 import { PackFn, UnpackFn } from './packer/base';
-import { FontFaceDescriptors } from 'css-font-loading-module';
 import { IframeManager } from './record/iframe-manager';
 import { ShadowDomManager } from './record/shadow-dom-manager';
 import type { Replayer } from './replay';
```

jtheberge: ah nice idea, hadn't heard of this before but appreciate the info!

Yuyz0112:
Thanks, going to merge this and have a try by myself.

PR #786, opened 2021-12-23
Add option to spoof URL in repl

Juice10:
Really helps when experiencing CORS issues or when loading images in webgl.

Interface looks like this:
image
When you select "Start replay on original url (helps when experiencing CORS issues)" it opens a browser window that spoofs the original url and bypasses any CORS issues you would otherwise have in the replayer.
image


Yuyz0112:
LGTM

PR #795, opened 2022-01-12
Fixes and tweaks

croqaz:
This PR is a short continuation of https://github.com/rrweb-io/rrweb/pull/770 , to add 2 log messages.
Also fix a non-obvious error, when `cssText = getCssRulesString(...)` crashes because stylesheet=null and snapshot function doesn't return. It happens sometimes: https://github.com/rrweb-io/rrweb/issues/764#issuecomment-993091039
~Also removed `CSSStyleSheet.rules` which is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet#legacy_properties~


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -523,8 +527,8 @@ function serializeNode(
           canvasService.height = image.naturalHeight;
           canvasCtx.drawImage(image, 0, 0);
           attributes.rr_dataURL = canvasService.toDataURL();
-        } catch {
-          // ignore error
+        } catch (err) {
+          console.warn(`Cannot inline image: ${image}! Error: ${err}`);
```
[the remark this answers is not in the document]
croqaz: Hi, I added these warnings because they were agreed in the previous PR and they don't do any harm :) If you run the snapshot in a real browser, you can see them and they are useful. If you run them in a remote Puppeteer, you just ignore them and that's also great.

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -41,7 +41,8 @@ function getValidTagName(element: HTMLElement): string {
 
 function getCssRulesString(s: CSSStyleSheet): string | null {
   try {
-    const rules = s.rules || s.cssRules;
+    // handle reading remote CSS rules
+    const rules = s.cssRules;
```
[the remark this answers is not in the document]
croqaz: I'm not sure when the ".rules" were deprecated. Do you have a real usecase, when a browser doesn't have the new ".cssRules", so you actually need the old property?
We can have both, but how about we reverse the order: `s.cssRules || s.rules` because it's more likely that the new property works?

croqaz:
Hi @Mark-Fenng ! Thanks for taking a look at the PR.
I noticed you fixed some issues with the inline image, but there's a try catch that I think can be improved a bit, so I hope you don't mind I moved it into the new function `recordInlineImage` that you have created: https://github.com/rrweb-io/rrweb/pull/795/commits/4089127fa81d07c82d26f9227284b1d0e87e06b5
I believe it doesn't make sense to try catch the function definition, so it should be merged.

Feel free to suggest any changes, I'm open :)

packages/rrweb-snapshot/src/snapshot.ts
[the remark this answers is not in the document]
Mark-Fenng: https://caniuse.com/?search=CSSStyleSheet.rules
You can see that "CSSStyleSheet.rules" has better compatibility and cssRules actually won't be executed.

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -510,24 +514,24 @@ function serializeNode(
           canvasCtx = canvasService.getContext('2d');
         }
         const image = n as HTMLImageElement;
-        const oldValue = image.crossOrigin;
-        image.crossOrigin = 'anonymous';
-        try {
-          const recordInlineImage = () => {
+        const recordInlineImage = () => {
+          const oldValue = image.crossOrigin;
+          image.crossOrigin = 'anonymous';
```
Mark-Fenng, line 519: const oldValue = image.crossOrigin;
image.crossOrigin = 'anonymous';
These two lines shouldn't be included in the function because `image.crossOrigin = 'anonymous';` may leave the image in an unloaded state. So they should be evaluated before the condition judgment.

packages/rrweb-snapshot/src/snapshot.ts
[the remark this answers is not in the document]
Mark-Fenng: If so, ${image} will actually call image.toString() and the warning message will always be `Cannot inline image: ${image}! Error:[object HTMLImageElement]! Error: ...`. Is there a way to provide more useful information?

packages/rrweb-snapshot/src/snapshot.ts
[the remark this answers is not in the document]
croqaz: I don't want to delay this PR too much, so I'll revert, because it's not that important.
But for reference https://caniuse.com/?search=CSSStyleSheet.rules says "This feature is deprecated/obsolete and *should not be used*" and the global usage is `95.06%`.
And for https://caniuse.com/?search=CSSStyleSheet.cssRules there's no deprecation message and the global usage is `95.6%` so only 0.5% higher.
That's why I don't think it's so important.

packages/rrweb-snapshot/src/snapshot.ts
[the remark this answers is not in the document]
croqaz: That's a very good point! I added img src in the console, it will be safe because the src is over-written inside the try that fails, so it should say the correct URL of the image, which is useful for debugging.

packages/rrweb-snapshot/src/snapshot.ts
croqaz: You're totally right! I fixed it in the latest commit. Thanks!!

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -512,22 +515,22 @@ function serializeNode(
...
+          } catch (err) {
+            console.warn(`Cannot inline img src=${image.currentSrc}! Error: ${err}`);
+          }
+          oldValue
+            ? (attributes.crossOrigin = oldValue)
+            : delete attributes.crossOrigin;
           };
           // The image content may not have finished loading yet.
           if (image.complete && image.naturalWidth !== 0) recordInlineImage();
           else image.onload = recordInlineImage;
```
Mark-Fenng, line 533: please format the code of this file with prettier

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -586,8 +589,8 @@ function serializeNode(
               (n.parentNode as HTMLStyleElement).sheet!,
             );
           }
-        } catch {
-          // ignore error
+        } catch (err) {
+          console.warn(`Cannot get CSS styles for: ${n}! Error: ${err}`);
```
Mark-Fenng, line 593: Same issue as the former one. The output will always be `Cannot get CSS styles for: [object Text]! Error:...`
And the try block is actually getting CSS styles of node's parentNode rather than itself. So this warning message isn't accurate.

Mark-Fenng:
@croqaz Thanks for your contribution. You may still have two little points to correct and finally, don't forget to rebase the latest master branch coz there are conflicts between them.

packages/rrweb-snapshot/src/snapshot.ts
croqaz: Good suggestion.
It's hard to differentiate one <style> tag from another, because they generally don't have ID or Class or anything, so I'm thinking to add some of the CSS content, it should be safe to display for debug.

Mark-Fenng: Maybe just directly log n and its parentNode as object?

croqaz: Sorry, my Javascript is not that good, so I don't understand how to represent that object in a useful way in a log message...
You mean JSON represent the object? DOM elements have hundreds or properties from the whole inheritance objects, which one is meaningful for this warning message?

croqaz: There's also the option to remove the log.warning altogether, if you think it's not that important and to merge this PR faster.

Mark-Fenng: I mean you may print it like this:
console.warn(`Can't get CSS styles from text's parentNode. Error: ${err}.`, n);

croqaz: Ugh you're right! Fixing now!

Mark-Fenng: Don't forget to rebase your PR on the latest master branch

croqaz: I just synced our repo, I suppose it's the same thing?
I don't see any code conflicts.

Mark-Fenng (approved):

croqaz:
Thank you for all the guidance @Mark-Fenng !

Issue #746, opened 2021-11-03
<select> value is not masked when maskInputOptions is turned on


userback:
**RRWEB Settings:**
```
rrwebRecord({
    maskInputOptions: {
        select: true
    }
})
```

**HTML:**
```
<select>
    <option selected>Item 1</option>
    <option>Item 2</option>
</select>
```

**Bug:**
When the `<select>` is loaded with a default selected value, the default value "Item 1" is still displayed without any masking. When I change the option to "Item 2", "Item 2" shows as blank in the replay. So it looks like there is some masking going on when the select value is changed, but the default value does not have any masking. 

**Expected behaviour:**
The select element should show "********" instead of "Item 1".

PR #810, opened 2022-01-23
feat: add observer for 'selected' setter of HTMLOptionElement and try to fix issue #746

Mark-Fenng:
1. add observer for 'selected' setter of HTMLOptionElement
   If we change the 'selected' value of an option element, the value of its parent select element would be changed as well. So I add an observer for the 'selected' setter. 
2. try to fix issue #746

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -491,7 +491,7 @@ function serializeNode(
         }
       }
       if (tagName === 'option') {
-        if ((n as HTMLOptionElement).selected) {
+        if ((n as HTMLOptionElement).selected && !maskInputOptions['select']) {
           attributes.selected = true;
         } else {
           // ignore the html attribute (which corresponds to DOM (n as HTMLOptionElement).defaultSelected)
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -371,8 +371,10 @@ function initInputObserver(
   userTriggeredOnInput: boolean,
 ): listenerHandler {
   function eventHandler(event: Event) {
-    const target = getEventTarget(event);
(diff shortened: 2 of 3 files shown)


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -80,10 +80,10 @@ let canvasCtx: CanvasRenderingContext2D | null;
 
 function initCanvasService(doc: Document) {
   if (!canvasService) {
-      canvasService = doc.createElement('canvas');
+    canvasService = doc.createElement('canvas');
```
Yuyz0112, line 83: could you rebase master to avoid the format diff?

packages/rrweb/src/record/observer.ts
```diff
@@ -369,8 +369,10 @@ function initInputObserver(
   userTriggeredOnInput: boolean,
 ): listenerHandler {
   function eventHandler(event: Event) {
-    const target = getEventTarget(event);
+    let target = getEventTarget(event);
     const userTriggered = event.isTrusted;
+    if (target && (target as Element).tagName === 'OPTION')
+      target = (target as Element).parentElement;
```
Yuyz0112, line 375: please inline the mechanism as comment at here

BTW, is there any doc/spec pointing out the parent element of `option` should be `select`?

Mark-Fenng: No problem, I will add it.

_is there any doc/spec pointing out the parent element of option should be select_
This is a good point, parent element could also be \<optgroup> and \<datalist> according to 
image captured from [OptionElement MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option).
But these two elements aren't hooked for now, so this exception could be excluded by this condition`INPUT_TAGS.indexOf((target as Element).tagName) < 0`

Yuyz0112 (approved):
LGTM

PR #812, opened 2022-01-24
improve robustness of inlineImages feature

Mark-Fenng:
Add a workaround for situations the content of images hasn't finished loading. On my laptop (mac M1 chip and chrome/firefox v96), `image.crossOrigin = 'anonymous';` will leave the image in an unloaded state so that the recorded rr_dataURL will be empty.
This PR may fix issue #799

### DIFF
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -227,10 +227,13 @@ function buildNode(
               }
             };
           } else if (tagName === 'img' && name === 'rr_dataURL') {
-            const image = (node as HTMLImageElement);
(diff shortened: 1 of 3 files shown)


Mark-Fenng:
This situation is similar to https://github.com/rrweb-io/rrweb/issues/764#issuecomment-993091039 in some ways. I think it's ok to fetch the image data asynchronously because normally the recording process would last for a relatively long time compared to the loading process. The closure of onload callback contains the reference of recorded data so it will finally make the recorded data complete.

Yuyz0112 (approved):
good catch 👍

liuwanj:
When will this `inlineImages` feature be released

PR #815, opened 2022-01-27
Add skipActivity on setConfig

troglotit:
Hi!

This PR adds ability to fast-forward while replayer is already playing. Not sure that 100% is correct though

### DIFF
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -289,6 +289,38 @@ export class Replayer {
     if (!this.config.skipInactive) {
       this.backToNormal();
     }
+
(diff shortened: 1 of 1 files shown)


Mark-Fenng:
Hi troglotit, I'm so sorry that I don't quite understand your newly added feature. Could you please provide a more detailed description of the application scenario?

troglotit:
I wanted to have an option to set `skipInactive: true` while the player is already playing and for it to start fast-forward, rather than to have to wait for next `case EventType.IncrementalSnapshot` event

Juice10:
@troglotit could you add a test case which would fail (before you patch) and pass after this PR is applied?

PR #822, opened 2022-02-02
Fix inline image issues

croqaz:
This PR fixes 2 issues:

- ~fix `img.complete` being false, which breaks inline image feature completely~ EDIT: this solution was not good, see the comments
It seems that immediately after you set `image.crossOrigin = 'anonymous'` the image becomes "not loaded yet", which breaks the logic. It's easy to check just by adding a quick `console.log('IMAGE:', img, img.complete, img.naturalWidth, img.naturalHeight)` just before this if condition: https://github.com/rrweb-io/rrweb/blob/b0dc3885d9bfbe4fb54f7f2a1256cbe0c1c359f2/packages/rrweb-snapshot/src/snapshot.ts#L534-L535
I moved that logic just before we call `toDataURL()` just like this article says: https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image

- remove `srcset` which overwrites the `src` and breaks restoring images that have `srcset`

ping @Mark-Fenng because you have the context of this feature from previous PR.

### DIFF
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -228,15 +228,23 @@ function buildNode(
             };
           } else if (tagName === 'img' && name === 'rr_dataURL') {
             const image = node as HTMLImageElement;
+            image.removeAttribute(name);
             if (!image.currentSrc.startsWith('data:')) {
               // Backup original img src. It may not have been set yet.
               image.setAttribute(
                 'rrweb-original-src',
-                n.attributes['src'] as string,
+                n.attributes.src as string,
               );
               image.src = value;
             }
-            image.removeAttribute(name);
+            if (n.attributes.srcset) {
+              // Backup original img srcset
+              image.setAttribute(
+                'rrweb-original-srcset',
+                n.attributes.srcset as string,
+              );
+              image.removeAttribute('srcset');
+            }
           }
 
           if (name === 'rr_width') {
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -214,7 +214,8 @@ function buildNode(
               n.attributes.srcset &&
               n.attributes.rr_dataURL
             ) {
-              // ignore img srcset here, it will be used below
+              // backup original img srcset
+              node.setAttribute('rrweb-original-srcset', n.attributes.srcset as string);
             } else {
               node.setAttribute(name, value);
             }
@@ -241,13 +242,6 @@ function buildNode(
                 n.attributes.src as string,
               );
               image.src = value;
-            }
-            if (n.attributes.srcset) {
-              // Backup original img srcset
-              image.setAttribute(
-                'rrweb-original-srcset',
-                n.attributes.srcset as string,
-              );
             }
           }
 
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -209,6 +209,12 @@ function buildNode(
               n.attributes.href.endsWith('.js')
             ) {
               // ignore
+            } else if (
+              tagName === 'img' &&
+              n.attributes.srcset &&
+              n.attributes.rr_dataURL
+            ) {
+              // ignore img srcset here, it will be used below
             } else {
               node.setAttribute(name, value);
             }
@@ -228,7 +234,6 @@ function buildNode(
             };
           } else if (tagName === 'img' && name === 'rr_dataURL') {
             const image = node as HTMLImageElement;
-            image.removeAttribute(name);
             if (!image.currentSrc.startsWith('data:')) {
               // Backup original img src. It may not have been set yet.
               image.setAttribute(
@@ -243,7 +248,6 @@ function buildNode(
                 'rrweb-original-srcset',
                 n.attributes.srcset as string,
               );
-              image.removeAttribute('srcset');
             }
           }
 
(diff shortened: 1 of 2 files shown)


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -513,13 +513,15 @@ function serializeNode(
...
-        image.crossOrigin = 'anonymous';
         const recordInlineImage = () => {
+          const oldValue = image.crossOrigin;
           try {
             canvasService!.width = image.naturalWidth;
             canvasService!.height = image.naturalHeight;
             canvasCtx!.drawImage(image, 0, 0);
+            // see this article about crossorigin = "anonymous"
+            // https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image
+            image.crossOrigin = 'anonymous';
```
Mark-Fenng, line 524: Like what I said in this comment: https://github.com/rrweb-io/rrweb/pull/812#issue-1112834149
> image.crossOrigin = 'anonymous'; will leave the image in an unloaded state so that the recorded rr_dataURL will be empty.

croqaz: Nice!! We both discovered the same thing!! I actually didn't see that PR.

However the current implementation doesn't work for me, I don't get the dataURL even after waiting for 5 seconds or more. Moving the `crossOrigin = 'anonymous'` just before calling `toDataURL()` fixes the problem.

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -228,15 +228,23 @@ function buildNode(
             };
           } else if (tagName === 'img' && name === 'rr_dataURL') {
             const image = node as HTMLImageElement;
+            image.removeAttribute(name);
```
Mark-Fenng, line 231: BTW, I don't really understand the purpose of this line in the original PR #770

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -228,15 +228,23 @@ function buildNode(
...
               image.src = value;
             }
-            image.removeAttribute(name);
+            if (n.attributes.srcset) {
+              // Backup original img srcset
+              image.setAttribute(
+                'rrweb-original-srcset',
+                n.attributes.srcset as string,
+              );
+              image.removeAttribute('srcset');
```
Mark-Fenng, line 246: How to make sure 'srcset' has been set before 'rr_dataURL'?

packages/rrweb-snapshot/src/snapshot.ts
Mark-Fenng: Can you now get full base64 data when the image is unloaded?

Mark-Fenng: Have you tried to replace `image.onload = recordInlineImage` with `image.addEventListener('load', recordInlineImage, false)`?

croqaz: OK so I discovered that when I set the crossOrigin property on an image, the browser makes a separate request, that you can actually see in the Network tab in Developer tools. I didn't know that.

This does work for me right now, because I already have some cached images and I'm intercepting the request, which returns the cached value instantly. But it won't work in a slow network.

I will have to find a better, general solution. Maybe even hack the images to have `crossOrigin = 'anonymous'` BEFORE calling rrWeb snapshot.

packages/rrweb-snapshot/src/rebuild.ts
croqaz: Good point! I thought 'rr_dataURL' is attached on the DOM node and I wanted to make sure it's not duplicated, because it's a lot of text in there.

packages/rrweb-snapshot/src/snapshot.ts
Mark-Fenng: > Maybe even hack the images to have crossOrigin = 'anonymous' BEFORE calling rrWeb snapshot.

This sounds like a great solution for me

packages/rrweb-snapshot/src/rebuild.ts
Mark-Fenng: I think when it calls `image.removeAttribute(name);`, the `rr_dataURL` hasn't been set before

croqaz: Removed in the latest commit.

packages/rrweb-snapshot/src/rebuild.ts
croqaz: Handled in the latest commit.

packages/rrweb-snapshot/src/snapshot.ts
croqaz: I have reverted this change, because the original solution was better.

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -232,11 +238,17 @@ function buildNode(
...
               );
               image.src = value;
             }
-            image.removeAttribute(name);
+            if (n.attributes.srcset) {
+              // Backup original img srcset
+              image.setAttribute(
+                'rrweb-original-srcset',
+                n.attributes.srcset as string,
+              );
```
Mark-Fenng, line 250: I suggest moving this function to [line #278](https://github.com/rrweb-io/rrweb/pull/822/files#diff-0d9a03e75eacff8f6be89336b49419b210455c1a81d8f46fae7cea2d83a8f93eR217)

croqaz: Done!

Mark-Fenng (approved):

PR #823, opened 2022-02-02
fix: custom style rules don't get inserted into some iframe elements

Mark-Fenng:
### bug
If an iframe element is added to the dom first and its document nodes are added later, the custom style rules won't get inserted.

### reason
The custom style rules were designed to add through [code](https://github.com/rrweb-io/rrweb/blob/master/packages/rrweb/src/replay/index.ts#L1485-L1488) . But when we apply the iframe's document content mutation later, the inserted style element would be wiped [here](https://github.com/rrweb-io/rrweb/blob/master/packages/rrweb-snapshot/src/rebuild.ts#L332).

### idea of the fix
We can insert our custom rules whenever a new HTML element is appended to an iframe's content document. Because `attachDocumentToIframe`'s afterAppend callback seems to be the common execution path of all situations.

### DIFF
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -660,10 +660,6 @@ export class Replayer {
       this.newDocumentQueue = this.newDocumentQueue.filter(
         (m) => m !== mutationInQueue,
       );
-      if (builtNode.contentDocument) {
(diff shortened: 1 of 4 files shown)


Yuyz0112 (approved):
LGTM

PR #826, opened 2022-02-04
refactoring observer options

Yuyz0112:
This PR does not have any functional change to the library, just a code refactoring to the rrweb's observer module.

Previously, we are repeating the type of parameters many times across the code base, after the refactoring, most parameters are using the same interface.

Also, we are changing the function parameters from multiple arguments into a single object argument and using destructing to access the properties, which reduces the mistakes of passing arguments with the same type in the wrong order.


Juice10 (approved):
This is great! So much cleaner, less repetitive and much easier to add something without having to change it in 4 or 5 places. Well done!

PR #827, opened 2022-02-04
Disable virtual parent optimization on <head>

eoghanmurray:
Was experiencing an issue where the virtual parent optimization was destroying already applied styles during <head> mutations


eoghanmurray:
Re. https://github.com/rrweb-io/rrweb/pull/827/commits/b27789eb6a59c4f549b93d00b52d99b398e139b7 I managed to patch over the failing test by also cancelling the virtual parent optimization for subsequent adds if any or under <head>...  however I didn't figure out why exactly it wasn't working, and the solution in https://github.com/rrweb-io/rrweb/pull/827/commits/b27789eb6a59c4f549b93d00b52d99b398e139b7 is likely brittle.

Hopefully this is okay as the virtual parent optimization is slated for removal.

eoghanmurray:
OK so I couldn't create a test to recreate this!
I also couldn't recreate it by reverting our codebase.
The changeset here definitely fixed it for us at the time, but I'm happy to let this fade away given that this optimization is going to be replaced.

PR #828, opened 2022-02-07
fix: an error when I stop the recording process

Mark-Fenng:
### bug
When the option 'recordCanvas' was false and I stopped the recording process, I got this error:
image

### reason
When 'recordCanvas' is false, the CanvasManager is created but `resetObservers` isn't initiated. When I stop recording, `resetObservers` is still executed, which is actually undefined.


Juice10:
Sorry about that @Mark-Fenng, thanks for finding and fixing this!

Juice10 (approved):

Issue #754, opened 2021-11-08
Shadowroot support not working properly


maximilianMairinger:
A recent [project of mine](https://github.com/maximilianMairinger/xcenic/tree/record) heavily utilizes nested custom elements with shadowroots. I tried implementing rrweb but failed. Here is a running example (implementing the [getting started guide](https://github.com/rrweb-io/rrweb/blob/master/guide.md#getting-started); view source [recording start](https://github.com/maximilianMairinger/xcenic/blob/record/app/record.ts) and [recording play](https://github.com/maximilianMairinger/xcenic/blob/record/app/_component/_themeAble/_frame/_page/adminPage/adminPage.ts)): http://035401c.xcenic.maximilian.mairinger.com/admin/. After clicking "Stop recording" view the recording object in the console. Everything outside the first shadowroot is properly recorded, but nothing inside (very similar to a lot of other solutions out there).

I noticed however your effort to support shadow roots and tried setting a decently low checkoutEveryNms time. Which **worked**, although at the cost of absurdly high cpu utilization (not surprising), proving that the pure serialization process works.

Yuyz0112:
looks like the new PR from @Mark-Fenng fixed this.

Issue #38, opened 2019-01-10
Does not record shadow-dom/custom-elements


orenmizr:
scenario: by using repl record a demo page like https://codepen.io/chriscoyier/full/rwvjVN
expected: the page should show the elements
actual: the page shows up empty

Yuyz0112:
Thanks for the report! Will dig into it ASAP.

Yuyz0112:
After taking a look at shadow-dom, it looks like we need to check the `shadowRoot` property when traversing and rebuild shadow dom.

Yuyz0112:
@orenmizr 

After play with shadow dom a little more, I think this is a similar situation with iframe.
Because the mutation observer and event listeners also need to be attached to the shadowRoot, which is a case of `nested recording`.

So this may take more time to discuss whether to implement the feature.
Currently, if you want to track something happened in a shadow-dom/iframe, the best way is to init a recorder inside it.

kerbeh:
@Yuyz0112 

> Currently, if you want to track something happened in a shadow-dom/iframe, the best way is to init a recorder inside it.

Do you have an example of init-ing a recorder inside shadow-dom?

TobiaszCudnik:
[ShadowMutationObserver](https://github.com/medialize/ally.js/blob/b8fd3cdbb9464df08708f11cba35b7481ad846d4/src/observe/shadow-mutations.js#L31) from ally.js may be relevant here.

Im considering taking a dab into this. Would you expect any problems with node IDs in this case? I suppose listeners disposal will have to be added, as handlers won't be delegated any more. Thx.

vipulwairagade:
@TobiaszCudnik Do you have workaround how to record iframe on my webpage ?

flowertree:
我也在期待是否能支持iframe 的录制

wupinlang:
我也在期待是否能支持iframe 的录制

TobiaszCudnik:
> 我也在期待是否能支持iframe 的录制

Is this spanish?

wupinlang:
> > 我也在期待是否能支持iframe 的录制
> 
> Is this spanish?

it is chinese

dcaravana:
> Currently, if you want to track something happened in a shadow-dom/iframe, the best way is to init a recorder inside it.

Yes @Yuyz0112 that's smart, but in my case the problem are page changes that will generate a new recording for each page vs a single recording produced by a recorder instantiated on the page hosting the iframe.

[...]

PR #834, opened 2022-02-16
add support for nested shadow dom

Mark-Fenng:
### scenario
If a page highly depends on custom elements, there may be lots of nested shadow dom. #754 may as an example
And the current version doesn't support recording shadow doms which were attached after the full snapshot.

### what changed

1. Hijack HTMLElement.attachShadow. When we attach a shadow dom on a created element, the old version can't detect the change and the new shadow dom can't be observed.
2. If we attach a shadow dom on a created element, the serialized element data won't contain 'isShadowHost' property. So I have to add a workaround on the replay side.
3. `notInDoc` check doesn't work for nested shadow dom so I changed the checking way.
4. add recording test for nested shadow dom 

Related issues: #38, #754

### DIFF
--- a/packages/rrweb/src/record/shadow-dom-manager.ts
+++ b/packages/rrweb/src/record/shadow-dom-manager.ts
@@ -19,6 +19,7 @@ export class ShadowDomManager {
   private scrollCb: scrollCallback;
   private bypassOptions: BypassOptions;
   private mirror: Mirror;
+  private originalAttachShadow: (init: ShadowRootInit) => ShadowRoot;
 
   constructor(options: {
     mutationCb: mutationCallBack;
@@ -30,6 +31,16 @@ export class ShadowDomManager {
     this.scrollCb = options.scrollCb;
     this.bypassOptions = options.bypassOptions;
     this.mirror = options.mirror;
+
+    // Monkey patch 'attachShadow' to observe newly added shadow doms.
+    this.originalAttachShadow = HTMLElement.prototype.attachShadow;
+    const attachShadow = this.originalAttachShadow;
+    const manager = this;
+    HTMLElement.prototype.attachShadow = function () {
+      const shadowRoot = attachShadow.apply(this, arguments);
+      manager.addShadowRoot(this.shadowRoot, document);
+      return shadowRoot;
+    };
   }
 
   public addShadowRoot(shadowRoot: ShadowRoot, doc: Document) {
@@ -51,5 +62,9 @@ export class ShadowDomManager {
       doc: (shadowRoot as unknown) as Document,
       mirror: this.mirror,
     });
+  }
+
+  public reset() {
+    HTMLElement.prototype.attachShadow = this.originalAttachShadow;
   }
 }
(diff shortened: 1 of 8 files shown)


packages/rrweb/src/record/shadow-dom-manager.ts
```diff
@@ -30,6 +31,16 @@ export class ShadowDomManager {
     this.scrollCb = options.scrollCb;
     this.bypassOptions = options.bypassOptions;
     this.mirror = options.mirror;
+
+    // Monkey patch 'attachShadow' to observe newly added shadow doms.
+    this.originalAttachShadow = HTMLElement.prototype.attachShadow;
+    const attachShadow = this.originalAttachShadow;
+    const manager = this;
+    HTMLElement.prototype.attachShadow = function () {
```
Mark-Fenng, line 39: I guess this line may potentially have the same issue as #843 if the shadow dom is inside an iframe. I will check it later

Mark-Fenng: A new patch has been pushed to solve this problem.

packages/rrweb/src/record/shadow-dom-manager.ts
```diff
@@ -30,6 +39,17 @@ export class ShadowDomManager {
...
+      const shadowRoot = attachShadow.apply(this, arguments);
+      if (this.shadowRoot)
+        manager.addShadowRoot(this.shadowRoot, this.ownerDocument);
+      return shadowRoot;
+    };
```
Juice10, line 52: You could use `patch` for this: https://github.com/rrweb-io/rrweb/blob/f74b7bb28d1f63d0a8398fdd7f80a5f86b1df01c/packages/rrweb/src/utils.ts#L169-L208

packages/rrweb/src/record/shadow-dom-manager.ts
```diff
@@ -52,4 +72,37 @@ export class ShadowDomManager {
...
+  /**
+   * Monkey patch 'attachShadow' of an IFrameElement to observe newly added shadow doms.
+   */
+  public observeAttachShadow(iframeElement: HTMLIFrameElement) {
+    if (iframeElement.contentWindow) {
+      const originalAttachShadow = (iframeElement.contentWindow as WindowWithHTMLElement)
+        .HTMLElement.prototype.attachShadow;
+      const manager = this;
+      (iframeElement.contentWindow as WindowWithHTMLElement).HTMLElement.prototype.attachShadow = function () {
+        const shadowRoot = originalAttachShadow.apply(this, arguments);
```
Juice10, line 85: You could use `patch` for this as well: https://github.com/rrweb-io/rrweb/blob/f74b7bb28d1f63d0a8398fdd7f80a5f86b1df01c/packages/rrweb/src/utils.ts#L169-L208

Juice10 (approved):
I'm new to shadowDom, so probably not the best to judge this, but the PR looks good to me!

Yuyz0112:
Looks this is the only way to observe the attaching behavior.

PR #836, opened 2022-02-16
Decrease embedded img size for inlineImages

croqaz:
Hi.

This is a small PR for decreasing the final snapshot size, for the majority of pages.

For example a page with lots of images like: https://www.1samsautos.com/2005-GMC-Safari-Passenger-Denver-CO-Fort-Collins-Colorado-Springs-Kansas-City-DENVER/used_car/Wrqim%5bzGkRg%3d ; currently the snapshot size jumps between 96MB and 120MB, depending on how much I'm waiting for the page to load, or if I'm scrolling, because the images are loaded lazily on scroll event.

If we merge this PR, the snapshot size will be only 12MB, so 10x smaller.

However there may be some issues with this hard-coded convert-all-to-JPEG approach:
- what if the original image is a GIF, or PNG with transparency? the snapshot will make the transparency white, which can make some pages look broken or ugly
- what if the `canvas.toDataURL()` function would return a smaller string if we used PNG? there are large, mostly monochrome images that can be smaller as PNG.

There doesn't seem to be a built-in way to check the HTMLImageElement Content-type, or MIME-type. If the image was GIF or PNG we could just convert it to PNG, otherwise would be JPEG. We could use the img src extension, but that's not reliable.

Ping @Mark-Fenng again, because you know this feature very well already. I hope you don't mind :smile:

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -7,6 +7,7 @@ import {
   idNodeMap,
   MaskInputOptions,
   SlimDOMOptions,
+  DataURLOptions,
   MaskTextFn,
   MaskInputFn,
   KeepIframeSrcFn,
@@ -378,6 +379,7 @@ function serializeNode(
     maskInputOptions: MaskInputOptions;
     maskTextFn: MaskTextFn | undefined;
     maskInputFn: MaskInputFn | undefined;
+    dataURLOptions?: DataURLOptions,
     inlineImages: boolean;
     recordCanvas: boolean;
     keepIframeSrcFn: KeepIframeSrcFn;
@@ -393,6 +395,7 @@ function serializeNode(
     maskInputOptions = {},
     maskTextFn,
     maskInputFn,
+    dataURLOptions = {},
     inlineImages,
     recordCanvas,
     keepIframeSrcFn,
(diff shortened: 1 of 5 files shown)


Mark-Fenng:
Thank you for pinging me! I think the work is great. About the compression switch, compression ratio, and final format, personally, I suggest making them configurable. How do you think about it @Yuyz0112

Yuyz0112:
Same here, extending inline images config to an object may help.

croqaz:
Guys, how about creating an optional param object `dataURLOptions` or something like that.
The default would be `{ type: null, quality: null }`.

Then if you provide something like `{ type: "image/png", quality: 0.7 }`, it would be applied to record canvas and inline images?
Because both tags use `canvas.toDataURL(...)`.
My reasoning is that you can save for both tags and if the browser supports webp, it makes sense to use it everywhere.

What do you guys thing?

Mark-Fenng:
Good point! I think it could work. If changing canvas recording, I think @Juice10 may take a look as well.

Juice10:
> Good point! I think it could work. If changing canvas recording, I think @Juice10 may take a look as well.
Thanks for looping me in @Mark-Fenng 

 I can't think of any reason to use separate image types for canvas, versus img.
> 
> What do you guys thing?

Unifying the config for canvas and images makes sense to me

Juice10:
If you'd like to decrease the embedded img size for inline images, why not try to use the existing image?
I know it's hard to try and figure out what format an image is in once its loaded into the browser in an image element, but we could request the image a second time by using [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). This will most likely come straight from the browser cache and will be really fast to load. With the added bonus that this might already be the very smallest file possible if the website owner is serious about performance.

Using fetch comes with the benefit that we won't have to use `canvas.toDataURL`.
I've been doing some experimenting with recording canvas snapshots periodically and I've been bumping into a lot of issues regarding performance of `toDataURL`. Depending on what format you choose its either slow or very slow and larger images tend to make browsers drop frames. Alternatively [`toBlob`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) is async and about 2-3x faster as it doesn't have to do any base64 encoding but that just delays the inevitable base64 encoding (maybe we can do that part in a web worker?).

Big downside to both `canvas.toBlob` and `fetch(img.src)` is that they are async. So it means a different approach when saving these too events. This might be a good starting point for a discussion around async events @Mark-Fenng @Yuyz0112 @eoghanmurray (in a different issue probably)

croqaz:
I have actually tried the image fetch option, when I started working on the offline image feature. It's a bit of work to make all the functions in the snapshot.ts file async, but it makes it easier to wait for some callbacks and events.
It's a breaking feature tho, and I shouldn't probably implement it in this PR.
Because the code is sync, currently I don't see any other way of of decreasing the size of the images. I already implemented the code and I'm already using it on my end.

Juice10:
> I have actually tried the image fetch option, when I started working on the offline image feature. It's a bit of work to make all the functions in the snapshot.ts file async, but it makes it easier to wait for some callbacks and events.
> 
> It's a breaking feature tho, and I shouldn't probably implement it in this PR.
> 
> Because the code is sync, currently I don't see any other way of of decreasing the size of the images. I already implemented the code and I'm already using it on my end.

I totally agree that this PR is the next logical step to what we already have, and makes sense to do any image fetching or async work in a separate PR. Especially if we want to make the snapshot async.

What we could potentially do to keep everything sync, is to create a new type of event that contains the url of the image, and the contents. And then in the replayer we can loop through all of these new types of events to know what urls have been saved, and when the replayer creates an image it can check to see if we have an offline copy of that url and replace the contents of it with the saved version. 
We are doing something similar for the canvas replay. If you search for `imageMap` throughout the source code you'll see it in action. https://github.com/rrweb-io/rrweb/blob/f74b7bb28d1f63d0a8398fdd7f80a5f86b1df01c/packages/rrweb/src/replay/index.ts#L125
Credits go to @Yuyz0112 for coming up with this concept

Mark-Fenng (approved):

Issue #668, opened 2021-08-16
Page get stuck while manipulate milion's of nodes


PeterChen1997:
Our page has a lot of nodes, if we open rrweb while spa page jumping, the page will get stuck for 1-3s than not open it. Here is the performance while spa page jumping

open:
![image](https://user-images.githubusercontent.com/21072589/129531836-265d9154-2d53-4c3f-8db3-70ad70d44646.png)

close:
![image](https://user-images.githubusercontent.com/21072589/129531818-eac77d33-9514-439f-b876-01868618b2c8.png)

Is there any config that i can speed up this action?

Yuyz0112:
Is it possible to provide the record page URL or similar example page for performance testing?

PeterChen1997:
> Is it possible to provide the record page URL or similar example page for performance testing?

the env is so complicated.. i think it's hard to replay this 1:1

I think it's the node manipulation cause this problem, it's fine when node is smaller than 400W ...

Yuyz0112:
@PeterChen1997 Maybe you can dump the dev tool's profile.json so we can check which code path takes most of the time.

PeterChen1997:
> @PeterChen1997 Maybe you can dump the dev tool's profile.json so we can check which code path takes most of the time.

can i send the json in wechat? Here is my number: 13093023233

Yuyz0112:
@PeterChen1997 send a request to you. You can delete the number now:)

Yuyz0112:
After some debug, most of the time was used to iterating DOM nodes and node's classList.

PeterChen1997:
> After some debug, most of the time was used to iterating DOM nodes and node's classList.

Got it, is it the bottleneck of our algorithm? I found when my computer cpu is free, every will be all right...

PR #840, opened 2022-02-21
Use the built in .closest method to checked for blocked class

eoghanmurray:
Transfer recursive parent class checking into the browser engine for an expected performance improvement.
I've seen `isBlocked` come up in profilers but don't have a badly performing example to hand for direct comparison when I came up with this today.
A very minor benefit is that this commit will prevent the recursive isBlocked method from showing up lots of times in profiling!

https://developer.mozilla.org/en-US/docs/Web/API/Element/closest available on all modern browsers

### DIFF
--- a/packages/rrweb/src/utils.ts
+++ b/packages/rrweb/src/utils.ts
@@ -230,7 +230,11 @@ export function isBlocked(node: Node | null, blockClass: blockClass): boolean {
   if (node.nodeType === node.ELEMENT_NODE) {
     let needBlock = false;
     if (typeof blockClass === 'string') {
-      needBlock = (node as HTMLElement).classList.contains(blockClass);
+      if ((node as HTMLElement).closest !== undefined) {
+        return (node as HTMLElement).closest('.' + blockClass) !== null;
+      } else {
+        needBlock = (node as HTMLElement).classList.contains(blockClass);
+      }
     } else {
       (node as HTMLElement).classList.forEach((className) => {
         if (blockClass.test(className)) {


Juice10 (approved):
LGTM!

Yuyz0112:
Interesting!

I've done a simple benchmark and looks like this change has a ~5x improvement to the performance.

```html
<html>
  <body>
    <div id="root" class="class-to-check"></div>
    <script>
      const roolEl = document.querySelector("#root");
      let el = roolEl;
      let count = 2000;
      while (count) {
        count--;
        const childEl = document.createElement("div");
        childEl.classList.add("other-class");
        el.appendChild(childEl);
        el = childEl;
      }

      function suite(fn, round) {
        while (round) {
          round--;
          fn();
        }
      }

      // user land impl
      function exist(node, className) {
        const e = node.classList.contains(className);
        return e || exist(node.parentNode, className);
      }
      // benchmark
      const round = 1000;
      console.time("native");
      suite(() => Boolean(el.closest(`.class-to-check`)), round);
      console.timeEnd("native");
      console.time("user land");
      suite(() => exist(el, "class-to-check"), round);
      console.timeEnd("user land");
    </script>
  </body>
</html>

```

Yuyz0112:
May helps in #668 #789

eoghanmurray:
Thanks for doing the hard work of performance testing @Yuyz0112 !

One further change (for another PR) that suggested itself is to check the page for presence of the blocked class once upon fullsnapshot, and convert `isBlocked` to a no-op if the blocked class isn't present in the first place.  We'd have to update this upon mutation of course.

Issue #841, opened 2022-02-23
SVG in iframe can be recorded but cannot be displayed during playback


Jiacheng787:
录制一个同源 iframe 中的页面，发现 svg 渲染的图表已被录制下来，但无法展示，另外还有 svg 的图标也被录制，但同样无法展示

![image](https://user-images.githubusercontent.com/43509711/155269078-afb18a21-dd48-4ced-8d1b-9f43afa982a1.png)

Mark-Fenng:
could you provide the code of that SVG icon?

Jiacheng787:
Actually, any svg code is reproducible, but just make sure to include it in a iframe page.

```html
<svg width="42" height="24" fill="none" xmlns="http://www.w3.org/2000/svg" class="dash-debug-menu__icon dash-debug-menu__icon--debug"><path d="M14.85 21.197 5.616 12l9.225-9.197L12.028 0 0 12l12.037 12 2.813-2.803Zm12.3 0L36.375 12 27.15 2.803 29.963 0 42 12 29.962 24l-2.812-2.803Z" fill="#fff"></path></svg>
```

Mark-Fenng:
Thanks for the report!  I'll look into it.

Jiacheng787:
Ok, please let me know if there is any progress. Waiting for your fix method.

PR #843, opened 2022-02-23
fix:  #841 can't record SVG element inside iframe properly

Mark-Fenng:
This PR is trying to fix issue #841.

SVGElement inside iframe is different from SVGElement outside. So even though an el inside iframe is actually an SVG element`, el instanceof SVGElement` is still false and the `isSVG` property is missing in the recorded data. 

`ownerSVGElement` even has better compatibility than `mutation observer`. 
https://caniuse.com/?search=ownerSVGElement
https://caniuse.com/?search=mutationObserver

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -219,7 +219,7 @@ export function absoluteToDoc(doc: Document, attributeValue: string): string {
 }
 
 function isSVGElement(el: Element): boolean {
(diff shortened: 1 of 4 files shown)


Yuyz0112:
LGTM, does the diff in the snapshot expected?

Mark-Fenng:
> LGTM, does the diff in the snapshot expected?

yes, I check them carefully

Mark-Fenng:
Oh sorry, I forgot to say that the snapshot is changed because I add the integration test for this situation.

PR #844, opened 2022-02-23
fix: Uncaught TypeError: Illegal invocation when recording incremental canvas mutation

Mark-Fenng:
reproducible code: https://codepen.io/mark-fenng/pen/JjOZBdN?editors=1111
image
Steps to reproduce the error:
First, click ‘start to record’, then click ‘create a canvas’ and the error will just show up.

Thanks for @Juice10's fix!


Juice10 (approved):
@Mark-Fenng Thanks for creating such a great reproduction environment!

PR #846, opened 2022-02-26
Avoid triggering a CSP (content security policy) error

eoghanmurray:
Fix for #816 with `.setAttribute('style')` — see CSP style-src: unsafe-inline

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -462,7 +462,7 @@ export default class MutationBuffer {
     if (isIgnored(m.target, this.mirror)) {
       return;
     }
-    const unattachedDoc = new Document(); // avoid upsetting original document from a Content Security point of view
+    const unattachedDoc = document.implementation.createHTMLDocument(); // avoid upsetting original document from a Content Security point of view
     switch (m.type) {
       case 'characterData': {
         const value = m.target.textContent;
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -462,7 +462,14 @@ export default class MutationBuffer {
     if (isIgnored(m.target, this.mirror)) {
       return;
     }
-    const unattachedDoc = document.implementation.createHTMLDocument(); // avoid upsetting original document from a Content Security point of view
+    let unattachedDoc;
+    try {
+      // avoid upsetting original document from a Content Security point of view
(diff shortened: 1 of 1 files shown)


packages/rrweb/src/record/mutation.ts
```diff
@@ -426,6 +426,7 @@ export default class MutationBuffer {
     if (isIgnored(m.target)) {
       return;
     }
+    const unattached_doc = new Document();  // needed to avoid upsetting original document from a Content Security point of view
```
razor-x, line 429: I'm guessing to match the code style this should be `unattachedDoc`

eoghanmurray: Have fixed this in a subsequent force-push

Juice10 (approved):

YunFeng0817 (approved):

YunFeng0817:
Seems like this pull request can cause the failure of two test cases. Could you please check it out?

eoghanmurray:
I've got errors running test suite on my computer even on the trunk.
What exactly are the tests that are failing for you?
Should the test suite have been run against this PR here?  I don't see any message.

packages/rrweb/src/record/mutation.ts
```diff
@@ -462,6 +462,14 @@ export default class MutationBuffer {
     if (isIgnored(m.target, this.mirror)) {
       return;
     }
+    let unattachedDoc;
+    try {
+      // avoid upsetting original document from a Content Security point of view
+      unattachedDoc = document.implementation.createHTMLDocument();
+    } catch (e) {
+      // fallback to more direct method
```
razor-x, line 470: Do we know when we will fallback? I'm curious what would trigger errors and the fallback case? Unsupported browsers?

eoghanmurray: I'm just not 100% about that `document.implementation.createHTMLDocument();` so took the prudent option of falling back to the previous method (which is working fine but wouldn't work for you from a CSP POV)

I just looked it up now and it seems fine:
https://caniuse.com/?search=implementation.createhtmldocument
E.g. it could fail on IE11 as there is a 'title' attribute missing
I don't think anyone is supporting IE11

You could add some reporting in there if it happens, or to add a custom commit on top (for your deployment) to revert this last commit.
I'm putting this out to a 'canary' environment today, but I don't have any reporting on whether the fallback kicks in; I'm happy enough that it worked in the rrweb test environment (Puppeteer/Chrome).

PR #853, opened 2022-03-06
fast-forward implementation v2: virtual dom optimization

Mark-Fenng:
### Motivation

In PR #241, we applied a `virtual parent optimization` to the fast-forward playback where we got a 10 times performance improvement. But this approach still has some unavoidable defects, e.g. #550, #529, #827.

The cause and solutions of these issues have been explained here: https://github.com/rrweb-io/rrweb/issues/548#issuecomment-826039286. I believe the virtual dom could be a good solution to solve these issues and it can also bring an approximately 2.5x performance boost.

### changes

This PR has a huge amount of code changes and I'll explain them to help understand.

1. In #419 and PR #613, we designed and implemented an ad-hoc DOM called 'rrdom' as an environment to replay the session. In this PR, I make some changes based on it and make it a browser runtime specially designed for virtual-dom. The feature of this runtime is that it is slim and only covers the dom operations used in the rrweb replayer. In other words, the virtual dom is implemented with rrdom.
2. The main mechanism of virtual dom optimization is described below. At the start of fast-forward process, a virtual dom tree is created which is a clone of the real dom tree. More precisely, the virtual dom is created right before the replayer starts to process the first node mutation event. After that, any node mutations are applied to the virtual dom tree. After all incremental events have been applied synchronously, the replayer goes into the `flush stage`. At this stage, the replayer executes a diff algorithm between the virtual-dom and the real dom, and applies all mutations to the real dom.
3. PR #241 uses a data structure called [treeIndex](https://github.com/rrweb-io/rrweb/blob/rrweb%401.1.2/packages/rrweb/src/utils.ts#L320-L512) to improve the performance of applying input events and scroll events.  This PR takes a similar approach to replace it. The key idea is to delay the handling of these types of events until the flush stage. Input events, scroll events, canvas mutations, and CSS style mutations are stored on their target virtual nodes at first. These data will be processed only after the real dom is updated in the flush stage. BTW, the virtual stylesheet created by #618 is also integrated into the virtual dom in this PR.
4. To improve the quality and reliability of this PR, the rrdom package has reached a high test coverage listed below:

File                | % Stmts | % Branch | % Funcs | % Lines |
--------------------|---------|----------|---------|---------|
All files           |   97.35 |    94.56 |   97.28 |   97.35 |
 diff.ts            |   92.34 |    91.33 |   91.66 |   92.15 |
 document-nodejs.ts |     100 |      100 |     100 |     100 |
 document.ts        |     100 |      100 |     100 |     100 |
 polyfill.ts        |   95.83 |      100 |    90.9 |   95.34 |
 style.ts           |     100 |    83.33 |     100 |     100 |
 virtual-dom.ts     |    97.6 |    91.78 |   89.28 |   98.37 |

Several integration test cases are also added to cover some types of replay scenes.

Some minor changes may relate to these PRs: #427, #481, #662, #666, #671, #756.


Yuyz0112:
@Mark-Fenng This is great! Could you describe the changed files' intention and give some suggestions for reviewing? Then we can have a better overview of the whole patch.

Mark-Fenng:
Sure, I will start to work on that!

Yuyz0112:
Technically, is it possible to use the virtual DOM during recording? I'm wondering whether we can implement session editing features based on it.

Mark-Fenng:
I think this gonna be a really interesting and valuable application scenario for rrdom

Mark-Fenng:
### code changes in rrdom package:
- document.ts
   - define interfaces: IRRNode, IRRDocument, IRRDocumentType, IRRElement, IRRText, IRRComment, IRRCDATASection. 
   - basic implementation of these interfaces: BaseRRNode, BaseRRDocument, BaseRRDocumentType, BaseRRElement, BaseRRText, BaseRRComment, BaseRRCDATASection
   - BaseRRMediaElement is an implementation of HTMLMediaElement
- virtual-dom.ts
   - This is a version of RRNode specially designed for virtual-dom optimization
   - class `RRNode` directly extends BaseRRNode
   - class `RRDocument` extends both RRNode and BaseRRDocument by using [mixins](https://www.typescriptlang.org/docs/handbook/mixins.html#constrained-mixins)
   - class `RRElement` , etc
   - The relationship of these interfaces and classes is shown as below: Untitled-2021-11-12-2134
- document-nodejs.ts
   - This is an implementation of RRDom designed for nodejs environment. The `functionality` keeps the same as the old version but code structures have been `refactored`.
   -  class `RRNode` directly extends BaseRRNode. More indexes or properties for more RRDom features will be added in the third stage. Please refer to:  rrdom's milestones #613 
   - Different from the virtual-dom version of rrdom, the nodejs version contains some APIs for selector queries and some dependencies only executable in nodejs.
- diff.ts
   - diff algorithm and virtual style part (#618).

### code changes in rrweb package:
- replayer/index.ts
   - replace the virtual parent optimization with virtual-dom optimization
- replayer/virtual-styles.ts
   - this part is moved to rrdom/diff.ts (@Juice10 There's a special change here:  SnapshotRule is removed because we don't have to build snapshot of style rules anymore with virtual-dom optimization) 
- utils.ts
   - the treeIndex data structure is removed as it's replaced by virtual-dom
- types.ts
   - remove some deprecated data types
- canvas
   - just try to solve the error `Error: @rollup/plugin-typescript TS1371: This import is never used as a value and must use 'import type' because the 'importsNotUsedAsValues' is set to 'error'` in rrweb-player. Because rrweb-player extends eslint rules from [svelte](https://svelte.dev/) and I don't know why I can't close its strict type check to rrweb package. There are many code changes that have been made for this reason as well e.g. packer/base.ts, record/observer.ts

### recommend review order
1. document.ts/document.test.ts
2. virtual-dom.ts/virtual-dom.test.ts
3. diff.ts/diff.test.ts
4. replayer/index.ts
5. document-nodejs.ts/document-nodejs.test.ts
6. other minor changes made for solving building errors in rrweb-player

### some aspects may need to care for reviewers
1. A new type RRNode is imported to replayer/index.ts and node mutations may be applied to either the real document tree or a RRDom tree. If a virtual node is added to a real dom tree or a real node is added to a virtual dom tree, this will be a serious problem.
2. Although there's no code change in rrweb-snapshot/rebuild.ts, all code in it is also run at virtual-dom environment during the fast-forward process.
3. A new virtual-dom tree is created and destroyed at every fast-forward process and as the replayer runs, a large amount of trees will be created. So that any risk of memory leak may accumulate and cause large amounts of memory to be used up.

Juice10:
> canvas
> just try to solve the error Error: @rollup/plugin-typescript TS1371: This import is never used as a value and must use 'import type' because the 'importsNotUsedAsValues' is set to 'error' in rrweb-player. Because rrweb-player extends eslint rules from [svelte](https://svelte.dev/) and I don't know why I can't close its strict type check to rrweb package. There are many code changes that have been made for this reason as well e.g. packer/base.ts, record/observer.ts

This is actually great, as it means if we'd ever want to migrate from rollup to vite we'd have to do this anyway.

Juice10:
I noticed rrweb is importing rrdom, and rrdom is importing rrweb. This is being done thanks to a rollup build step but it feels like a code smell to me. Is there any way we could break this loop (in the future)?

Mark-Fenng:
> I noticed rrweb is importing rrdom, and rrdom is importing rrweb.
> I've had some bad experiences with circular dependencies in the past. Is there any way we could break this loop (in the future)?

This concern is really a potential risk. Maybe move diff.ts to rrweb package or make these types declaration a third package😂. These are ways that I can come up with for now.

Juice10:
> > I noticed rrweb is importing rrdom, and rrdom is importing rrweb.
> > I've had some bad experiences with circular dependencies in the past. Is there any way we could break this loop (in the future)?
> 
> This concern is really a potential risk. Maybe move diff.ts to rrweb package or make these types declaration a third package😂. These are ways that I can come up with for now.

Hmmm, here are some more options I could come up with: move the types down to the rrweb-snapshot, since rrweb-snapshots is needed by all packages. Or move most of the rrweb types down to rrdom, since that is now also at the core of rrweb.

packages/rrweb/test/replayer.test.ts
```diff
@@ -223,6 +226,143 @@ describe('replayer', function () {
...
+      await contentDocument!.$eval(
+        '#container',
+        (element: Element) => element.scrollTop,
+      ),
+    ).toEqual(0);
+
+    const delay = 50;
+    // restart the replayer
+    await page.evaluate('replayer.play(0);');
+    await page.waitForTimeout(delay);
```
Juice10, line 251: Suggestion: You could use `await waitForRAF(page)` (`waitForRAF` needs to be imported from `utils.ts`). It guarantees a paint happened and is better for slow machines or running many tests in parallel.

Juice10 (changes requested):
What an absolute beast of a PR, great job @Mark-Fenng!
I had a couple small suggestions and questions I've added to the PR but they are all quite minor.
I think this is a very big step for rrweb, really well done!

packages/rrdom/src/document.ts
```diff
@@ -0,0 +1,697 @@
...
+      return super.toString('RRDocument');
+    }
+  };
+}
+
+export function BaseRRDocumentTypeImpl<
+  RRNode extends ConstrainedConstructor<IRRNode>
+>(RRNodeClass: RRNode) {
+  // @ts-ignore
+  return class BaseRRDocumentType
```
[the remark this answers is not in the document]
Juice10: I might have figured it out, looks like typescript has difficulties with the factory functions and their dynamic classes.

packages/rrdom/test/document.test.ts
```diff
@@ -0,0 +1,943 @@
...
+
+    it('should not implement appendChild', () => {
+      const parentNode = new RRNode();
+      const childNode = new RRNode();
+      expect(() =>
+        parentNode.appendChild(childNode),
+      ).toThrowErrorMatchingInlineSnapshot(
+        `"RRDomException: Failed to execute 'appendChild' on 'RRNode': This RRNode type does not support this method."`,
+      );
+    });
```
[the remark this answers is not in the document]
Mark-Fenng: Because most types of node don't support appendChild, for example TextNode, CommentNode and DocumentType

packages/rrdom/test/diff.test.ts
```diff
@@ -0,0 +1,1148 @@
...
+
+      diff(element, node, replayer);
+      expect(((element as Node) as SVGElement).getAttribute('xmlns')).toBe(
+        value,
+      );
+      expect(setAttributeNS).toHaveBeenCalledWith(
+        'http://www.w3.org/2000/xmlns/',
+        'xmlns',
+        value,
+      );
```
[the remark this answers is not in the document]
Mark-Fenng: Thanks! Really great suggestions!

packages/rrweb/src/replay/index.ts
```diff
@@ -693,33 +715,49 @@ export class Replayer {
...
-      (styleEl.sheet! as CSSStyleSheet).insertRule(injectStylesRules[idx], idx);
+    if (this.usingVirtualDom) {
+      const styleEl = this.virtualDom.createElement('style') as RRStyleElement;
+      styleEl.__sn = {
+        type: NodeType.Element,
+        tagName: 'style',
+        childNodes: [],
+        attributes: {},
+        id: this.virtualDom.notSerializedId,
+      };
```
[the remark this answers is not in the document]
Mark-Fenng: refactored! please take a look at this commit https://github.com/rrweb-io/rrweb/pull/853/commits/3e12f66293b9f193c40f15beb3c194daa2cf1fb4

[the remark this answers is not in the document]
Juice10: This is great! Thanks @Mark-Fenng

packages/rrdom/test/document.test.ts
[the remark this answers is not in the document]
Juice10: Totally makes sense!

Juice10 (approved):
Great stuff @Mark-Fenng, thanks for the changes!

Mark-Fenng:
I add a switch for the virtual-dom optimization for two reasons: 
1. If something goes wrong with the optimization, developers could simply turn it off as a workaround.
2. When we want to run the replayer on the rrdom nodejs runtime, rrdom (virtual dom) will run on rrdom (nodejs runtime). So we need to turn off the optimization to avoid this situation.

Juice10 (approved):

packages/rrdom/src/document.ts
```diff
@@ -0,0 +1,755 @@
+import { NodeType, serializedNodeWithId } from 'rrweb-snapshot';
+import { parseCSSText, camelize, toCSSText } from './style';
+
+export interface IRRNode {
+  __sn: serializedNodeWithId;
+  parentElement: IRRNode | null;
+  parentNode: IRRNode | null;
+  childNodes: IRRNode[];
+  ownerDocument: IRRDocument;
+  ELEMENT_NODE: 1;
```
Yuyz0112, line 10: Maybe we can define this as an enum and add comments about why we only port these two types(because we are an ad-hoc runtime and will have the minimal API?). I think this will be helpful for developers to understand the code.

packages/rrdom/src/document.ts
```diff
@@ -0,0 +1,755 @@
...
+  body: IRRElement | null;
+
+  head: IRRElement | null;
+
+  implementation: IRRDocument;
+
+  firstElementChild: IRRElement | null;
+
+  createDocument(
+    _namespace: string | null,
```
Yuyz0112, line 52: why starts with `_`?

packages/rrdom/src/document.ts
```diff
@@ -0,0 +1,755 @@
...
+          id,
+          type: this.RRNodeType,
+          textContent: '',
+        };
+        break;
+    }
+  }
+
+  public toString(nodeName?: string) {
+    return `${this.__sn?.id || ''} ${nodeName}`;
```
Yuyz0112, line 229: will this format `undefined` when nodeName is not passed in?

packages/rrdom/src/document.ts
```diff
@@ -0,0 +1,755 @@
...
+}
+
+export function BaseRRDocumentImpl<
+  RRNode extends ConstrainedConstructor<IRRNode>
+>(RRNodeClass: RRNode) {
+  return class BaseRRDocument extends RRNodeClass implements IRRDocument {
+    public readonly nodeType = 9;
+    public readonly RRNodeType = NodeType.Document;
+    public textContent: string | null = null;
+    _notSerializedId = -1; // used as an id to identify not serialized node
```
Yuyz0112, line 240: the name is a little confusing when I first read it

packages/rrdom/src/document.ts
```diff
@@ -0,0 +1,755 @@
...
+  public toString(nodeName?: string) {
+    return `${this.__sn?.id || ''} ${nodeName}`;
+  }
+}
+
+export function BaseRRDocumentImpl<
+  RRNode extends ConstrainedConstructor<IRRNode>
+>(RRNodeClass: RRNode) {
+  return class BaseRRDocument extends RRNodeClass implements IRRDocument {
+    public readonly nodeType = 9;
```
Yuyz0112, line 237: 9 is a magic number

packages/rrdom/src/document.ts
```diff
@@ -0,0 +1,755 @@
...
+    /**
+     * Every time the id is used, it will minus 1 automatically to avoid collisions.
+     */
+    public get notSerializedId(): number {
+      return this._notSerializedId--;
+    }
+
+    public get documentElement(): IRRElement | null {
+      return (
+        (this.childNodes.find(
```
Yuyz0112, line 251: I'm not sure whether other implementations are using linked-list or other data structures to store nodes, but ort current implementation using an array looks to cause a lot of iteration.

As a reference, [symbol-tree](https://github.com/jsdom/js-symbol-tree/blob/master/lib/SymbolTree.js) marked methods with time complexity.

packages/rrweb/src/replay/index.ts
```diff
@@ -5,7 +5,23 @@ import {
...
+  RRIFrameElement,
+  RRMediaElement,
+  RRCanvasElement,
+  StyleRuleType,
+  VirtualStyleRules,
+  createOrGetNode,
+  buildFromNode,
+  buildFromDom,
+  diff,
+} from 'rrdom/es/virtual-dom';
```
Yuyz0112, line 24: why we are import from a bundled path instead of `rrdom` itself?

packages/rrweb/src/replay/index.ts
```diff
@@ -102,6 +110,9 @@ export class Replayer {
 
   public config: playerConfig;
 
+  public usingVirtualDom = false;
```
Yuyz0112, line 113: it would be better if we have some comments here about the meaning of the flag. and the current naming is also a little confused with the config option.

Yuyz0112 (changes requested):

packages/rrdom/src/document.ts
Yuyz0112: Since we already have some performance improvement in this PR, we can treat this as an improvement in the following PRs with a benchmark suite.

packages/rrdom/src/document.ts
Juice10: Maybe `unserializedId`?

packages/rrdom/src/document.ts
Mark-Fenng: Because these parameters aren't actually used so I add `_` to disable lint warnings.

packages/rrdom/src/document.ts
Mark-Fenng: Good suggestion and I'll improve it in the next PR

Juice10 (approved):

packages/rrweb/src/replay/index.ts
Juice10: Would dynamic imports help in this situation? https://javascript.info/modules-dynamic-imports

Mark-Fenng: @Juice10 Would you like to explain a bit more about this idea?

packages/rrweb/test/events/style-sheet-rule-events.ts
```diff
@@ -154,6 +138,22 @@ const events: eventWithTime[] = [
...
+    timestamp: now + 400,
+  },
+  // mutation that adds style rule to existing stylesheet
+  {
+    data: {
+      id: 101,
+      adds: [
+        {
+          rule:
+            '.css-added-at-400-overwritten-at-3000 {border: 1px solid blue;}',
```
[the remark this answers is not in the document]
Juice10: Also the event above is now timestamp 400 and the css added is contains "500", that should be changed to "400"

Juice10 (changes requested):
Looks good, couple small css naming  things that need fixing since event order got switched

packages/rrweb/test/events/style-sheet-rule-events.ts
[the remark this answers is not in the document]
Mark-Fenng: Thanks, really great catch!

Yuyz0112:
@Mark-Fenng do you have any chance to resolve the conflicts, then I can merge this PR and looks into the bundle things.

Juice10:
@Yuyz0112 @Mark-Fenng its a pretty daunting merge which will include some refactoring in rrdom. I've started on it and have gotten most tests passing. There are still a couple that have been quite tricky to debug. I'm hoping to get it resolved this week

Juice10 (approved):

Yuyz0112 (approved):

PR #856, opened 2022-03-08
Upgrade Typescript to 4.6.2, and bump up other packages

Juice10:
While working on another PR I came across some bugs in Typescript that are fixed in a later version of Typescript. This PR updates Typescript to 4.6.2 and updates a bunch of other packages too.


Mark-Fenng (approved):
TS warnings in mutation.ts and observer.ts finally got fixed!

PR #859, opened 2022-03-10
Record canvas snapshots N times per second

Juice10:
# Problem & solution summary
Sometimes recording a canvas via regular means (`recordCanvas: true`) isn't possible or is to data intensive.
When regular canvas recording isn't feasible it's nice to be able to record full snapshots of the canvas on a period basis.

## Configuration
```js
rrwebRecord({
  emit,
  // records at a maximum of 10 frames per second
  // actual snapshot record count is probably lower for large canvases
  // as we wait till a snapshot is finished before we make a new one
  recordCanvas: 10,
});
```

## Problem explained
Recording snapshots is especially useful in the event that an rrweb recording gets initialized _after_ a canvas was added to the dom and was populated [as there is no way to read the initial state of these canvas events](https://twitter.com/Tojiro/status/1489651811518541824).

Another useful case is if a canvas gets repainted every requestAnimationFrame with the same (large) sources but doesn't actually change that much.
For example if you were to call:
```js
const fn = ()=>{
  canvasContext.drawImage(HUGE_IMAGE_SOURCE, 0,0);
  requestAnimationFrame(fn);
};
fn()
```
Every invocation of drawImage would get recorded, the `HUGE_IMAGE_SOURCE` would get base64 encoded and saved, rapidly ballooning the size of recorded events. Utilizing the current recording mechanism I was able to generate about 300mb worth of json events in just 10-20 seconds of recording.

## Performance
This feature uses web workers to minimize the impact of recording on the performance of the website being recorded.
If the computer experiences heavy load the web workers will slow down, the recorded frame rate will go down, but no frames will get dropped (because of the recording).

## Browser support
This feature uses OffscreenCanvas and isn't supported by every browser yet. According to [caniuse](https://caniuse.com/offscreencanvas) this OffscreenCanvas is currently supported by 77% of browsers with webgl(2) support hidden behind feature flags in Firefox and [pretty far along in Safari](https://bugs.webkit.org/show_bug.cgi). There is currently no polyfill for browsers that don't support OffscreenCanvas so any canvas snapshotting attempted in those browsers will fail gracefully.

# Other benefits of this PR
The web worker is currently used to do the heavy operation of converting a Canvas element to a base64 encoded string.
We are currently using [`canvas.toDataURL` ](https://github.com/rrweb-io/rrweb/blob/34d4108528396d2d69b4133d8003c2022b0c509d/packages/rrweb/src/record/observers/canvas/2d.ts#L40-L50) and wrapping it in a `setTimeout` in the hopes of not locking the main thread. Building on the work in this PR, we can move this to a web worker where supported.

Serializes & deserializes 2d canvas arguments (fixes https://github.com/rrweb-io/rrweb/issues/851)


Mark-Fenng (approved):
I'm not familiar with webGL so there are some parts I don't understand. But I built a demo to test this PR and it does solve the problem.

Juice10:
@Mark-Fenng Thanks Mark!

Juice10:
Thanks @DeanGracey! It's currently waiting on one more rrweb core team member review and then it should be good to go.
There are a number of other big pull requests in the pipeline which also take some time to review

DeanGracey:
Amazing, thanks @Juice10. We're currently thinking of patching rr-web in our system to get this functionality in. Do you have a rough idea when this would be released? If not too long we would prefer to wait for the release rather than patching

docs/recipes/canvas.md
```diff
@@ -11,6 +11,15 @@ rrweb.record({
 });
 ```
 
+Alternatively enable image snapshot recording of Canvas at a maximum of 15 frames per second：
+
+```js
+rrweb.record({
+  emit(event) {},
+  recordCanvas: 15,
```
Yuyz0112, line 19: does it make sense to move the fps option to our sampling object?

docs/recipes/canvas.zh_CN.md
```diff
@@ -12,6 +12,15 @@ rrweb.record({
 });
 ```
 
+或者启用每秒 15 帧的 Canvas 图像快照记录：
```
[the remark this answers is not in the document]
Juice10: I'm sorry if it doesn't make any sense 😅 I figured I'd use Google translate to put in something and then you, @Mark-Fenng or another community member could put in a suggestion for something that is correct.

docs/recipes/canvas.md
Juice10: Oh yes that totally makes sense, thanks for the suggestion @Yuyz0112
I think I have some time this Monday to move it over!

Juice10: https://github.com/rrweb-io/rrweb/pull/859/commits/3e6a901255d66d27de99fa303ba4a5b875ae50fb moves the config over to the sampling config. I think this makes much more sense, thanks for the suggestion @Yuyz0112!

ryanZiegler:
@Juice10 
When will this feature be merged into master? I can't wait to experience this feature.

Yuyz0112:
@ryanZiegler done

ryanZiegler:
@Yuyz0112 Thank you very much for your quick response 👍 👍 👍 !  Hope NPM can also update this version synchronously.

PR #863, opened 2022-03-23
Added scripts for saving and viewing snapshots

croqaz:
Hi. I wanted to share these 2 useful scripts that I actually use.
They can showcase how to use rrWeb, and the script for viewing snapshots is really useful in debug.
The scripts are kind of rough. I didn't want to use any special CLI flag parsing libraries, but maybe they are useful? Any suggestions are welcome.

### DIFF
--- a/packages/rrweb-snapshot/.gitignore
+++ b/packages/rrweb-snapshot/.gitignore
@@ -6,3 +6,4 @@ dist
 es
 lib
 temp
+*-snapshot.json
--- a/packages/rrweb-snapshot/package.json
+++ b/packages/rrweb-snapshot/package.json
@@ -9,6 +9,9 @@
     "test:watch": "jest --watch",
     "bundle": "rollup --config",
     "bundle:es-only": "cross-env ES_ONLY=true rollup --config",
+    "bundle:browser-only": "cross-env BROWSER_ONLY=true rollup --config",
+    "rebuild": "npm run bundle:browser-only && node scripts/rebuild.js",
+    "snapshot": "npm run bundle:browser-only && node scripts/snapshot.js",
     "dev": "yarn bundle:es-only --watch",
     "typings": "tsc -d --declarationDir typings",
     "prepublish": "npm run typings && npm run bundle"
@@ -49,7 +52,7 @@
     "jest": "^27.2.4",
     "jest-snapshot": "^23.6.0",
     "jsdom": "^16.4.0",
-    "puppeteer": "^1.15.0",
+    "puppeteer": "^13.5.1",
     "rollup": "^2.45.2",
     "rollup-plugin-terser": "^7.0.2",
     "ts-jest": "^27.0.5",
--- a/packages/rrweb-snapshot/rollup.config.js
+++ b/packages/rrweb-snapshot/rollup.config.js
@@ -6,7 +6,8 @@ function toMinPath(path) {
   return path.replace(/\.js$/, '.min.js');
 }
 
-let configs = [
+let configs = [];
+let es_configs = [
   // ES module - for building rrweb
   {
     input: './src/index.ts',
@@ -19,7 +20,7 @@ let configs = [
     ],
   },
 ];
-let extra_configs = [
+let browser_configs = [
   // browser
   {
     input: './src/index.ts',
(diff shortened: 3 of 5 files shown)


packages/rrweb-snapshot/script/rrweb_snapshot.js
```diff
@@ -0,0 +1,99 @@
...
+function delay(time) {
+    return new Promise((resolve) => setTimeout(resolve, time));
+}
+
+(async function main() {
+    const browser = await puppeteer.launch({
+        args: [
+            '--disable-breakpad',
+            '--disable-default-apps',
+            '--disable-features=IsolateOrigins,site-per-process',
```
croqaz, line 22: These flags are pretty important to allow loading of CORS resouces like images and CSS. A lot of capturing of styles and inline images depends on these args.

packages/rrweb-snapshot/script/rrweb_snapshot.js
```diff
@@ -0,0 +1,99 @@
...
+        await page.goto(url, { waitUntil: 'networkidle0', timeout: PAGE_TIMEOUT });
+        await delay(PAGE_DELAY);
+    } catch (err) {
+        console.error(err);
+        await browser.close();
+        return;
+    }
+
+    // hack all images with img.crossOrigin="anonymous" before calling snapshot
+    // this is MANDATORY to capture images from websites that host images on CDNs
```
croqaz, line 55: If we don't hack all the images with `crossOrigin = 'anonymous'`, a LOT of websites with images won't work, the images will not be saved inline.

packages/rrweb-snapshot/script/rrweb_view.js
```diff
@@ -0,0 +1,85 @@
...
+    const rrCode = await fs.promises.readFile(rrWeb, { encoding: 'utf8' });
+    let snap = await fs.promises.readFile(rrFile);
+    if (rrFile.endsWith('.gz')) {
+        snap = (await promisify(unzip)(snap)).toString();
+    }
+
+    await page.evaluate(`(function(){
+      console.log('Restoring the snaphot...');
+      ${rrCode};
+      rrwebSnapshot.rebuild(${snap}, {doc: document});
```
croqaz, line 73: Alternatively I'm thinking it makes sense to process the snapshot and generate HTML in a local /tmp folder and point Pupeteer at the local file.
The advantage is you can click on some links in the temp page and then return back to the snapshot. Currently the page is restored in a blank tab, so if you click on anything, you just can't go back to the snapshot, it's just a blank page.

croqaz:
Thank you for the comment @Juice10 , it sounds great and I implemented everything.
I also added a shortcut command for generating the `dist/rrweb-snapshot.js` file, I hope that's OK.
I think I only need to use the /temp/ file that you mentioned.

Yuyz0112:
@croqaz Thanks for the PR! Do you mind updating the script usage from `yarn snapshot` and `yarn rebuild` to `yarn repl snapshot` and `yarn repl rebuild`? This will help us keep the script's entry minimal.

If so, I can help with this change.

croqaz:
> @croqaz Thanks for the PR! Do you mind updating the script usage from `yarn snapshot` and `yarn rebuild` to `yarn repl snapshot` and `yarn repl rebuild`? This will help us keep the script's entry minimal.

I don't think I mind if that helps, but from I could look at the rest of the repl commands, they are located in packages/rrweb.
Doesn't that make running these scripts much more complicated? Or we should move them in that folder as well?

croqaz (commented):
Sorry I cannot comment for some reason.
I will work on this next week.

Juice10:
FYI: https://github.com/rrweb-io/rrweb/pull/1033 will get merged soon which will change the build command to `yarn build` and the location of the `rrweb-snapshot.js` to `dist/rrweb-snapshot.umd.cjs`. That will impact this PR a little bit

croqaz:
Hi folks! I don't use rrweb for a very long time and I won't be working on this PR anymore.
Not sure if I should close the PR... maybe somebody else can pickup the scripts and continue building? They personally helped me a lot, but maybe other people have a different way of working with this.
Anyway, feel free to close this! Cheers

PR #866, opened 2022-03-29
Don't serialize all cssRules if multiple text nodes exists

Juice10:
[Emotion](https://emotion.sh) adds css rules by appending text child nodes to a style element. This currently causes all cssRules to get serialized each time an extra rule gets added. That leads to huge stylesheets in replay (45mb+ in some cases) and can freeze the browser whenever more rules get added.

This PR only serializes _all_ `.cssRules` when one text node exists in a `style` element.

### DIFF
--- a/packages/rrweb-snapshot/test/snapshot.test.ts
+++ b/packages/rrweb-snapshot/test/snapshot.test.ts
@@ -1,5 +1,13 @@
+/**
+ * @jest-environment jsdom
+ */
 import { JSDOM } from 'jsdom';
-import { absoluteToStylesheet, _isBlockedElement } from '../src/snapshot';
+import {
+  absoluteToStylesheet,
+  serializeNodeWithId,
+  _isBlockedElement,
+} from '../src/snapshot';
+import { serializedNodeWithId } from '../src/types';
 
 describe('absolute url to stylesheet', () => {
   const href = 'http://localhost/css/style.css';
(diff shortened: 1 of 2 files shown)


packages/rrweb-snapshot/test/snapshot.test.ts
```diff
@@ -126,3 +134,48 @@ describe('isBlockedElement()', () => {
...
+      isStyle: true,
+      rootId: undefined,
+      textContent: 'section {color: blue;}body {color: red;}',
+      type: 3,
+    });
+  });
+
+  it('should serialize individual text nodes on stylesheets with multiple child nodes', () => {
+    const styleEl = render(`<style>body { color: red; }</style>`);
+    styleEl.append(document.createTextNode('section { color: blue; }'));
```
Mark-Fenng, line 173: If a rule is also inserted here, would it be recorded?

Juice10: Any rule added before the append would be wiped by the webbrowser. Any rule added after the append will not be picked up in rrweb-snapshot unfortunately. And I don't know of a good way of capturing those.
In rrweb these rule additions are captured by the observers.

This PR hopes that you either add css via the `.sheet.inserRule` api _or_ `.append(TEXT_NODE)` api. It's not foolproof but it's better than what we had.

Mark-Fenng (approved):

Juice10:
I mentioned that emotion.sh is what cases this. But I’ve seen the exact same behavior happen with vue.js too

Yuyz0112 (approved):

eoghanmurray:
@Juice10 Would a better solution be to run the `stringifyStylesheet` during `serializeElementNode` on <style> rather than `serializeTextNode` on it's children?

PR #868, opened 2022-04-04
Remove INode (`node.__sn`) and use Mirror as source of truth

Juice10:
Currently we can't record rrweb with rrweb because both the replay and recording use the `node.__sn` attribute which causes id collisions.

To move away from `INode`'s (nodes with `__sn` attribute) this PR uses the `Mirror` to save all serialized data, and id numbers.

This PR could potentially make it easier for multiple rrweb recordings to happen at the same time without the recording sessions needing to use the exact same ids for the same nodes.

### Breaking change
The mirror was moved to rrweb-snapshot, has its own class, and the `map` attribute isn't supported anymore. We could create a `.map` fallback to help with backward compatibility.

### DIFF
--- a/packages/rrweb-snapshot/src/utils.ts
+++ b/packages/rrweb-snapshot/src/utils.ts
@@ -1,12 +1,77 @@
-import { INode, MaskInputFn, MaskInputOptions } from './types';
+import {
+  idNodeMap,
+  MaskInputFn,
+  MaskInputOptions,
+  nodeIdMap,
+  nodeMetaMap,
+  serializedNode,
+} from './types';
 
-export function isElement(n: Node | INode): n is Element {
+export function isElement(n: Node): n is Element {
   return n.nodeType === n.ELEMENT_NODE;
 }
 
 export function isShadowRoot(n: Node): n is ShadowRoot {
   const host: Element | null = (n as ShadowRoot)?.host;
   return Boolean(host && host.shadowRoot && host.shadowRoot === n);
+}
+
+export class Mirror {
+  private idNodeMap: idNodeMap = new Map();
+  private nodeIdMap: nodeIdMap = new WeakMap();
+  private nodeMetaMap: nodeMetaMap = new WeakMap();
+
+  getId(n: Node | undefined | null): number {
+    if (!n) return -1;
+
+    const id = this.nodeIdMap.get(n);
+
+    // if n is not a serialized Node, use -1 as its id.
+    return id ?? -1;
+  }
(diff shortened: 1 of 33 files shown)


Mark-Fenng (changes requested):
👍A great patch that will solve many issues and inconveniences in the current branch.
There are only several typecast that have space to improve.

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -9,6 +16,64 @@ export function isShadowRoot(n: Node): n is ShadowRoot {
   return Boolean(host && host.shadowRoot && host.shadowRoot === n);
 }
 
+export class Mirror {
+  private idNodeMap: idNodeMap = new Map();
+  private nodeIdMap: nodeIdMap = new WeakMap();
+  private nodeMetaMap: nodeMetaMap = new WeakMap();
```
Yuyz0112, line 22: does it make sense to store `node -> serializedNodeWithId` in this map, and remove the nodeIdMap?

maintain two maps to be synced looks like something not robust.

Yuyz0112 (changes requested):

Yuyz0112:
This  PR is a lifesaver, `INode` may be one of the worst types I've designed:)

packages/rrweb/src/replay/index.ts
```diff
@@ -905,7 +906,7 @@ export class Replayer {
           d.adds.forEach((m) => this.treeIndex.add(m));
           d.texts.forEach((m) => {
             const target = this.mirror.getNode(m.id);
-            const parent = (target?.parentNode as unknown) as INode | null;
+            const parent = (target?.parentNode as unknown) as Node | null;
```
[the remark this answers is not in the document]
Juice10: Thanks @Mark-Fenng! I've removed as many as I could find in https://github.com/rrweb-io/rrweb/pull/868/commits/fa5dfebede6183ad46d0477d8c785cefa8ba7230

packages/rrweb-snapshot/src/utils.ts
Juice10: I'm happy to merge these two.

Naming-wise what do you think? I've been calling it `nodeMetaMap` and `getMeta`, but that's inconsistent with the `sn` & `serializedNode` naming we've been using. I could rename it to something like `nodeSerializedNodeMap` and `getSN`. Would be more inline with what we had before.

Mark-Fenng (approved):

PR #871, opened 2022-04-06
Allow people to disable blockClass and speed up recording

Juice10:
When recording pages with tens of thousands of dom elements rrweb, the `isBlocked` checks can slow down the recording. This PR allows you to disable that check and speed up the recording on those large pages.


Juice10:
Recording is much faster now thanks to https://github.com/rrweb-io/rrweb/pull/903
Closing this PR

PR #881, opened 2022-04-28
Fix issue with scrolling not working

eoghanmurray:
Fix issue where only indication I could see in any attribute that the document was scrolling was on `window.pageYOffset`, so we hadn't been able to replay scrolling

Website was https://www.malibal.com/boutique/pc/configurePrd.asp?idproduct=2061

I've no idea what conditions make this the case.

### DIFF
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -9,6 +9,7 @@ import {
   on,
   getWindowWidth,
   getWindowHeight,
+  getWindowScroll,
   polyfill,
   hasShadowRoot,
   isSerializedIframe,
@@ -296,22 +297,7 @@ function record<T = eventWithTime>(
         type: EventType.FullSnapshot,
         data: {
           node,
-          initialOffset: {
-            left:
-              window.pageXOffset !== undefined
-                ? window.pageXOffset
-                : document?.documentElement.scrollLeft ||
-                  document?.body?.parentElement?.scrollLeft ||
-                  document?.body.scrollLeft ||
-                  0,
-            top:
-              window.pageYOffset !== undefined
-                ? window.pageYOffset
-                : document?.documentElement.scrollTop ||
-                  document?.body?.parentElement?.scrollTop ||
-                  document?.body.scrollTop ||
-                  0,
-          },
+          initialOffset: getWindowScroll(window),
         },
       }),
     );
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -4,6 +4,7 @@ import {
   throttle,
   on,
   hookSetter,
+  getWindowScroll,
   getWindowHeight,
   getWindowWidth,
   isBlocked,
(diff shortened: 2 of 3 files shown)


Yuyz0112 (commented):
LGTM, we can merge this after the CI passed

Mark-Fenng (commented):
The CI failed just because of some TS errors. I fixed these errors in the comments and wish these can help you.

packages/rrweb/src/record/index.ts
```diff
@@ -296,22 +297,7 @@ function record<T = eventWithTime>(
...
-                  0,
-            top:
-              window.pageYOffset !== undefined
-                ? window.pageYOffset
-                : document?.documentElement.scrollTop ||
-                  document?.body?.parentElement?.scrollTop ||
-                  document?.body.scrollTop ||
-                  0,
-          },
+          initialOffset: getWindowScroll(window),
```
Mark-Fenng, line 300:
```suggestion
          initialOffset: getWindowScroll(window.document),
```

packages/rrweb/src/record/observer.ts
```diff
@@ -272,11 +273,11 @@ export function initScrollObserver({
     }
     const id = mirror.getId(target as Node);
     if (target === doc) {
-      const scrollEl = (doc.scrollingElement || doc.documentElement)!;
+      const scrollLeftTop = getWindowScroll(doc.defaultView);
```
Mark-Fenng, line 276:
```suggestion
      const scrollLeftTop = getWindowScroll(doc);
```

packages/rrweb/src/utils.ts
```diff
@@ -169,6 +169,28 @@ export function patch(
...
+      doc.scrollingElement ? doc.scrollingElement.scrollTop
+      : win.pageYOffset !== undefined
+      ? win.pageYOffset
+      : doc?.documentElement.scrollTop ||
+      doc?.body?.parentElement?.scrollTop ||
+      doc?.body.scrollTop ||
+      0,
+  }
+}
+
```
Mark-Fenng, line 193:
```suggestion
export function getWindowScroll(doc: Document) {
  const win = doc.defaultView;
  return {
    left: doc.scrollingElement
      ? doc.scrollingElement.scrollLeft
      : win?.pageXOffset !== undefined
      ? win.pageXOffset
      : doc.documentElement.scrollLeft ||
        doc.body?.parentElement?.scrollLeft ||
        doc.body.scrollLeft ||
        0,
    top: doc.scrollingElement
      ? doc.scrollingElement.scrollTop
      : win?.pageYOffset !== undefined
      ? win.pageYOffset
      : doc.documentElement.scrollTop ||
        doc.body?.parentElement?.scrollTop ||
        doc.body.scrollTop ||
        0,
  };
}
```

packages/rrweb/src/record/index.ts
eoghanmurray: Hi Mark, thanks for this review!  What is the idea behind switching to document rather than window?  I'm happy with either, just curious.  
We could probably change it to `getDocumentScroll` too; I'm just wondering about the semantics of it.

Mark-Fenng: `document.defaultView` is nullable but `window.document` always exists.
If the parameter is a window and the window is null, the function can't retrieve a document object from it (`win.document = null.document`). The document and window can both be null. This is the reason those TS errors occurred.
I think renaming with `getDocumentScroll` is a good idea.

Mark-Fenng:
Close this because it is a duplicate of #1054

PR #885, opened 2022-05-04
Perf: Apply the latest text mutation only

Juice10:
Some frameworks dynamically append styles to a stylesheet via `styleEl.textContent = styleEl.textContent + newStyleRules`
This can end up in a barrage of expanding text mutations on the same element.

This PR makes sure only the last `textContent` change gets applied, as any earlier `textContent` changes would be overwritten anyway.

### DIFF
--- a/packages/rrweb/src/utils.ts
+++ b/packages/rrweb/src/utils.ts
@@ -594,12 +594,16 @@ export function hasShadowRoot<T extends Node>(
  * @returns {textMutation[]} The filtered text mutations.
  */
 export function uniqueTextMutations(mutations: textMutation[]): textMutation[] {
-  const textMuatationsMap = new Map<number, textMutation>();
-  mutations
-    .slice()
-    .reverse()
-    .forEach((text) => {
-      if (!textMuatationsMap.has(text.id)) textMuatationsMap.set(text.id, text);
-    });
-  return Array.from(textMuatationsMap.values());
+  const idSet = new Set<number>();
+  const uniqueMutations: textMutation[] = [];
+
+  for (let i = mutations.length; i--; ) {
+    const mutation = mutations[i];
+    if (!idSet.has(mutation.id)) {
+      uniqueMutations.push(mutation);
+      idSet.add(mutation.id);
+    }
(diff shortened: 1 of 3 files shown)


Mark-Fenng (approved):

packages/rrweb/src/utils.ts
```diff
@@ -587,3 +587,19 @@ export function hasShadowRoot<T extends Node>(
...
+
+/**
+ * Returns the latest mutation in the queue for each node.
+ * @param  {textMutation[]} mutations The text mutations to filter.
+ * @returns {textMutation[]} The filtered text mutations.
+ */
+export function uniqueTextMutations(mutations: textMutation[]): textMutation[] {
+  const textMuatationsMap = new Map<number, textMutation>();
+  mutations
+    .slice()
```
eoghanmurray, line 599: why the slice?  Is it to produce a copy?

eoghanmurray (approved):
LGTM

packages/rrweb/src/utils.ts
Juice10: That's correct, I was thinking of doing `[...mutations].reverse()` as it might be more readable, but I wasn't sure of it's performance implications so I went with what I knew, `slice`

Yuyz0112: maybe we can do this in one iteration?

```ts
const idSet = new Set<string>();
const uniqueMutations = [];

for (const mutation of mutations) {
  if (!idSet.has(mutation.id)) {
    uniqueMutations.push(mutation)
    idSet.add(mutation.id)
  }
}
```

Juice10: Ooh yes thats a good idea, lets do that with a reverse for loop so we only get the latest item

Juice10: Done: https://github.com/rrweb-io/rrweb/pull/885/commits/f1bf52ea8e50ef3c819827e67340fb8da74cb8d8

Yuyz0112:
LGTM!

PR #886, opened 2022-05-04
Test: Stylesheet append text node

Juice10:
Added a test to make sure appending child text elements doesn't get broken in stylesheets.


Yuyz0112 (approved):
LGTM

PR #865, opened 2022-03-25
Apply textContent on flush

Juice10:
When scrubbing through replays I bumped into the issue that `textContent` doesn't always get applied on stylesheets.
This PR fixes that issue and adds tests for a number of edge cases

PR #887, opened 2022-05-05
test: add a test case for StyleSheet mutation exceptions while fast-forwarding

Mark-Fenng:
This PR is related to a bug that @Juice10 and I found at the last core team meeting. 
image
The added test case can reproduce the problem but I haven't found a solution yet. So that the CI will absolutely fail and it is expected.

The reason for this bug is similar to PR #865. When childNodes of a style element are mutated (append or remove), all applied stylesheet rules programmatically are lost automatically.


Juice10:
Thanks for adding this @Mark-Fenng!

PR #894, opened 2022-05-11
isBlocked factors in the selector

dbseel:
Support a list of blocked classes in a CSS selector (blockSelector) in addition to the blockClass.
Also ensure only node types are passed into contains.
Also support providing a list of CSS properties to ignore on set/remove stylesheet mutations


IMFIL:
@Yuyz0112 Hey man, we've added a bunch of changes to Rrweb as we are using it on thousands of websites and noticed some flaws. Let us know if certain things are missing for this to get merged in.

Juice10:
Thanks for the PR @dbseel! When refactoring and speeding up `isBlocked` for https://github.com/rrweb-io/rrweb/pull/903 I also noticed this was missing from rrweb. Happy you've created a PR for it.
Because https://github.com/rrweb-io/rrweb/pull/903 has checked the guts of `isBlocked` drastically I expect it to create some merge issues for this branch once it gets merged in.

Juice10:
Hey @dbseel, any chance you could fix the merge issues?

dbseel:
@Juice10 Done. Can we get this merged in? It has been sitting here 3 months, don't want to have to rebase again.

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -276,27 +276,39 @@ export function transformAttribute(
...
-  if (blockSelector) {
-    return element.matches(blockSelector);
+  if (node.nodeType === node.TEXT_NODE) {
+    // check parent node since text node do not have class name
+    return _isBlockedElement(node.parentNode, blockClass, blockSelector); 
   }
-
-  return false;
+  return _isBlockedElement(node.parentNode, blockClass, blockSelector);
 }
```
Juice10, line 312: This reverts the performance optimizations from https://github.com/rrweb-io/rrweb/pull/903 and I'm not quite sure what extra functionality this brings.

packages/rrweb/src/utils.ts
```diff
@@ -203,14 +205,17 @@ export function isBlocked(
...
   if (typeof blockClass === 'string') {
     if (el.classList.contains(blockClass)) return true;
     if (checkAncestors && el.closest('.' + blockClass) !== null) return true;
   } else {
     if (classMatchesRegex(el, blockClass, checkAncestors)) return true;
   }
-  return false;
+  if (blockSelector) {
+    if ((node as HTMLElement).matches(blockSelector)) return true;
+  }
```
Juice10, line 217: If `checkAncestors` is true we should also check to see if any of the parents of this node are blocked

```suggestion
  if (blockSelector) {
    if ((node as HTMLElement).matches(blockSelector)) return true;
    if (checkAncestors && el.closest(blockSelector) !== null) return true;
  }
```

packages/rrweb/src/utils.ts
```diff
@@ -203,14 +205,17 @@ export function isBlocked(
...
+  return isBlocked(node.parentNode, blockClass, blockSelector, checkAncestors);
```
Juice10, line 218: Parents where already checked if `checkAncestors` was set to true, no need to go further up the tree
```suggestion
  return false;
```

packages/rrweb-snapshot/src/snapshot.ts
Juice10, line 312: This reverts the performance optimizations from https://github.com/rrweb-io/rrweb/pull/903 and I'm not quite sure what extra functionality this brings.

Maybe we should revert this code?

Juice10 (changes requested):
Thanks for rebasing @dbseel, apologies that this has been sitting for so long. Thank you for your fixes! I'm going to do my best to get this merged ASAP for you

I noticed `ignoreCSSAttributes` isn't documented yet in https://github.com/rrweb-io/rrweb/blob/5f59f9171e481c78d08a1008186cff31a40ee199/guide.md#options could you add one to `guide.md` and a placeholder (or maybe google translated) to the Chinese [`guide.zh_CN.md`](https://github.com/rrweb-io/rrweb/blob/5f59f9171e481c78d08a1008186cff31a40ee199/guide.zh_CN.md#配置参数)

Juice10 (approved):
Thanks for this @dbseel! Really appreciate your contribution. Apologies that it took so long for us to review it

dbseel:
> Thanks for this @dbseel! Really appreciate your contribution. Apologies that it took so long for us to review it

No problem! Are you able to merge? I don't have permissions.

Juice10:
I don't but I've messaged the other members of the core team, so it should get merged soon

Yuyz0112 (approved):
good job!

eoghanmurray:
Late to the party here!

@dbseel what are the use cases for `ignoreCSSAttributes`?  I see `color` in the test case but curious as to the others.  Is the motivation for this privacy, or performance (like slimDOM options).

Also wondering whether `ignoreCSSProperties` would have been a better name?

dbseel:
> @dbseel what are the use cases for `ignoreCSSAttributes`?
We have encountered some websites where certain attributes were constantly being added/removed/added to the DOM, but they did not affect the replay. So there was a huge performance improvement by ignoring these attributes.

eoghanmurray:
These are CSS rule attributes rather than DOM attributes though?
I guess just thinking it must be very unusual that rules would be continually added/removed.
Was it a standard part of some framework, or just some once off craziness on a particular website?

dbseel:
> These are CSS rule attributes rather than DOM attributes though? I guess just thinking it must be very unusual that rules would be continually added/removed. Was it a standard part of some framework, or just some once off craziness on a particular website?

In our case a website was using quantum metrics. This plugin was adding/removing attributes thousands of times a second. Attributes like this: --quantum-metric-background-image

eoghanmurray:
I've just come across this again today; I notice that `ignoreCSSAttributes` doesn't omit them from capture from the initial snapshot. This wouldn't be significant for your usecase of mutations to `--quantum-metric-background-image`, but for completeness does anyone think it needs to be added?

PR #912, opened 2022-06-12
use nx as task runner

Yuyz0112:
The main strategy is to keep using Lerna's bootstrapping and publishing features, but use Nx for the fast task scheduling to speed up Lerna workspaces.


Yuyz0112:
going to try turborepo

PR #915, opened 2022-06-14
Added null check on iframe.contentWindow

shanptl12:
Getting this error "cannot read property on null (reading scrollTo)". Stacktrace shows error on this line.
Screen Shot 2022-06-14 at 1 23 47 PM


Yuyz0112:
Thanks! But actually, this iframe is created by the replayer, so it should exist at that moment.

Could you show the code to reproduce this?

Yuyz0112:
I think @Mark-Fenng 's latest commit: https://github.com/rrweb-io/rrweb/commit/5f59f9171e481c78d08a1008186cff31a40ee199 already pick this in, thanks and I'm going to close this PR.

PR #900, opened 2022-05-19
Chore: Add issue/pr template and general housekeeping tools and docs

Juice10:
Adds the following:
- Bug template
- Feature request template
- PR template
- Coding style guide
- Markdown style guide
- Homogenique linting cross-project
- Fixes basic linting errors (not all of them, but its a start)

A lot of the templates and docs are based off of the Electron project which does quite a good job IMO of guiding people to do the right thing for the project. Some things (like having people lint) might be a bit pre-mature for us, but its a good goal to work towards.

PR #920, opened 2022-06-25
refactor: eliminate eslint errors

Mark-Fenng:
In PR #900, we replaceed tslint with eslint with stricter lint rues, which is an excellent start to improving code quality. But unfixed lint errors keep slowing down our development speed. I create this PR to make it a base for more fixing code. This can also help avoid conflicts over the changes we all made. Welcome to fix more eslint errors based on this PR.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,5 +19,8 @@ module.exports = {
   plugins: ['@typescript-eslint', 'eslint-plugin-tsdoc'],
   rules: {
     'tsdoc/syntax': 'warn',
+    '@typescript-eslint/unbound-method': 'off',
+    '@typescript-eslint/no-unsafe-member-access': 'off',
+    '@typescript-eslint/restrict-template-expressions': 'off',
   },
 };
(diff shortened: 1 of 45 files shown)


Yuyz0112:
nice

Juice10:
You are an absolute hero @Mark-Fenng! Thank you for this! hero image

.eslintrc.js
```diff
@@ -19,5 +19,8 @@ module.exports = {
   plugins: ['@typescript-eslint', 'eslint-plugin-tsdoc'],
   rules: {
     'tsdoc/syntax': 'warn',
+    '@typescript-eslint/unbound-method': 'off',
```
Juice10, line 22: I have bumped into the issue before where I referenced an unbound method, and forgot to call it via apply. This rule doesn't really fix that, it just disallows accessing those methods in general. I'm wondering if there is a something in-between. Maybe set this to a warning instead of off?

.eslintrc.js
```diff
@@ -19,5 +19,8 @@ module.exports = {
...
+    '@typescript-eslint/no-unsafe-member-access': 'off',
```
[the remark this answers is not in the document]
Mark-Fenng: awesome! I learned something from it.

.eslintrc.js
Mark-Fenng: I agree with your suggestion about turning it into a warning level.

Yuyz0112 (approved):
I think we can remove tslint in a following PR

PR #926, opened 2022-06-30
DRAFT: Inline image edge cases

Juice10:
Add more test coverage for inlining images.
Also fixes recording of iframe mutations once the page gets reloaded


Juice10:
#927 already includes these tests and fixes

PR #927, opened 2022-07-01
Handle negative ids in rrdom correctly + extra tests

Juice10:
When iframes get inserted they create untracked elements, both on the dom and rrdom side.
Because they are untracked they generate negative numbers when fetching the id from mirror.
This creates a problem when comparing and fetching ids across mirrors.
This PR tries to get away from using negative ids as much as possible in rrdom's comparisons.

PR also adds more test coverage for inlining images.
Also fixes recording of iframe mutations once the page gets reloaded


Mark-Fenng:
Is this PR also a fix for #917?
I also encountered the problem caused by negative ids and this fix is really nice!

packages/rrdom/src/diff.ts
```diff
@@ -259,37 +259,61 @@ function diffChildren(
...
     indexInOld;
   while (oldStartIndex <= oldEndIndex && newStartIndex <= newEndIndex) {
+    const oldStartId = replayer.mirror.getId(oldStartNode);
+    const oldEndId = replayer.mirror.getId(oldEndNode);
+    const newStartId = rrnodeMirror.getId(newStartNode);
+    const newEndId = rrnodeMirror.getId(newEndNode);
+    console.log('oldStartId', oldStartId);
+    console.log('oldEndId', oldEndId);
+    console.log('newStartId', newStartId);
+    console.log('newEndId', newEndId);
```
Mark-Fenng, line 269:
```suggestion
```

Mark-Fenng (approved):
The negative ids part looks good to me.

Juice10:
> Is this PR also a fix for #917?
> I also encountered the problem caused by negative ids and this fix is really nice!

I don't think so, I tried replaying the [JSON Blobs](https://jsonblob.com/987252543365332992) on my branch and on [rrwebdebug](https://rrwebdebug.com/play/index.html?url=https%3A%2F%2Fjsonblob.com%2F987252543365332992&version=0.7.14&play=on) and they look the same.

Juice10:
I set the first unserialised id to be -2 so its clear when an id gets returned from the mirror if it is a mirror miss or an unserialised id

Mark-Fenng (approved):
Congrats, you figured out the method to fix the test failure. Your solution also looks good.

Juice10:
Thanks @Mark-Fenng!

Yuyz0112 (approved):
Nice patch! Let's merge this after a rebase.

PR #939, opened 2022-07-19
Guard access to frame.contentWindow 

ryan953:
Instead of relaying on TS [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) rrweb should explicitly check that `contentWindow` is set. This will avoid errors when the `root` of a Replayer instance is not connected to the window.


ryan953:
Seems like eslint is unhappy but i'm not sure why. The output I see is:

```
0 ESLint error(s) and 19 ESLint warning(s) found in pull request changed files.
0 ESLint error(s) and 56 ESLint warning(s) found in files outside of the pull request.
✖  fatal     Resource not accessible by integration
```

Yuyz0112:
The eslint config issue has been fixed by @Mark-Fenng.

For this PR, I think we intend to use the non-null assertion here because this iframe is just the replayer iframe built by rrweb, and if it does not have `contentWindow`, then all the functions could not work.

Could you tell us the case when the `contentWindow` will be null?

ryan953:
I was experimenting with instantiating a `Replayer` instance so I could run the recording in the background, hidden from the user, and pull out some stats & data. Specifically I wanted to read the state of the recorded html at certain timestamps.

Simplified the setup is basically:
```
const root = document.createElement('div');
const replayer = new Replayer(events, {
  root,
  useVirtualDom: true, // for visibility, this defaults to `true`
  speed: 99999, // arbitrary high number
  skipInactive: true,
  mouseTail: false,
  plugins: [new MyPlugin()],
});

const lastEvent = events[events.length - 1];
replayer.pause(lastEvent.timestamp);
```

Currently I'm attaching that `root` div to the document, and making it hidden:
```
const {style} = root;
style.position = 'fixed';
style.inset = '0';
style.width = '0';
style.height = '0';
style.overflow = 'hidden';

document.body.appendChild(root);
```
It's generally working with this, probably with big thanks to the `useVirtualDom` flag. But I've noticed with some recordings that I can still get `scrollTo of undefined` errors if the cleanup happens too quickly for example. And cleanup is required otherwise i'll have another whole copy of the recorded webpage sitting in my dom.

So I was thinking I should try to get to the source of the issue... I'd rather not have to pick arbitrary high `speed` numbers and it would be good to guarantee that the background processing is not able to affect the performance of the window document either.

So that's where I'm coming from, if there are other ways around it that'd be interesting to me as well. LMK!

Yuyz0112:
@ryan953 Thanks for the detailed explanation! This use case looks reasonable to me, the only minor change is I think you can use `?.` to replace `!.` which is the optional chaining syntax. 

BTW, I would like to have @Mark-Fenng do a double check here. When using the virtual DOM, the replayer's iframe DOM was not virtualized, so it's the expected behavior to get this kind of error right?

Mark-Fenng:
When I tried your setup, I found that the replayer's content was empty because of this guard
image
The replayer's iframe DOM wasn't virtualized also because of similar guards.
I'm not sure whether this hidden replayer can pull out your expected info.
I think the sync-replayer https://github.com/rrweb-io/rrweb/blob/trim/packages/rrweb/src/replay/sync-replayer.ts can be a choice even though it's still under development.

ryan953:
Woah, the `trim` branch looks like it's building up the same kind of infrastructure that I would like to leverage. 

https://github.com/rrweb-io/rrweb/blob/trim/packages/rrweb-cutter/src/index.ts#L30-L57 in particular looks like the same kind of callback that I'm doing too, in terms of inspecting the DOM as the event stream is getting processed!

BTW that event-cutting feature looks really useful too. Lots of benefits, like being able to get a snapshot for any timestamp and then start streaming data to the browser from that point on. Less pressure to configure frequent snapshots upfront.

I think I will watch this branch and I would just `import {SyncReplayer}` to get the job done when it's available.

Mark-Fenng:
@ryan953 Thanks for letting me know I'm developing things in the correct direction! I'll try to get done as soon as possible.
I think this PR can be merged because more guards could be safer in case there are more scenarios we don't know yet.

Issue #933, opened 2022-07-09
[Bug]: Custom HTML elements attach a Shadow DOM


pavlogavryliuk:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb-snapshot

### Expected Behavior

Looks like rrweb is processing incorrectly custom elements.

**Custom elements** are not a part of Shadow DOM, but during recording, all these elements receive the following options: `isShadow` and `isShadowHost`. In some cases it may break replay.

### Recording

Custom elements receive `isShadow` and `isShadowHost` options.

**Custom HTML element:**
`
<community_navigation-global-navigation-trigger className="cAltToggleNav slds-icon_x-small" data-aura-rendered-by="367:0">
  <lightning-button-icon></lightning-button-icon>
  <community_navigation-back-navigation-button></community_navigation-back-navigation-button>
</community_navigation-global-navigation-trigger>
`

**The same elements at snapshot:**
`
{"type": 2,
"tagName": "community_navigation-global-navigation-trigger",
                                            "attributes": {
                                                "class": "cAltToggleNav slds-icon_x-small",
                                                "data-aura-rendered-by": "367:0"
                                            },
                                            "childNodes": [{
                                                "type": 2,
                                                "tagName": "lightning-button-icon",
                                                "attributes": {},
                                                "childNodes": [],
                                                "id": 167,
                                                "isShadowHost": true,
                                                "isShadow": true
                                            }, {
                                                "type": 2,
                                                "tagName": "community_navigation-back-navigation-button",
                                                "attributes": {},
                                                "childNodes": [],
                                                "id": 175,
                                                "**isShadowHost**": true,
                                                "isShadow": true
                                            }],
                                            "id": 166,
                                            "isShadowHost": true
                                        }
`

### Replay
On investigate how Replayer handle **isShadowHost** and **isShadow**  options i came across a next comment in the code:

> Since node is newly rebuilt, it should be a normal element without shadowRoot. But if there are some weird situations that has defined custom element in the scope before we rebuild node, it may register the shadowRoot earlier. The logic in the 'else' block is just a try-my-best solution for the corner case, please let we know if it is wrong and we can remove it.

`
if (n.isShadowHost) {
  if (!node.shadowRoot) {
    node.attachShadow({ mode: 'open' });
  } else {
    while (node.shadowRoot.firstChild) {
[...]

PR #942, opened 2022-07-21
fix issue #933

Mark-Fenng:
This PR is trying to fix shadow dom recording problems #933 caused by [shadydom](https://github.com/webcomponents/polyfills/tree/master/packages/shadydom) and [@lwc/synthetic-shadow](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/synthetic-shadow). These two packages are all shadow dom polyfills but don't act exactly like native shadow dom, e.g. they don't provide style isolation.
The main idea of this PR is that on the recording side, we record poly-filled shadow doms as if they are native shadow doms but on the replayer side, we rebuild them as normal elements. That is we don't add `isShadowHost` and `isShadow` properties to them in the recorded results.

### DIFF
--- a/packages/rrweb-snapshot/src/utils.ts
+++ b/packages/rrweb-snapshot/src/utils.ts
@@ -20,8 +20,22 @@ export function isShadowRoot(n: Node): n is ShadowRoot {
  * To fix the issue https://github.com/rrweb-io/rrweb/issues/933.
  * Some websites use polyfilled shadow dom and this function is used to detect this situation.
  */
-export function isNativeShadowDom() {
-  return window.ShadowRoot?.toString().indexOf('[native code]') > -1;
+export function isNativeShadowDom(shadowRoot: ShadowRoot) {
+  return (
+    /**
+     * Some websites use shady dom https://github.com/webcomponents/polyfills/tree/master/packages/shadydom to polyfill the shadow dom and this will cause some issues during the recording.
+     * This statement is to determine whether the current website is polyfilled by shadydom or not.
+     */
+    !(window as Window & typeof globalThis & { ShadyDOM: object }).ShadyDOM &&
+    shadowRoot &&
+    /**
+     * https://github.com/salesforce/lwc/blob/v2.20.3/packages/@lwc/shared/src/keys.ts#L9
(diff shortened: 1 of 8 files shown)


Mark-Fenng:
@pavlogavryliuk @lele0108 need your reviews too

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -16,6 +16,28 @@ export function isShadowRoot(n: Node): n is ShadowRoot {
   return Boolean(host?.shadowRoot === n);
 }
 
+/**
+ * To fix the issue https://github.com/rrweb-io/rrweb/issues/933.
+ * Some websites use polyfilled shadow dom and this function is used to detect this situation.
+ */
+export function isNativeShadowDom(shadowRoot: ShadowRoot) {
```
lele0108, line 23: I have not tested this extensively but based on YouTube's usage of `shadyDOM` and #933 this line might be enough to check native shadow DOM or polyfill:

`Object.prototype.toString.call(shadowRoot) === '[object ShadowRoot]'`

Otherwise for polyfill I notice that it returns `[object DocumentFragment]`

lele0108: For `shadyDOM` heuristic being used, shady by default does not fill if there is native shadowDOM support unless the `force` flag is set to true in `settings` object.

lele0108:
@Mark-Fenng Nice PR!

I have been testing on Youtube.com (uses shadyDOM force fill) and notice that this captures more of the page correctly but does not rebuild mutations correctly. I suspect it has to do with the fact that Youtube will patch `childNodes` getter on `Node` objects so both `snapshot` and MutationObserver serialization is being affected. Would have to investigate more to see understand why.

pavlogavryliuk:
@Mark-Fenng i have been testing on google partner page, works fine. Thank you for your work. It's really cool. 💪🏼💪🏼💪🏼

image

packages/rrweb-snapshot/src/utils.ts
Mark-Fenng: Your solution is better! Thanks for the suggestion.

Yuyz0112:
great job!

PR #944, opened 2022-07-24
fix: canvas data in iframe wasn't applied in the fast-forward mode

Mark-Fenng:
In #853, I implemented the virtual dom for the fast-forward mode.  But there is a corner case I didn't deal with.
If a canvas element is created with image data (serialized to `rr_dataURL`) in an iframe, it will be rebuilt as an empty canvas without the initial image data, in the fast-forward mode (using Virtual Dom optimization).
A simple test case is also added in this PR to reproduce this corner case.

### DIFF
--- a/packages/rrdom/src/diff.ts
+++ b/packages/rrdom/src/diff.ts
@@ -136,14 +136,27 @@ export function diff(
           break;
         }
         case 'CANVAS':
-          (newTree as RRCanvasElement).canvasMutations.forEach(
-            (canvasMutation) =>
+          {
+            const rrCanvasElement = newTree as RRCanvasElement;
+            // This canvas element is created with image data in an iframe element.
+            if (rrCanvasElement.rr_dataURL !== null) {
+              const image = document.createElement('img');
+              image.src = rrCanvasElement.rr_dataURL;
+              image.onload = () => {
+                const ctx = (oldElement as HTMLCanvasElement).getContext('2d');
+                if (ctx) {
+                  ctx.drawImage(image, 0, 0, image.width, image.height);
+                }
+              };
+            }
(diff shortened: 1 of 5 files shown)


packages/rrdom/src/diff.ts
```diff
@@ -136,14 +136,27 @@ export function diff(
...
+            // This canvas element is created with initial data in an iframe element. https://github.com/rrweb-io/rrweb/pull/944
+            if (rrCanvasElement.rr_dataURL !== null) {
+              const image = document.createElement('img');
+              image.src = rrCanvasElement.rr_dataURL;
+              image.onload = () => {
+                const ctx = (oldElement as HTMLCanvasElement).getContext('2d');
+                if (ctx) {
+                  ctx.drawImage(image, 0, 0, image.width, image.height);
+                }
+              };
```
Juice10, line 150: These should be flipped to insure onload isn't called before the event handler gets added
```suggestion
              image.onload = () => {
                const ctx = (oldElement as HTMLCanvasElement).getContext('2d');
                if (ctx) {
                  ctx.drawImage(image, 0, 0, image.width, image.height);
                }
              };
              image.src = rrCanvasElement.rr_dataURL;
```

Juice10 (changes requested):
Good catch! Found one little tweak to make the code slightly more robust

packages/rrdom/src/diff.ts
Mark-Fenng: Flipped one is more straightforward for sure. But I suddenly realize that according to the JS engine's execution mechanism, I think the onload callback won't be called before the function definition. 
image.onload = ... definition is synchronous while the onload callback is asynchronous so that the callback will be always invoked after the function definition.

Yuyz0112 (approved):

PR #953, opened 2022-08-11
feat: add a destroy function to destroy the whole player

### LINKED ISSUE #883: new rrweb.Replayer 销毁
请问通过 `new rrweb.Replayer` 创建的回放，有办法销毁吗？
页面上可能先后重复创建多个 `new rrweb.Replayer` ，担心会有内存或者重复创建的事件等问题。
目前只是手动清空了 replayer-wrapper DOM 节点。

Mark-Fenng:
This PR adds a destroy function to remove the player from dom. Also, try to fix #883.

Changes in the PR:
- add a destroy function
- add a test case
- add this API to the guidance document
- remove several Eslint warnings

### DIFF
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -433,7 +433,7 @@ export class Replayer {
 
   public resume(timeOffset = 0) {
     console.warn(
-      `The 'resume' was departed in 1.0. Please use 'play' method which has the same interface.`,
+      `The 'resume' was deprecated in 1.0. Please use 'play' method which has the same interface.`,
     );
     this.play(timeOffset);
     this.emitter.emit(ReplayerEvents.Resume);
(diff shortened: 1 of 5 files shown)


packages/rrweb/src/replay/index.ts
```diff
@@ -433,12 +433,22 @@ export class Replayer {
 
   public resume(timeOffset = 0) {
     console.warn(
-      `The 'resume' will be departed in 1.0. Please use 'play' method which has the same interface.`,
+      `The 'resume' was departed in 1.0. Please use 'play' method which has the same interface.`,
```
Juice10, line 436:
```suggestion
      `The 'resume' was deprecated in 1.0. Please use 'play' method which has the same interface.`,
```

Juice10 (approved):

Yuyz0112 (approved):
LGTM

Maybe we can improve this implementation by rejecting further function calls after being destroyed. But at this moment, I think the current implementation is good enough.

PR #956, opened 2022-08-14
fix: can't record shadow root's children except the last one

Mark-Fenng:
PR #834 added the ability to record incremental shadow dom mutations. But I found that it has a bug: If a shadow root has more than one child, the recorder can only record the last child and other children are ignored.

In the test case, I added one more child to the shadow root, but generated snapshot of the old version rrweb didn't change at all, which means the recorder did not record that child. 

The reason is that all children except the last one of a shadow root are all added to the DoubleLinkedList in the `pushAdd` function because their siblings haven't been added yet. But while resolving these nodes in the next stage, their parentNode is the shadow root that is not in the Mirror so they are all ignored to break out the while loop. This PR adds a check for this situation.

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -384,7 +384,7 @@ export default class MutationBuffer {
             }
             // nextId !== -1 && parentId === -1 This branch can happen if the node is the child of shadow root
             else {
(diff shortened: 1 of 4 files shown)


Juice10 (approved):
LGTM!

Yuyz0112 (approved):

Issue #958, opened 2022-08-17
[Bug]: The emit callback is not called when checkoutEveryNms is used and there's no user activity


darind:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Expected Behavior

I've configured rrweb to do a full checkout every 60 seconds by using the **checkoutEveryNms** option:

```
rrweb.record({
    checkoutEveryNms: 60000,
    emit: (event, isCheckout) => {
        console.log(isCheckout);
    }
});
```

I expect that the **emit** callback is called every 60 seconds with the isCheckout parameter set to true and the corresponding event represents a full DOM snapshot.

### Actual Behavior

This works fine if there is some activity on the page (e.g. click, input, mouse move, ...etc). But if there's no activity, the **emit** callback is never called.

Since I need to replay only a sliding portion of the video (e.g. the last 3 minutes from a 10 minutes video), I scheduled invisible DOM changes every 30 seconds (like adding a hidden span) in order to get the correct number of events in the eventsMatrix.

Also I tried using the addCustomEvent and in this case the **emit** callback is raised but the **isCheckout** parameter is always false (see https://github.com/rrweb-io/rrweb/issues/763)

Yuyz0112:
Good catch, currently the checkout feature is lazy, which means only calculates when new events come. So the semantics of checkoutEveryNms does not work like setInterval.

Using a custom event is a good try and I will check why it does not work. (Looks like I missed #763  before...)

Yuyz0112:
I've sent a draft PR and comments are welcomed!

sce9966:
I have same question,how can I resove it

Issue #763, opened 2021-12-07
addCustomEvent never triggers checkout when checkoutEveryNms is used


userback:
I've configured rrweb to do a full checkout every 1 min by using the **checkoutEveryNms** option. It works fine when I move my mouse around on the page. But when I call **addCustomEvent()**, the event gets recorded fine, but the **is_checkout** in the **emit** is always false for the custom events. 

Expected behaviour: **is_checkout** should be true no matter it's a DOM  event or a custom event. 

```
{
  checkoutEveryNms: 60 * 1000,
  emit: function(event, is_checkout) {
      console.log(is_checkout); // this is always false when the event is a custom event
  }
}
```

PR #960, opened 2022-08-17
add custom event into the counting of checkout

Yuyz0112:
related to #958, #763

I've marked this PR as a draft because I think there are several things we can discuss.

## The benefit of counting all events into the checkout calculating

This let the behavior of checkout become more transparent and can use things like custom event and plugin to do fancy features.

But there is also a sub-question of whether we should count all the meta, load, custom, plugin events, or just part of it. For example, the current PR only count custom event because I think the custom events can be controlled by the recorder user.

## The downside of this change

1. It will break some current behavior.
2. If there is a plugin that can produce a lot of events, it may mass up things.

But we are on the schedule for v2.0, so I think it's good timing if we want to do some changes.

### DIFF
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -160,9 +160,12 @@ function record<T = eventWithTime>(
     if (e.type === EventType.FullSnapshot) {
       lastFullSnapshotEvent = e;
       incrementalSnapshotCount = 0;
-    } else if (e.type === EventType.IncrementalSnapshot) {
+    } else if (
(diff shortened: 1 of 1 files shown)


Mark-Fenng:
IMO, I'd prefer to include all events because it's more consistent with the meaning `checkoutEveryNth`. I haven't used this config before so I can't provide suggestions from a user's perspective.
BTW, is it possible for us to implement `checkoutEveryNms ` with setInverval?

Juice10:
I also haven't used this in production so I don't have a proper opinion on the matter, but doing it as part of the 2.0 release makes sense to me

PR #962, opened 2022-08-18
fix: reset fns when when stopping record

wfk007:
reset `wrappedEmit` and `takeFullSnapshot` when stopping record

https://github.com/rrweb-io/rrweb/blob/5f59f9171e481c78d08a1008186cff31a40ee199/packages/rrweb/src/record/index.ts#L528

https://github.com/rrweb-io/rrweb/blob/5f59f9171e481c78d08a1008186cff31a40ee199/packages/rrweb/src/record/index.ts#L547

### DIFF
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -518,6 +518,9 @@ function record<T = eventWithTime>(
     }
     return () => {
       handlers.forEach((h) => h());
+      // reset init fns when stopping record
+      (wrappedEmit as unknown) = undefined;
+      (takeFullSnapshot as unknown) = undefined;
     };
   } catch (error) {
     // TODO: handle internal error


packages/rrweb/src/record/index.ts
```diff
@@ -518,6 +518,9 @@ function record<T = eventWithTime>(
     }
     return () => {
       handlers.forEach((h) => h());
+      // reset init fns when stopping record
+      (wrappedEmit as unknown) = undefined;
+      (takeFullSnapshot as unknown) = undefined;
```
Juice10, line 523: This doesn't reset the fns to their original state, this removes the functions

wfk007: > This doesn't reset the fns to their original state, this removes the functions

These two fns(`wrappedEmit ` and `takeFullSnapshot `) are `undefined` when `record` is not invoked

image

In `record.addCustomEvent`  and `record.takeFullSnapshot`:
Determine whether it is recording by judging whether `wrappedEmit` or `takeFullSnapshot` is `falsy`

image

wfk007: In the below case:

```js
import { record } from "rrweb"

// start record
const stopRecord = record({
  emit: xxx,
})

// stopRecord
stopRecord()

// still invoke 
record.addCustomEvent()
record.takeFullSnapshot()
```

Juice10: Good catch @wfk007, thanks for finding this and clarifying!

Juice10 (approved):

Yuyz0112:
Thanks!

Vadman97:
Hi @wfk007 ! After this change, we've noticed that when calling the stop recording callback, observers are not immediately stopped throw the following error: 
image

https://github.com/rrweb-io/rrweb/blob/2286c11912652e9158eefa35a42b4b608b7bfb29/packages/rrweb/src/record/index.ts#L561-L566

Could the `mousemoveCb` call from the mouse / viewport observers be trying to call `wrappedEmit` which is set `undefined` by this PR?
https://github.com/rrweb-io/rrweb/blob/2286c11912652e9158eefa35a42b4b608b7bfb29/packages/rrweb/src/record/observer.ts#L157

wfk007:
> Hi @wfk007 ! After this change, we've noticed that when calling the stop recording callback, observers are not immediately stopped throw the following error: image
> 
> https://github.com/rrweb-io/rrweb/blob/2286c11912652e9158eefa35a42b4b608b7bfb29/packages/rrweb/src/record/index.ts#L561-L566
> 
> Could the `mousemoveCb` call from the mouse / viewport observers be trying to call `wrappedEmit` which is set `undefined` by this PR?
> 
> https://github.com/rrweb-io/rrweb/blob/2286c11912652e9158eefa35a42b4b608b7bfb29/packages/rrweb/src/record/observer.ts#L157

@Vadman97 thx for your reply. I encounterred this error too.
It is because that the `MoveObserver` uses **double throttle** to track mouseMove events
The first one is **updatePosition** with options **trailing false**
the second one is **wrappedCb** **WITHOUT** options **trailing false**
which means that the second throttle will be triggerred async
when stopping record, wrappedEmit will be reset to undefined, but wrappedCb will still execute because of trailing: true

Mark-Fenng:
@wfk007 I encountered this error as well. Do you have any idea or plan to fix this?

wfk007:
> @wfk007 I encountered this error as well. Do you have any idea or plan to fix this?

@Mark-Fenng @Yuyz0112 @Juice10 @Vadman97 I opened a PR to fix this error: https://github.com/rrweb-io/rrweb/pull/1034

PR #966, opened 2022-08-19
fix: record canvas by fps when blockClass is RegExp

QxQstar:
解决当 blockClass 是 RegExp 时，用 FPSObserver 不能录制 canvas 的 bug

### DIFF
--- a/packages/rrweb/src/record/observers/canvas/canvas-manager.ts
+++ b/packages/rrweb/src/record/observers/canvas/canvas-manager.ts
@@ -141,7 +141,7 @@ export class CanvasManager {
     const getCanvas = (): HTMLCanvasElement[] => {
       const matchedCanvas: HTMLCanvasElement[] = [];
       win.document.querySelectorAll('canvas').forEach(canvas => {
-        if (!isBlocked(canvas, blockClass, false)) {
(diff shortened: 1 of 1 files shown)


Juice10 (dismissed):

packages/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -137,6 +138,16 @@ export class CanvasManager {
     let lastSnapshotTime = 0;
     let rafId: number;
 
+    const getCanvas = (): HTMLCanvasElement[] => {
+      const matchedCanvas: HTMLCanvasElement[] = [];
+      win.document.querySelectorAll('canvas').forEach(canvas => {
+        if (!isBlocked(canvas, blockClass, false)) {
```
Juice10, line 144: In this case you need to check if the ancestors are blocked too
```suggestion
        if (!isBlocked(canvas, blockClass, true)) {
```

Juice10 (commented):
Thanks for submitting this PR!

Juice10 (approved):

Yuyz0112 (approved):

PR #970, opened 2022-08-22
bugfix: Sort attributes to make `rr_*` attributes handled last

Juice10:
`rr_dataURL` overwrites `src` attribute, because of this we need to evaluate `rr_dataURL` last so `src` doesn't accidentally get overwritten again.


packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -116,6 +117,31 @@ export function createCache(): BuildCache {
   };
 }
 
+/**
+ * `rr_` attributes are magic, they change some of the other attributes on the elements,
+ * so we need to parse them last so they can overwrite any conflicting attributes.
+ *
+ * @param attributes list of html attributes to be added to the element
```
Juice10, line 124:
```suggestion
 * @param attributes - list of html attributes to be added to the element
```

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -116,6 +117,31 @@
...
 }
 
+/**
+ * `rr_` attributes are magic, they change some of the other attributes on the elements,
+ * so we need to parse them last so they can overwrite any conflicting attributes.
+ *
+ * @param attributes - list of html attributes to be added to the element
+ * @returns attributes with rr_* attributes last in the array
+ */
+function sortAttributes(attributes: attributes): attributes {
```
Yuyz0112, line 127: This code looks clean.

But since this is a hot data path, shall we reduce the times of loop in order to keep the performance?

Juice10:
Had a chat with @eoghanmurray and @Mark-Fenng, I'm going to do a refactor which eliminates the loop overhead.

Juice10:
> I think you've got the alternative version without the `specialAttributes` array, but with the boolean flag?

@eoghanmurray I did some performance tests and this one was faster so I ended up going with this

eoghanmurray:
Cool, happy to approve!
This way is also elegant now that I see it.
I thought the `specialAttributes` was from before Friday when we had a chat about it.

Juice10:
@eoghanmurray thanks Eoghan! Yes it was from our chat initially. But it got deleted and then came back again

Mark-Fenng (approved):
Great! A clean and performant version.

Yuyz0112 (approved):

Issue #983, opened 2022-09-01
[Bug]: Iframe  replay fails after the second full snapshot


luocongqiu:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb-snapshot

### Version

v1.1.14

### Expected Behavior

replay success for every full snapshot

### Actual Behavior

take a full snapshot after every 10000 ms
` const events = [];
        rrweb.record({
            emit(event) {
                events.push(event);
            },
            checkoutEveryNms: 10 * 1000
        });`

the first full snapshot replay success，after second full snapshot，iframe not show

[ifame.txt](https://github.com/rrweb-io/rrweb/files/9468582/ifame.txt)

### Steps to Reproduce

use the test code in rrweb/packages/rrweb/test/html

have iframe node

pass checkoutEveryNms param

luocongqiu:
I pull a request [pull request](https://github.com/rrweb-io/rrweb/pull/984)

PR #984, opened 2022-09-01
fix: Iframe replay fails after the second full snapshot #983

luocongqiu:
the frame document node root id is undefined after first full snapshot, is the same as id after second full snapshot, then replay fail


Mark-Fenng (approved):
Thanks for your great pull request. I also ran into this problem before. I tried to fixed it on the replayer side. https://github.com/rrweb-io/rrweb/pull/989/files#diff-0d9a03e75eacff8f6be89336b49419b210455c1a81d8f46fae7cea2d83a8f93eR339-R341
It can resolve the error for events recorded with the old version.
Your fix is on the recorder side and it's also very nice.

Yuyz0112:
Thanks!

Issue #669, opened 2021-08-16
Cannot record web Component css


GoncaloDuarte98:
I cannot record the css inside a lit Element web Component. I can see the component but no style which breaks the recording

Yuyz0112:
@GoncaloDuarte98 Could you provide an example code for debugging?

GoncaloDuarte98:
This is the web component:
![image](https://user-images.githubusercontent.com/47533890/129536381-1b57d62a-4c20-4364-bb44-fd1ba44d70a1.png)
And this is the recording using rrweb:
![image](https://user-images.githubusercontent.com/47533890/129536289-23e247ba-228f-493f-9c11-96e0b42cbd73.png)

GoncaloDuarte98:
` [
    {
        "type": 0,
        "data": {},
        "timestamp": 1629103007609,
        "delay": -5.300048828125
    },
    {
        "type": 1,
        "data": {},
        "timestamp": 1629103007637,
        "delay": 22.699951171875
    },
    {
        "type": 4,
        "data": {
            "href": "http://localhost:8000/",
            "width": 1041,
            "height": 889
        },
        "timestamp": 1629103007638,
        "delay": 23.699951171875
    },
    {
        "type": 2,
        "data": {
            "node": {
                "type": 0,
                "childNodes": [
                    {
                        "type": 1,
                        "name": "html",
                        "publicId": "",
                        "systemId": "",
                        "id": 2
                    },
                    {
                        "type": 2,
                        "tagName": "html",
                        "attributes": {
                            "lang": "en-GB"
                        },
                        "childNodes": [
                            {
                                "type": 2,
                                "tagName": "head",
                                "attributes": {},
                                "childNodes": [
                                    {
                                        "type": 3,
                                        "textContent": "\n    ",
                                        "id": 5
                                    },
                                    {
                                        "type": 2,
                                        "tagName": "meta",
                                        "attributes": {
                                            "charset": "utf-8"
                                        },
                                        "childNodes": [],
                                        "id": 6
                                    },
                                    {
                                        "type": 3,
                                        "textContent": "\n    ",
                                        "id": 7
                                    },
                                    {
                                        "type": 2,
                                        "tagName": "style",
                                        "attributes": {},
                                        "childNodes": [
                                            {
                                                "type": 3,
                                                "textContent": "\n      body {\n        background: #fafafa;\n      }\n    ",
                                                "isStyle": true,
                                                "id": 9
                                            }
                                        ],
[...]

PR #989, opened 2022-09-07
feat: add support for recording and replaying adoptedStyleSheets API

Mark-Fenng:
## Background

The **adoptedStyleSheets** property of the Document or shadowRoot is used for setting an array of constructed stylesheets to be used by the host. reference: [https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets](https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets)

Nowadays, more and more web libraries and pages are using custom web components and this API like [Lit](https://lit.dev/), [material wc](https://material-components.github.io/material-web/demos/button/). There’s also a related issue #669.

## Implementation

This PR mostly follows @Yuyz0112's  proposal: [https://github.com/rrweb-io/rrweb/issues/669#issuecomment-937522437](https://github.com/rrweb-io/rrweb/issues/669#issuecomment-937522437). 

I create a StyleMirror which is similar to Node Mirror that we used to manage nodes and their serialized ids. In the recording side, if I get a constructed stylesheet from adoptedStyleSheets API, firstly ask StyleMirror whether it appears before or not. If it’s a new stylesheet, I add it to the StyleMirror, assign a new unique styleId, and serialize it into an event. 

Here’s an example event to serialize a new stylesheet object:

```tsx
{
  type: EventType.IncrementalSnapshot,
  data: {
    source: IncrementalSource.AdoptedStyleSheet,
    id: 1, // id indicates the node id of document or shadow DOMs' host element.
    // serialize the new stylesheet
    styles: [
      {
        rules: [
          {
            rule: 'div { color: yellow; }',
          },
        ],
        styleId: 1,
      },
    ],
    styleIds: [1], // style ids of adopted stylesheets
  }
}
```

If it already has a styleId, I will reference (reuse) it directly like this:

```tsx
{
  type: EventType.IncrementalSnapshot,
  data: {
    source: IncrementalSource.AdoptedStyleSheet,
    id: 1, // id indicates the node id of document or shadow DOMs' host element.
    styleIds: [1], // reference styleIds directly
  }
}
```

On the replaying side, I deserialize stylesheet objects from events and assign them to the target documents.

## Changes

1. add a new incremental event type: AdoptedStyleSheet.
2. change the current styleSheetRule and styleDeclaration event and make them support tracking mutation of adoptedStyleSheets.
3. totally remove virtualStylesheet from rrdom to reduce duplicate code.
4. add support for [replace](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/replace) and [replaceSync](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/replaceSync) APIs.
5. refactor stylesheet-manager and move the main code of adopedStylesheets into it.
6. 5 test cases (3 on the recording side and 2 on the replaying side).
7. StyleMirror implementation and unit tests.
8. refactor the stylesheet part in the replayer and add support for VirtualDom optimization.
9. patch adoptedStyleSheet API to track its modification after the full snapshot is created.

## Other things

Because adoptedStypeSheets is a new API, there is a browser compatibility issue for the replaying side. The replayer can replay adoptedStyleSheet events only when the browser supports the API. reference: [https://caniuse.com/?search=adoptedStyleSheets](https://caniuse.com/?search=adoptedStyleSheets)

Special thanks to @lele0108 's sponsorship for making this pull request possible.

### Before this PR:

https://user-images.githubusercontent.com/27533910/188832637-cf87e5e8-178d-4e02-a4fd-6991df04bec2.mp4

### After this PR applied:

https://user-images.githubusercontent.com/27533910/188832739-0ce45aac-a079-4a8a-9b1d-72412eb2e31f.mp4


Mark-Fenng:
@dbseel  This PR is relatively huge and it may take some time for Yuyz and Justin to review. I believe that it will get merged finally.

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -336,11 +336,9 @@ export function buildNodeWithSN(
...
     return null;
   }
-  if (n.rootId) {
-    console.assert(
-      (mirror.getNode(n.rootId) as Document) === doc,
-      'Target document should have the same root id.',
-    );
+  // If the snapshot is created by checkout, the rootId doesn't change but the iframe's document can be changed automatically when a new iframe element is created.
+  if (n.rootId && (mirror.getNode(n.rootId) as Document) !== doc) {
+    mirror.replace(n.rootId, doc);
```
Yuyz0112, line 341: I think this is still need because legacy events may have this issue.

packages/rrweb/src/record/observer.ts
```diff
@@ -497,8 +498,30 @@ function getNestedCSSRulePositions(rule: CSSRule): number[] {
   return recurse(rule, positions);
 }
 
+/**
+ * For StyleSheets in Element, this function retrieves id of its host element.
+ * For adopted StyleSheets, this function retrieves its styleId from a styleMirror.
+ */
+function getIdOrStyleId(
```
Yuyz0112, line 505: `getIdOrStyleId` -> `getIdAndStyleId`?

Yuyz0112 (commented):
Very exciting PR 👍 

Some minor tweaks and the defer solution may need to be verified in the real world. @Mark-Fenng feel free to merge this when you change the requests.

Mark-Fenng:
Thanks for the suggestions. There are many changes related to @Juice10's code base. I'll merge it after his review.

packages/rrweb/src/record/constructableStyleSheets.d.ts
```diff
@@ -0,0 +1,10 @@
+// This informs the TS compiler about constructed stylesheets.
+// It can be removed when this is fixed: https://github.com/Microsoft/TypeScript/issues/30022
+declare interface DocumentOrShadowRoot {
+  adoptedStyleSheets: CSSStyleSheet[];
```
Juice10, line 4: Since this is not always available (no current or pre-release version of safari supports it for example) it might make sense to add it as an 'optional' parameter
```suggestion
  adoptedStyleSheets?: CSSStyleSheet[];
```

packages/rrweb/src/record/constructableStyleSheets.d.ts
```diff
@@ -0,0 +1,10 @@
...
+}
+
+declare interface CSSStyleSheet {
+  replace(text: string): Promise<CSSStyleSheet>;
+  replaceSync(text: string): void;
+}
```
Juice10, line 10: Same for these
```suggestion
declare interface CSSStyleSheet {
  replace(text: string)?: Promise<CSSStyleSheet>;
  replaceSync(text: string)?: void;
}
```

packages/rrweb/src/record/constructableStyleSheets.d.ts
```diff
@@ -0,0 +1,10 @@
+// This informs the TS compiler about constructed stylesheets.
```
Juice10, line 1: So far we've mostly used kebab-case for file names. This file is deviating from that. Probably makes sense to rename it to `constructible-style-sheets.d.ts`

packages/rrweb/src/replay/index.ts
```diff
@@ -1817,6 +1751,162 @@ export class Replayer {
...
+       * The replayer has to get the correct host window to recreate a StyleSheetObject.
+       */
+      let hostWindow: IWindow | null = null;
+      if (hasShadowRoot(targetHost))
+        hostWindow = targetHost.ownerDocument?.defaultView || null;
+      else if (targetHost.nodeName === '#document')
+        hostWindow = (targetHost as Document).defaultView;
+
+      if (!hostWindow) return;
+      newStyleSheet = new hostWindow.CSSStyleSheet();
```
Juice10, line 1871: This will break Safari playback without a polyfill

Juice10 (changes requested):
What a great addition to rrweb, thanks for that @Mark-Fenng!
I am however concerned about Safari and breaking playback in that browser, we currently aren't using any polyfills to add support for it in the playback. And on record we could also be breaking things there since we're setting properties which would otherwise be empty, preventing people from doing feature detection.
I made this mistake myself once in the past when building the first iteration of the stylesheet manager and I really wanted to add https://github.com/amilajack/eslint-plugin-compat to the project but as we where using tslint back then instead of eslint I wasn't able to. Maybe this is a good moment to add it since we are using eslint at the moment.

packages/rrweb/src/record/observer.ts
```diff
@@ -620,15 +716,85 @@ function initStyleSheetObserver(
...
+      if (hostId !== null && hostId !== -1) {
+        try {
+          stylesheetManager.adoptStyleSheets(sheets, hostId);
+        } catch (e) {
+          // for safety
+        }
+      }
+      return result;
+    },
+  });
```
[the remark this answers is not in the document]
Mark-Fenng: I think https://github.com/rrweb-io/rrweb/pull/989/files/ca9da9a9e2fe2190768db364bf5f6a3a53673bf9#diff-b7a42c375ac5ab2de196b5e95e0b2f82b723969c5836fac0ff6f7ca6ad66ebd7R753 this line can check the existence of the API.

Mark-Fenng:
@Juice10 Geat catches on compatibility issues of these APIs. I also added the eslint-plugin-compat and configured the browser list for each package. 
If some APIs are rarely used in the package, and they are polyfilled or have a workaround in older browsers, I think this lint rule can be disabled. Otherwise, the browser list has to be updated.

packages/rrweb/src/record/index.ts
```diff
@@ -331,21 +348,32 @@ function record<T = eventWithTime>(
               window.pageXOffset !== undefined
                 ? window.pageXOffset
                 : document?.documentElement.scrollLeft ||
+                  // eslint-disable-next-line compat/compat
```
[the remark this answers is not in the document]
Juice10: This PR fixes all these issues: https://github.com/rrweb-io/rrweb/pull/1010

packages/rrweb/src/record/index.ts
```diff
@@ -331,21 +348,32 @@ function record<T = eventWithTime>(
...
                   document?.body?.parentElement?.scrollLeft ||
-                  document?.body.scrollLeft ||
+                  // eslint-disable-next-line compat/compat
```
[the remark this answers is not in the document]
Juice10: Fixed by: https://github.com/rrweb-io/rrweb/pull/1010

packages/rrweb/src/record/index.ts
```diff
@@ -331,21 +348,32 @@ function record<T = eventWithTime>(
...
+                  document?.body?.scrollLeft ||
                   0,
             top:
               window.pageYOffset !== undefined
                 ? window.pageYOffset
                 : document?.documentElement.scrollTop ||
+                  // eslint-disable-next-line compat/compat
```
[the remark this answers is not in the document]
Juice10: Fixed by: https://github.com/rrweb-io/rrweb/pull/1010

Juice10 (approved):
Great work @Mark-Fenng! Thanks for the changes, I think this is a great addition to rrweb

Issue #993, opened 2022-09-08
[Bug]: element in add mutation LOST when using slimDOMOptions


wfk007:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

rrweb@2.0.0-alpha.1

### Expected Behavior

div.ks-scrollbar not lost

![image](https://user-images.githubusercontent.com/22289445/189115128-1670c653-ffba-4b2c-9746-c56640d8ca82.png)

### Actual Behavior

div.ks-scrollbar lost

![image](https://user-images.githubusercontent.com/22289445/189114869-978dd953-fec4-4204-8e40-a4ec3d44f9a1.png)

### Steps to Reproduce

1. install rrweb@2.0.0-alpha.1
2. set `slimDOMOptions: true`
3. click Select to trigger drop down（you can use Select component in Element-UI：https://element.eleme.io/#/en-US/component/select）
the `div.ks-scrollbar ` needs to be followed by a `comment`

![image](https://user-images.githubusercontent.com/22289445/189116167-5257eaad-3631-4c0d-8018-fd3b3eee9094.png)

PR #994, opened 2022-09-08
fix: add mutation lost in slimDOMOptions

wfk007:
fix: https://github.com/rrweb-io/rrweb/issues/993

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -1014,13 +1014,14 @@ export function serializeNodeWithId(
   } else {
     id = genId();
   }
+
+  const serializedNode = Object.assign(_serializedNode, { id });
+  // add IGNORED_NODE to mirror to track nextSiblings
+  mirror.add(n, serializedNode);
+
   if (id === IGNORED_NODE) {
     return null; // slimDOM
   }
-
-  const serializedNode = Object.assign(_serializedNode, { id });
-
-  mirror.add(n, serializedNode);
 
   if (onSerialize) {
     onSerialize(n);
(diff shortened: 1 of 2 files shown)


wfk007:
in rrweb@2.0.0-alpha.1, mirror.getId() return -1 may caused by:

1. mutation add: child node may be pushed before its newly added parent, rrweb uses a DoubleLinkedList to store these nodes
2. slimDOMOptions: true, and the Node is an `IGNORED_NODE`, **but not in Mirror**

considering case 2:
in rrweb@1.1.3, we can use `node.__sn` to check whether nextSibling is an `IGNORED_NODE`
**BUT** in in rrweb@2.0.0-alpha.1, we can not
because the information with `IGNORED_NODE` neither in `node.__sn` nor in `Mirror`

wfk007:
before:

https://user-images.githubusercontent.com/22289445/189289797-2b36b3d1-fd54-4139-80d2-a80592945564.mov

after fixed:

https://user-images.githubusercontent.com/22289445/189290282-d14ae704-1ff8-47ed-aa8b-4178ea5654d0.mov

Mark-Fenng (approved):
Thanks for your very nice pull request. The description and solution are very clear and clean.

The failure of Format Check CI can be ignored because it may be caused by the old branch base. I've checked the snapshot.ts and there's no formatting issue.

wfk007:
@Mark-Fenng @Yuyz0112
Hi, is this pending any other changes?

Mark-Fenng:
Nope, I think everything is good. Just waiting for @Yuyz0112's review.

Yuyz0112:
sorry for the late

PR #909, opened 2022-05-31
Inline stylesheets on load

Juice10:
Fixes #764

PR #995, opened 2022-09-11
Fix inline link elements bug

### LINKED ISSUE #981: [Bug]: Style missing in microapp(used customElements)
### Preflight Checklist

rrweb

### Version

v2.0.0-alpha.0

### Expected Behavior

Ability to record microapp style changes( It is actually a custom element within which a link tag is appended )

### Actual Behavior

The microapp async route style was not successfully recorded

### Steps to Reproduce

demo code: https://github.com/hb-bobo/rrweb-microapp

```bash
git clone https://github.com/hb-bobo/rrweb-microapp.git
yarn
yarn dev
```

---------------

```html
<qiankun-head> 

<qiankun-head>

const link = document.createElement('link');
link.rel = 'stylesheet'
link.type = "text/css"
link.href = '//activity.hdslb.com/blackboard/static/1d5c9c63ddd602617ed9a6ec73bebe8f/test.css'
document.querySelector('qiankun-head').appendChild(link);
```

### Testcase Gist URL

_No response_

### Additional Information

_No response_

Mark-Fenng:
In PR #909, an async mechanism for inlining link elements is introduced but it still has a bug and some negative impacts on the replayer side.

### The Bug
In the stylesheet-manager.ts, when the link element loaded the styles, the manager emits an incremental mutation event to inline styles. But the mutation's parentId is wrong. https://github.com/rrweb-io/rrweb/blob/abc035fd00972c3ffa1a9cf379b46f53cb74d394/packages/rrweb/src/record/stylesheet-manager.ts#L34
As a result, the event looks like this: 
image https://github.com/rrweb-io/rrweb/compare/master...fix-inline-link-elements#diff-988864050cbe6d85b4cb140a06ffce65819d8967e071b175f1d27b8a864b5383L176-L189
The parentNode of the link element is itself and it can't be appended to the dom actually.

### Negative Impacts
This mechanism makes elements absent in the normal snapshot stage and causes performance issues and many warnings in the console.
Here's a diagram to explain this problem.

explanation

image

### Changes

I built this PR based on #989 so that it contains all changes of that PR. **This PR has to get merged after #989**.
I made some changes to the current mechanism. In the snapshot stage, the unloaded Link elements will still get serialized without inline styles. After they are loaded. an incremental mutation of attributes is appended which includes an attribute `_cssText`. Its value is the inlined CSS styles. 
```ts
{
  source: IncrementalSource.Mutation,
  adds: [],
  removes: [],
  texts: [],
  attributes: [
    {
      id: 9,
      attributes: {
        _cssText: "body { color: pink; }",
      },
    },
  ],
}
```
This way, the unloaded Link elements can be inlined without causing the above-mentioned side effects.
On the replaying side, when the replayer encounters the '_cssText' attribute change and its target is the Link element, the replayer can replace the old un-inlined one with the inlined loaded Link element.
This method can make it easy to implement the TODO feature as well. https://github.com/rrweb-io/rrweb/blob/abc035fd00972c3ffa1a9cf379b46f53cb74d394/packages/rrweb/src/record/stylesheet-manager.ts#L19-L24 

This PR can also fix #981.

### DIFF
--- a/packages/rrdom/src/index.ts
+++ b/packages/rrdom/src/index.ts
@@ -389,7 +389,12 @@ export class Mirror implements IMirror<RRNode> {
   }
 
   replace(id: number, n: RRNode) {
-    this.idNodeMap.set(id, n);
+    const oldNode = this.getNode(id);
+    if (oldNode) {
+      const meta = this.nodeMetaMap.get(oldNode);
+      if (meta) this.nodeMetaMap.set(n, meta);
+      this.idNodeMap.set(id, n);
+    }
   }
 
   reset() {
--- a/packages/rrdom/src/index.ts
+++ b/packages/rrdom/src/index.ts
@@ -393,8 +393,8 @@ export class Mirror implements IMirror<RRNode> {
     if (oldNode) {
       const meta = this.nodeMetaMap.get(oldNode);
       if (meta) this.nodeMetaMap.set(n, meta);
-      this.idNodeMap.set(id, n);
     }
+    this.idNodeMap.set(id, n);
   }
 
   reset() {
--- a/packages/rrweb/src/record/stylesheet-manager.ts
+++ b/packages/rrweb/src/record/stylesheet-manager.ts
@@ -1,8 +1,9 @@
-import type { Mirror, serializedNodeWithId } from 'rrweb-snapshot';
+import type { elementNode, serializedNodeWithId } from 'rrweb-snapshot';
 import { getCssRuleString } from 'rrweb-snapshot';
 import type {
   adoptedStyleSheetCallback,
   adoptedStyleSheetParam,
+  attributeMutation,
   mutationCallBack,
 } from '../types';
 import { StyleSheetMirror } from '../utils';
@@ -24,20 +25,20 @@ export class StylesheetManager {
   public attachLinkElement(
     linkEl: HTMLLinkElement,
     childSn: serializedNodeWithId,
-    mirror: Mirror,
   ) {
-    this.mutationCb({
-      adds: [
-        {
-          parentId: mirror.getId(linkEl),
-          nextId: null,
-          node: childSn,
-        },
-      ],
-      removes: [],
-      texts: [],
-      attributes: [],
-    });
+    if ('_cssText' in (childSn as elementNode).attributes)
+      this.mutationCb({
+        adds: [],
+        removes: [],
+        texts: [],
+        attributes: [
+          {
+            id: childSn.id,
+            attributes: (childSn as elementNode)
+              .attributes as attributeMutation['attributes'],
+          },
+        ],
+      });
 
     this.trackLinkElement(linkEl);
   }
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -6,6 +6,8 @@ import {
   createCache,
   Mirror,
   createMirror,
+  attributes,
+  serializedElementNodeWithId,
 } from 'rrweb-snapshot';
 import {
   RRDocument,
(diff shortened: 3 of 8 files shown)


packages/rrdom/src/index.ts
```diff
@@ -389,7 +389,12 @@ export class Mirror implements IMirror<RRNode> {
   }
 
   replace(id: number, n: RRNode) {
-    this.idNodeMap.set(id, n);
+    const oldNode = this.getNode(id);
+    if (oldNode) {
+      const meta = this.nodeMetaMap.get(oldNode);
+      if (meta) this.nodeMetaMap.set(n, meta);
+      this.idNodeMap.set(id, n);
+    }
```
Juice10, line 397: If replace gets called but there is nothing to replace, the old behaviour added that node to the idNodeMap, the new behaviour does not. I'm not sure if this is good or bad, maybe it should produce a warning if there is no node available in case we messed up somewhere.

packages/rrweb/src/record/stylesheet-manager.ts
```diff
@@ -25,20 +25,20 @@ export class StylesheetManager {
...
-          parentId: mirror.getId(linkEl),
-          nextId: null,
-          node: childSn,
-        },
-      ],
-      removes: [],
-      texts: [],
-      attributes: [],
-    });
+    if ('_cssText' in (childSn as elementNode).attributes)
```
Juice10, line 29: Is `_cssText` always going to be set, also if a link element hasn't finished loading yet?

packages/rrweb/src/replay/index.ts
```diff
@@ -1637,6 +1639,44 @@ export class Replayer {
...
+                  const parentNode = target.parentNode;
+                  if (newNode && parentNode) {
+                    parentNode.removeChild(target as Node & RRNode);
+                    parentNode.insertBefore(
+                      newNode as Node & RRNode,
+                      siblingNode as (Node & RRNode) | null,
+                    );
+                    mirror.replace(mutation.id, newNode as Node & RRNode);
+                    break;
+                  }
```
Juice10, line 1669: In this case we are replacing an existing Link or probably Style element with a new Style element with the updated css. However we don't seem to be keeping around the existing attributes on that element. That means that if someone set `media="print"` on their stylesheet, that will essentially get wiped off and will effectively work as if it was  `media="all"`. This could break some websites

packages/rrdom/src/index.ts
Juice10:
```suggestion
    const oldNode = this.getNode(id);
    if (oldNode) {
      const meta = this.nodeMetaMap.get(oldNode);
      if (meta) this.nodeMetaMap.set(n, meta);
      this.idNodeMap.set(id, n);
    } else {
      console.warn(`Mirror tried to replace node with id: ${id}, but no node exists with that id`);
    }
```

Juice10 (changes requested):
Thanks for finding and fixing this bug! I found a one bug and maybe another potential bug. Could you fix those and I'd be happy to re-review & approve!

packages/rrweb/src/record/stylesheet-manager.ts
Mark-Fenng: According the code below https://github.com/rrweb-io/rrweb/blob/08ab7ab45227a31f3f35e2ff43e6517f567130b2/packages/rrweb-snapshot/src/snapshot.ts#L653-L659
I think it will set _cssText when the link has been loaded. Otherwise, recorder will only keep its original attributes.

packages/rrweb/src/replay/index.ts
Mark-Fenng: Good catch, I add the code to keep their existing attributes before rebuilding the elements.

Mark-Fenng:
Today I encountered an issue with the replayer also caused by this.

This is the link element on the recorder side.
image

While this screenshot is the link element on the replayer side.
image
The element is self-nested in 4 depths.

Mark-Fenng:
I rebased the latest master branch and this pull request is ready to get reviewed and merged.

Yuyz0112 (approved):

PR #1002, opened 2022-09-16
fix duplicate shadow doms in the recorder

Mark-Fenng:
### bug description

In the recorder, some elements can be mutated and serialized multiple times. If they have attached shadow doms, these shadow doms can be also observed multiple times in the `shadow-dom-manager`. Then children of shadow doms would be created multiple times causing duplicate elements in the replayer.
For example:
image
The original page is on the left-hand side and the problematic page is on the RHS.

I also noticed a bug in the live-stream script. When navigating a page, the recorder would be executed twice and also cause duplicate shadow doms.

### Solution

Add a set in the `shadow-dom-manager` and check each shadow root before observing them.

### DIFF
--- a/packages/rrweb/scripts/stream.js
+++ b/packages/rrweb/scripts/stream.js
@@ -227,6 +227,9 @@ void (async () => {
         'window.__IS_RECORDING__',
       );
       if (!isRecording) {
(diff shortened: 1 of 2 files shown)


Juice10 (approved):

Yuyz0112 (approved):
LGTM!

PR #1007, opened 2022-09-20
Allow player to play a from-to range

Juice10:
And optionally trigger a callback when its done or loop continously

### DIFF
--- a/packages/rrweb-player/src/Controller.svelte
+++ b/packages/rrweb-player/src/Controller.svelte
@@ -39,6 +39,13 @@
   let step: HTMLElement;
   let finished: boolean;
 
+  let pauseAt: number | false = false;
+  let onPauseHook: () => unknown | undefined = undefined;
+  let loop: {
+    start: number;
+    end: number;
+  } | null = null;
+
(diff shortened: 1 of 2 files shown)


Mark-Fenng (commented):
Looks good to me except for the minor issue

packages/rrweb-player/src/Controller.svelte
```diff
@@ -152,6 +173,27 @@
...
+     startLooping: boolean = false,
+     afterHook: undefined | (() => void) = undefined,
+   ) => {
+    if (startLooping) {
+      loop = {
+        start: timeOffset,
+        end: endTimeOffset,
+      };
+    } else {
+      loop = undefined;
```
Mark-Fenng, line 188: Just a minor problem:
The loop's definition is 
```ts
{
    start: number;
    end: number;
} | null
```
So I think we should also assign null to 'loop' here.

Juice10, line 188:
```suggestion
      loop = null;
```

Yuyz0112 (approved):

PR #1008, opened 2022-09-22
Live mode - addActions didn't work with liveMode

eoghanmurray:
`addActions` (not `addAction`) had a bug where it would replace `this.actions` rather than mutating it, causing the `actions` array in the `raf` loop to not get new actions (because it still referred to the previous incarnation of `this.actions`)


eoghanmurray:
Sorry this is a dupe of #1001

PR #1011, opened 2022-09-27
Feat: pruneBranches isolates part of the dom tree, and removes all other parts in events

Juice10:
This is super useful for debugging huge events.

Syntax:
```js
const ids = [150]; // list of ids who's parents and children we'd like to keep
const prunedEvents = pruneBranches(events, {keep: ids});
```


packages/rrweb-cutter/src/index.ts
```diff
@@ -96,6 +98,132 @@ function cutEvents(
...
+            ...event,
+            data: {
+              ...event.data,
+              positions: prunedPositions,
+            },
+          } as eventWithTime);
+      } else if ('id' in event.data) {
+        if (treeSet.has(event.data.id)) result.push(event);
+      } else if (event.data.source === IncrementalSource.Mutation) {
+        const { removes, adds, texts, attributes } = event.data;
```
Mark-Fenng, line 151: For the adds mutation, I come up with one corner situation: if one node is moved to be a child of another pruned node, will this cause errors during the playback?

Mark-Fenng (commented):
Love this feature! It will really help us with debugging. The implementation looks good and I only have one question about a corner case I'm not so sure.

packages/rrweb-cutter/src/index.ts
Juice10: Hmm interesting (edge) case. Hadn't thought of that one yet

Juice10: Done: https://github.com/rrweb-io/rrweb/pull/1011/commits/57f6ad6fe8f1a8886490cc19fee04577ef01f555

Juice10: Thanks for the suggestion @Mark-Fenng

Juice10:
@Mark-Fenng I think this one is ready to be merged into your `trim` branch. You cool with me doing that?

Mark-Fenng (approved):

PR #1012, opened 2022-09-28
Bug: Call `afterAppend` for mutations and top level elements

Juice10:
The webrtc plugin couldn't call onBuild for canvas elements added as a mutation since afterAppend wasn't always being called. This PR fixes that issue.


Yuyz0112 (approved):
good catch

PR #1013, opened 2022-10-03
`prototype` here does not refer to the correct WebGLRenderingContext …

eoghanmurray:
…when there are multiple being patched - all were referring to the same one meaning that variables weren't being indexed correctly as `canvasVarMap` in serialize-args.ts was only seeing a single context

If you revert the change from `prototype` to `this` from this commit, you'll get the following failure in the new test added in this commit:

   @@ -199,11 +199,11 @@
                "property": "bindBuffer",
                "args": [
                  34962,
                  {
                    "rr_type": "WebGLBuffer",
    -               "index": 0
    +               "index": 1
                  }
                ]
              },

This is because the 'confound' canvas was populating a single entry in the canvasVarMap so when 'myCanvas' starts populating, there is already an entry (which doesn't exist on the replay side as the 'confound' canvas was never emitted)

### DIFF
--- a/packages/rrweb/test/html/canvas-webgl-shader.html
+++ b/packages/rrweb/test/html/canvas-webgl-shader.html
@@ -48,8 +48,11 @@
         gl.drawArrays(gl.TRIANGLES, 0, 3);
       }
       setTimeout(() => {
+        // canvas element that never gets appended to the DOM
         const confound = document.createElement('canvas');
-        createTriangle(confound);  // this should not be emitted (but was triggering a bug in the emitted canvas)
+        // These will not be emitted (but was triggering a bug in the emitted canvas)
+        // @see https://github.com/rrweb-io/rrweb/pull/1013
+        createTriangle(confound);
         const c = document.getElementById('myCanvas');
         createTriangle(c);
       }, 10);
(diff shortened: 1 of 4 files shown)


packages/rrweb/test/html/canvas-webgl-shader.html
```diff
@@ -0,0 +1,58 @@
...
+        gl.enableVertexAttribArray(coord);
+        gl.clearColor(0.5, 0.5, 0.5, 0.9);
+        gl.enable(gl.DEPTH_TEST);
+        gl.clear(gl.COLOR_BUFFER_BIT);
+        gl.viewport(0,0,canvas.width,canvas.height);
+        gl.drawArrays(gl.TRIANGLES, 0, 3);
+      }
+      setTimeout(() => {
+        const confound = document.createElement('canvas');
+        createTriangle(confound);  // this should not be emitted (but was triggering a bug in the emitted canvas)
```
Juice10, line 52: I was confounded myself a little for a second so I though maybe some extra comments could help future me, whenever I have to re-read it
```suggestion
        // canvas element that never gets appended to the DOM
        const confound = document.createElement('canvas');
        // These will not be emitted (but was triggering a bug in the emitted canvas)
        // @see https://github.com/rrweb-io/rrweb/pull/1013
        createTriangle(confound);
```

Juice10 (approved):
Really great catch @eoghanmurray! Thanks for incorporating some nice tests so this won't get broken in the future

Yuyz0112 (approved):

PR #1015, opened 2022-10-07
Separate out window monkeypatching into an optional rrweb-init.js 

eoghanmurray:
rrweb-init.js file is intended to be used much earlier in the page loading lifecycle (e.g. early injection from a browser extension) and which communicates with the main rrweb code via window.postMessage

### DIFF
--- a/packages/rrweb/rollup.config.js
+++ b/packages/rrweb/rollup.config.js
@@ -12,6 +12,12 @@ function toRecordPath(path) {
     .replace('rrweb', 'rrweb-record');
 }
 
+function toRecordInitPath(path) {
+  return path
+    .replace(/^([\w]+)\//, '$1/record/')
+    .replace('rrweb', 'rrweb-init');
+}
+
 function toRecordPackPath(path) {
   return path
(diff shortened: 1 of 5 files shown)


eoghanmurray:
- Where this patch might need more work is on the inter-window communication as the new init.js is designed to be included via the run_at: "document_start" hook, with the rest of rrweb using the normal content_script hook — scripts introduced by these different hooks live in different [isolated worlds](https://developer.chrome.com/docs/extensions/mv3/content_scripts/) and can't reference each others' context outside of postMessage.
 I had to add a means of referencing DOM nodes, and I came up with the ad-hoc method using css references: `canvas::nth-of-type(1)` , which could possibly fail if canvas elements were reordered

 - Other todo is to pull more early init stuff (like console.log, and what Justin mentioned) but it all needs testing and there is a question of whether the normal version of rrweb should also rely on these new inter-window communication methods rather than passing callback functions — that is an open question and potentially more work, but would mean that there would be less surface area for testing.

eoghanmurray:
I need to look at the work you did on cross-domain nested iframes — how postMessage is used

PR #1023, opened 2022-10-15
Use a custom Window object (aka: allow injection of rrweb from an iframe)

samber:
First, thanks for this awesome project 👏 

Currently, rrweb does not allow recording a user session from an iframe. This is a bit annoying for those who run rrweb in a third-party tool injected as a Javascript tag.

I would like to suggest a different implementation of #294.

#294 offers to look up the top `Window` object on the current page. This is a good start, but I would rather let the developer choose what `Window` HTML Element is the best: `window` or `window.top` or `window.parent` or `window.parent.parent.[...]` or `document.querySelector('iframe').contentWindow`...

PS: can anybody translate the doc to Chinese? :pray:


samber:
Can be tested with `stream.js` script, with `window: document.querySelector('#iframeResult').contentWindow` added line 50.

Tested with the following page: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_iframe

eoghanmurray:
OK I think building this on top of #294 is the way forward, however gonna query the core team for the best approach i.e. whether it's better not to assume that all users want the topmost window

PR #1024, opened 2022-10-15
Improve contributor experience

samber:
Hi there,

I'm adding some documentation for helping contributors get into the project.

Also adding an improvement to the "repl" script:
- when running the script in a loop, it defaults to the same URL instead of `https://react-redux.realworld.io`
- just type `<enter>` to fall back on the default URL instead of copy-pasting `https://react-redux.realworld.io` into the prompt


Mark-Fenng (approved):
Thank you so much.

PR #1028, opened 2022-10-17
Update Player.svelte: fix the bug.

### LINKED ISSUE #801: Show controls in fullscreen mode for rrweb-player
It would be great to add controls as part of the wrapper element used for toggling fullscreen mode.

MengZihan712:
# Before fixing the bug:
In full-screen mode, there is no progress bar or control panel.
Bugs appear in browsers: Chrome, Safari and Edge.
3491666005407_ pic

# Causes of bugs and repair ideas:
By checking the component code, the height of the content under the rr-player__frame tag is incorrectly calculated in full-screen mode, causing bugs to occur when the progress bar and control bar are rendered off-screen under the rr-controller tag. Fix the bug by modifying the way tag height is calculated.
3501666005410_ pic

# After the bug fix:
In full-screen mode, the progress bar and control bar can be displayed under the playback screen normally.
3511666005414_ pic

Close #801.

### DIFF
--- a/packages/rrweb-player/src/Player.svelte
+++ b/packages/rrweb-player/src/Player.svelte
@@ -24,7 +24,7 @@
   export let tags: Record<string, string> = {};
 
   let replayer: Replayer;
-  
+
   export const getMirror = () => replayer.getMirror();
 
   const controllerHeight = 80;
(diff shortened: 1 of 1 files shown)


Yuyz0112 (approved):
nice!

PR #1032, opened 2022-10-21
Fix: isBlocked throws on invalid element

dbseel:
If node is not an HTML element, matches throws an error. Use el instead.


dbseel:
@Mark-Fenng Are you able to review and approve this? Would like to move to using the latest rrweb but this bug is blocking us.

Mark-Fenng (approved):
Sorry for the late review. It's a good catch!

Juice10 (approved):

PR #1035, opened 2022-10-26
Cross origin iframe support

Juice10:
Adds support for recording cross origin iframes.
As long as rrweb is present in both top level frame a child iframe, the child iframe will send rrweb events to its parent for processing and emitting.

Demo: Cross origin iframe recording in action

https://user-images.githubusercontent.com/4106/199504830-2dc3772a-f73c-4122-a59a-808443f91688.mov

Demo: `canvas-webrtc` plugin also supported

https://user-images.githubusercontent.com/4106/199504844-4ad4a46b-9dc3-4841-9189-29714154ad92.mov

Breaking changes compared to rrweb 2.0.0.alpha.3 (but not present in rrweb 1.x):
 -  plugin `getMirror()` arguments have changed and now expects multiple mirrors as opposed to just one

Other changes:
 - We're now running all tests on GitHub-Actions as opposed to travis ci. Github Actions is faster (4 minute builds vs 5-7 minute builds) and gives us insight into any failing visual tests by uploading the failing diff image as an artefact on the run.

TODO:
 - [x] add documentation for cross origin iframe
 - [x] add zh_CN translations of documentation
 - [x] add cross-origin recording to repl


wfk007:
I am so excited about this great work!

.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,41 @@
+name: Tests
```
Yuyz0112, line 1: maybe we can merge this in a separate PR

packages/rrweb/src/record/cross-origin-iframe-mirror.ts
```diff
@@ -0,0 +1,95 @@
...
+    Map<number, number>
+  > = new WeakMap();
+  private iframeRemoteIdToParentIdMap: WeakMap<
+    HTMLIFrameElement,
+    Map<number, number>
+  > = new WeakMap();
+
+  constructor(private generateIdFn: () => number) {}
+
+  getParentId(
```
Yuyz0112, line 13: I think the naming in the mirror implementation can be improved a little more, `getParentId` is not quite intuitive in that it will set maps inside.

packages/rrweb/src/record/cross-origin-iframe-mirror.ts
```diff
@@ -0,0 +1,95 @@
+export default class CrossOriginIframeMirror {
+  private iframeParentIdToRemoteIdMap: WeakMap<
+    HTMLIFrameElement,
+    Map<number, number>
```
Yuyz0112, line 4: Shall we consider GC the id-to-id maps?

packages/rrweb/src/record/iframe-manager.ts
```diff
@@ -51,7 +83,200 @@ export class IframeManager {
...
+          ],
+          removes: [],
+          texts: [],
+          attributes: [],
+          isAttachIframe: true,
+        },
+      };
+    } else if (e.type === EventType.IncrementalSnapshot) {
+      switch (e.data.source) {
+        case IncrementalSource.Mutation: {
```
Yuyz0112, line 141: Is there any way we can force devs to update this part when they add a new kind of event 🤔

Yuyz0112:
100% agree with the current approach

packages/rrweb/src/record/cross-origin-iframe-mirror.ts
Yuyz0112: I mean looks like it will keep growing.

packages/rrweb/src/record/iframe-manager.ts
Juice10: Maybe we can do so in the tests, not sure how yet. 
Maybe just fail if the EvenType and IncrementalSource objects grow. And add some good documentation in that test explaining what to do when adding a new sub type.

I'm not sure if there is also a typescript way that we can have the compiler fail if it gets to the `else` or `default` statements.

packages/rrweb/src/record/cross-origin-iframe-mirror.ts
Juice10: To be honest I think you are right, but I'm not sure what would be better, what do you think of these? Or maybe someone else has some more suggestions.

Current situation:
getParentId
getRemoteId

Some options:
getLocalId
getRemoteId

getParentId
getCrossOriginId

getLocalId
getCrossOriginId

packages/rrweb/src/record/cross-origin-iframe-mirror.ts
Juice10: It's a weak map so it'll purge everything once the iframe disappears. Also on full snapshot this gets cleared too.
We could drop individual id pairs when looping through .data.removes, but I'm not sure if any events will arrive later for those ids that where removed, if that's the case a new set of ids will be generated for them

.github/workflows/ci-cd.yml
Juice10: Done: https://github.com/rrweb-io/rrweb/pull/1043

packages/rrweb/src/record/iframe-manager.ts
Juice10: > Is there any way we can force devs to update this part when they add a new kind of event 🤔

Found a way to do it with typescript: https://github.com/rrweb-io/rrweb/pull/1035/commits/2963357619cf35069aa23fdd51c3a787b9f15c05

Based on: https://stackoverflow.com/questions/59685270/typescript-ensure-all-cases-are-handled-for-union-type

packages/rrweb/src/record/cross-origin-iframe-mirror.ts
Juice10: I ended up going with `getId` and `getRemoteId`. Hope this is clearer

thatsnotright:
Thank you for dropping the ancestororigins use to let this support Firefox!

packages/rrweb/src/record/cross-origin-iframe-mirror.ts
Juice10: I did some more testing and we need to keep these ids around for as long as it’s on that page. After an id is deleted it could appear again (for example an element is moved, or removed from the dom and then added again later)

Mark-Fenng (approved):

Yuyz0112 (approved):

PR #1039, opened 2022-10-29
Implement: Inactive activity indicator on progress bar 

### LINKED ISSUE #621: Feature request: Inactive activity indicator on progressbar
The `rrweb-player` shows an indicator for snapshot events but it would also be nice, to have some sort of indication of the inactive portion of the recording during playback. Recorded events can span to several minutes and it might be useful to know before hand where the inactive part is so it can be skipped `manually` as-well

DexxDing:
# Before implementing this feature: 

There was no way to view which part of the replay is inactive.  
before

# After implementing this feature: 

The part of progress bar that contains non-active user behavior becomes gray so that the user can choose to skip it or not. We have created a function to detect inactive events or cycles and added a div tag to the progress bar within a cycle as an indicator 
after

# Demonstration video:

The screen on the left is the previous replayer, the one on the right is latest replayer after implementing this feature:

https://user-images.githubusercontent.com/82929929/198828965-1b157b57-d719-400c-9825-3233d6dd410f.mp4

close #621

### DIFF
--- a/packages/rrweb-player/src/Controller.svelte
+++ b/packages/rrweb-player/src/Controller.svelte
@@ -12,7 +12,7 @@
     createEventDispatcher,
     afterUpdate,
   } from 'svelte';
-  import { formatTime } from './utils';
+  import { formatTime, getInactivePeriods } from './utils';
   import Switch from './components/Switch.svelte';
 
   const dispatch = createEventDispatcher();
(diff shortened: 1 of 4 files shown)


packages/rrweb-player/src/Controller.svelte
```diff
@@ -95,6 +101,43 @@
...
+        tagEnd: number,
+      ) => {
+        const sessionDuration = endTime - startTime;
+        const eventDuration = tagEnd - tagStart;
+        const width = (eventDuration / sessionDuration) * 100;
+        return width.toFixed(2);
+      };
+      return periods.map((period) => ({
+        name: 'inactive period',
+        background: 'rgb(212 212 212)',
```
Yuyz0112, line 131: shall we add a props so user can override it?

Yuyz0112 (commented):
very cool, looks like we can move some utils to a shared package in the future!

packages/rrweb-player/src/Controller.svelte
DexxDing: Thanks for the suggestion! I will make the modifications ASAP.

Yuyz0112 (approved):
great

PR #1041, opened 2022-10-31
fix: bug when handling shadow doms

Mark-Fenng:
## Bug
On the website https://mixpanel.com/project/2195193/view/139237/app/dashboards#id=3679845, the bottom chart makes the recorder crash (in an infinite loop)

### How to reproduce:
1. Create an account and log in
2. goto the website and start recording
3. scroll down to the bottom of the page
4. when the chart is loaded, the whole browser would freeze (fall into an infinite loop)
image

### How to fix:

The chart is an SVG inside a shadow dom. The complicated dom structure raises a corner case for the recording code of shadow dom #956. In this case, the unhandled node is not a direct child of the shadow host and its parent node is not supposed to be the shadow host element. So I add code to check whether the unhandled node is a direct child of shadow dom or not. I tried to reproduce this through a simplified test case but I failed.

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -394,15 +394,20 @@ export default class MutationBuffer {
             }
             // nextId !== -1 && parentId === -1 This branch can happen if the node is the child of shadow root
             else {
-              const nodeInShadowDom = _node.value;
-              // Get the host of the shadow dom and treat it as parent node.
-              const shadowHost: Element | null = nodeInShadowDom.getRootNode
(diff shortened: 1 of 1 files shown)


Yuyz0112 (approved):

Yuyz0112:
LGTM, any chance to add some tests before we merge this?

Mark-Fenng:
I find it really hard to construct a test case for this. I tried to partially extracted the dom structure and the corner case didn't show up.

idosal:
Hey, thanks for fixing the bug! Do you happen to have an ETA for the fixed version?

PR #1042, opened 2022-11-03
fix: Failed to execute insertBefore on Node

wfk007:


Juice10:
Thank you for the thorough explanation, this must have been super time consuming to figure out, well done!
It would really be great if we could get a test that reproduces this failing state, that is fixed by the change you did. That way this issue will be solved forever and we won't have any regressions.

wfk007:
> Thank you for the thorough explanation, this must have been super time consuming to figure out, well done! It would really be great if we could get a test that reproduces this failing state, that is fixed by the change you did. That way this issue will be solved forever and we won't have any regressions.

ok, good idea

wfk007:
> Thank you for the thorough explanation, this must have been super time consuming to figure out, well done! It would really be great if we could get a test that reproduces this failing state, that is fixed by the change you did. That way this issue will be solved forever and we won't have any regressions.

@Juice10 Sorry for late!
I try to write a minimal reproducible code to generate events for test, but it keeps failing!
Additionally, I find some details in events of bad case:

image

generally speaking, id of node in childNodes should go after it in parentNode when taking fullSnapshot.
Id 65005 should be 8 in right case, and there are several wired id in events of bad case.
I event not sure how it really happened!

If new fullSnapshot are rebuild based on mirror with such wired ids. The mirror will be mistaken.
In case below:
- temp2 refer to domMirror
- temp1 refer to a textNode with id 82

image

omairnabiel:
Is this PR ready to be merged?

Juice10 (approved):

Mark-Fenng (approved):
Thanks for your great fix and thorough explanation!

PR #1043, opened 2022-11-03
Chore: Replace travis ci with github actions

Juice10:
We keep having to renew our Travis-ci credits and GitHub actions is a couple minutes quicker when it comes to running tests so it makes sense for us to migrate.
Also this change allows us to see what the broken visual tests look like as they are saved as artefacts on the test.


Yuyz0112 (approved):
great!

PR #1049, opened 2022-11-09
Fix: shadow dom bugs

Juice10:
Fixes a number of shadow dom bugs

- [ ] create test cases
- [ ] fix issues

### DIFF
--- a/packages/rrweb/test/integration.test.ts
+++ b/packages/rrweb/test/integration.test.ts
@@ -603,7 +603,97 @@ describe('record integration tests', function (this: ISuite) {
     });
     await page.waitForTimeout(50);
 
-    const snapshots = await page.evaluate('window.snapshots');
+    const snapshots = (await page.evaluate(
+      'window.snapshots',
+    )) as eventWithTime[];
+    assertSnapshot(snapshots);
+  });
+
+  it('should record shadow DOM 2', async () => {
+    const page: puppeteer.Page = await browser.newPage();
+    await page.goto('about:blank');
+    // page.on('console', (msg) => console.log(msg.text()));
+    await page.setContent(getHtml.call(this, 'blank.html'));
+
+    await page.evaluate(() => {
+      return new Promise((resolve) => {
+        /*
+      el = document.createElement('div');
+      el.attachShadow({ mode: 'open' });
+      el.shadowRoot.append(document.createElement('input'));
+      setTimeout(()=>{
+        document.body.append(el);
+      }, 500);
+      */
+
+        const el = document.createElement('div') as HTMLDivElement;
+        el.attachShadow({ mode: 'open' });
+        (el.shadowRoot as ShadowRoot).appendChild(
+          document.createElement('input'),
+        );
+        setTimeout(() => {
+          document.body.append(el);
+          resolve(null);
+        }, 1000);
+      });
+    });
+    await waitForRAF(page);
+
+    const snapshots = (await page.evaluate(
+      'window.snapshots',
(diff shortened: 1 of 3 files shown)


packages/rrweb/src/replay/index.ts
```diff
@@ -1482,6 +1482,9 @@ export class Replayer {
       const target = buildNodeWithSN(mutation.node, {
         doc: targetDoc as Document, // can be Document or RRDocument
         mirror: mirror as Mirror, // can be this.mirror or virtualDom.mirror
+        /**
+         * shadowHosts contain a snapshot of shadow dom nodes which must also be added.
+         */
```
Juice10, line 1487:
```suggestion
```

packages/rrweb/test/integration.test.ts
```diff
@@ -603,7 +655,122 @@ describe('record integration tests', function (this: ISuite) {
...
+      return new Promise((resolve) => {
+        const el = document.createElement('div') as HTMLDivElement;
+        el.attachShadow({ mode: 'open' });
+        (el.shadowRoot as ShadowRoot).appendChild(
+          document.createElement('input'),
+        );
+        setTimeout(() => {
+          document.body.append(el);
+          resolve(null);
+        }, 1000);
```
Juice10, line 678:
```suggestion
        setTimeout(() => {
          document.body.append(el);
          resolve(null);
        }, 10);
```

packages/rrweb-snapshot/test/rebuild.test.ts
```diff
@@ -47,6 +47,37 @@ describe('rebuild', function () {
     });
   });
 
+  describe('shadowDom', function () {
+    it('should rebuild shadowRoot with siblings', function () {
```
Mark-Fenng, line 51: The name of this test case is 'build shadowRoot with siblings' but actually only the shadowRoot itself is rebuilt without any siblings.

packages/rrweb/src/record/processed-node-manager.ts
```diff
@@ -0,0 +1,40 @@
+import type MutationBuffer from './mutation';
+
+/**
+ * Keeps a log of nodes that could show up in multiple mutation buffer but shouldn't be handled twice.
+ */
+export default class ProcessedNodeManager {
+  private nodeSet: WeakSet<Node> = new WeakSet();
```
Mark-Fenng, line 7: This nodeSet (including `has(node: Node)` function) is not used anywhere, so maybe removing it is a good idea.

Mark-Fenng (commented):
I apologise for my late review. It took me a long time to understand the logic of ProcessedNodeManager as well as the new appended shadow dom tests. They are actually essential and delicate. There are two points which I think can be improved.

packages/rrweb-snapshot/test/rebuild.test.ts
Juice10: Good catch @Mark-Fenng 
```suggestion
    it('rebuild shadowRoot without siblings', function () {
```

Mark-Fenng (approved):

bhavitsharma:
@Mark-Fenng @Juice10 
I think the implementation of processed node manager is causing some regression in the rrweb related to the order of scroll and click events. I am pretty sure `requestAnimationFrame` is causing some kind of delay but need to confirm again.

Juice10:
@bhavitsharma could you create a separate issue for this including a test case and/or example recording where this is going wrong? Looking at the requestAnimationFrame code, it shouldn’t impact any scroll or click events.

PR #1078, opened 2022-12-18
fix: do not assume event listeners are passive

aptlin:
Without this rrweb interferes with the behavior of the `onPointermove` listener of the standard drag and drop Angular component, which calls `preventDefault` internally.

To reproduce, compare the console logs of `yarn repl` on `https://v7.material.angular.io/cdk/drag-drop/overview`:

- on master: https://www.loom.com/share/ce576beab57c4b0a9a29073b59af9ebb
- with this change: https://www.loom.com/share/880381abb15f42009b21536b94f4184b

This has already been reported as an issue for session recorders like Logrocket:

https://github.com/angular/components/issues/24979

### DIFF
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -196,9 +196,9 @@ function initMoveObserver({
     },
   );
   const handlers = [
-    on('mousemove', updatePosition, doc),
-    on('touchmove', updatePosition, doc),
-    on('drag', updatePosition, doc),
+    on('mousemove', updatePosition, doc, false),
+    on('touchmove', updatePosition, doc, false),
+    on('drag', updatePosition, doc, false),
   ];
   return () => {
     handlers.forEach((h) => h());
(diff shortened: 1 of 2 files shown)


aptlin:
@Juice10 what do you think about this?

Juice10:
Looks good @aptlin thanks for contributing!

Juice10 (approved):

aptlin:
@Juice10 looks like not marking all events as passive [does hit](https://github.com/rrweb-io/rrweb/actions/runs/3722831577/jobs/6398215176) the perf too much, fixed in [f4705c3](https://github.com/rrweb-io/rrweb/pull/1078/commits/f4705c389919965837e51bff0a60ad80e19e7f7f). Could you approve running the tests again please?

eoghanmurray:
@aptlin
in f4705c3 you added `on('mousemove', updatePosition, doc, false)` which I believe turns off passive for mousemove (other lines do the same for touchmove & drag).
Was that intended?  Just wondering as you mentioned that that change fixes perf.

> This has already been reported as an issue for session recorders like Logrocket:

Logrocket also would use the passive flag for a good reason; would a better solution be to put a `try {} catch() {}` around the preventDefault call in the angular project?

PR #1084, opened 2023-01-06
Chore: Make tests less flakey & upgrade puppeteer to rrweb-snapshot test suite to run

Juice10:
Tests randomly fail making it very hard to figure out if a PR broke something. This PR fixes a lot of that

### DIFF
--- a/packages/rrweb-snapshot/test/__snapshots__/integration.test.ts.snap
+++ b/packages/rrweb-snapshot/test/__snapshots__/integration.test.ts.snap
@@ -189,7 +189,7 @@ exports[`integration tests [html file]: block-element.html 1`] = `
 `;
 
 exports[`integration tests [html file]: compat-mode.html 1`] = `
-"<!DOCTYPE html PUBLIC \\"-//W3C//DTD HTML 4.0 Transitional//EN\\"><!-- no doctype! --><html><head>
+"<!-- no doctype! --><html><head>
   <title>Compat Mode; image resizing</title>
 </head>
 <body>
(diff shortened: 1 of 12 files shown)


packages/rrweb-snapshot/test/__snapshots__/integration.test.ts.snap
```diff
@@ -287,12 +287,12 @@ exports[`integration tests [html file]: iframe.html 1`] = `
 `;
 
 exports[`integration tests [html file]: iframe-inner.html 1`] = `
-"<!DOCTYPE html PUBLIC \\"-//W3C//DTD HTML 4.0 Transitional//EN\\"><html><head></head><body><button>inner iframe button</button>
+"<html><head></head><body><button>inner iframe button</button>
```
eoghanmurray, line 290: These Transitional DOCTYPE here were added deliberately in #697

Juice10: Newer versions of Chrome strip them out if they aren’t in standards mode I think

Mark-Fenng (approved):

packages/rrweb-snapshot/test/__snapshots__/integration.test.ts.snap
Juice10: I had a moment to dig into what is going on. Newer versions of chrome don't add a transitional doctype if you never specified one to begin with. Chrome just leaves it empty. 

iframe-inner.html:
image

what chrome renders:
image

I went through each of the changes in this PR and compared them to their source document and this does seem correct.

Example of the `compat-mode.html` file:
image

How chrome renders it:
image
(consistent with the snapshot)

Mark-Fenng (approved):

packages/rrweb-snapshot/test/__snapshots__/integration.test.ts.snap
eoghanmurray: I'm not understanding!
The transitional doctypes are there in the source test files in order to test that the resultant replay also replays with compatMode=true
Maybe ye guys are one step ahead of me?

Juice10: There where never any doctypes in the source files. Not having doctypes also triggers compat-mode. Check this out: 

https://github.com/rrweb-io/rrweb/blob/master/packages/rrweb-snapshot/test/html/compat-mode.html

https://github.com/rrweb-io/rrweb/blob/master/packages/rrweb-snapshot/test/html/iframe-inner.html

eoghanmurray: Ah okay got it now!
Only problem I can think of is if future versions of Chrome decide that lack of a doctype should imply HTML5 ... but happy to defer worrying about that to the future.

PR #1086, opened 2023-01-09
Fix missed adopted style sheets of shadow doms in checkout full snapshot

Mark-Fenng:
### Bug
AdoptedStyleSheets in shadow doms are missed when a full snapshot is checked out after the recording has started.

This bug can be reproduced through the created test case.
### Error test snapshot:
adoptedStylesheets data is missed in the second full snapshot.
### Expected test snapshot:
New adoptedStylesheets data is also generated in the second full snapshot.

### DIFF
--- a/packages/rrweb/test/record.test.ts
+++ b/packages/rrweb/test/record.test.ts
@@ -591,27 +591,30 @@ describe('record', function (this: ISuite) {
 
   it('captures adopted stylesheets of shadow doms in checkout full snapshot', async () => {
     await ctx.page.evaluate(() => {
-      document.body.innerHTML = `
-        <div id="shadow-host-1">entry</div>
-      `;
+      return new Promise((resolve) => {
+        document.body.innerHTML = `
+            <div id="shadow-host-1">entry</div>
+          `;
 
-      let shadowHost = document.querySelector('div')!;
-      shadowHost!.attachShadow({ mode: 'open' });
-      const sheet = new CSSStyleSheet();
-      sheet.replaceSync!('h1 {color: blue;}');
-      shadowHost.shadowRoot!.adoptedStyleSheets = [sheet];
+        let shadowHost = document.querySelector('div')!;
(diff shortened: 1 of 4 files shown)


packages/rrweb/test/record.test.ts
```diff
@@ -586,6 +589,32 @@ describe('record', function (this: ISuite) {
...
+        emit,
+      });
+
+      setTimeout(() => {
+        // When a full snapshot is checked out manually, all adoptedStylesheets should also be captured.
+        rrweb.record.takeFullSnapshot(true);
+      }, 10);
+    });
+    await ctx.page.waitForTimeout(200);
+    assertSnapshot(ctx.events);
```
Juice10, line 615: This is less flakey because it doesn't depend on `waitForTimeout` (which is also deprecated in newer versions of puppeteer)

```suggestion
    await ctx.page.evaluate(() => {
      return new Promise((resolve) => {
        document.body.innerHTML = `
            <div id="shadow-host-1">entry</div>
          `;

        let shadowHost = document.querySelector('div')!;
        shadowHost!.attachShadow({ mode: 'open' });
        const sheet = new CSSStyleSheet();
        sheet.replaceSync!('h1 {color: blue;}');
        shadowHost.shadowRoot!.adoptedStyleSheets = [sheet];

        const { rrweb, emit } = (window as unknown) as IWindow;
        rrweb.record({
          emit,
        });

        setTimeout(() => {
          // When a full snapshot is checked out manually, all adoptedStylesheets should also be captured.
          rrweb.record.takeFullSnapshot(true);
          resolve(undefined);
        }, 10);
      });
    });
    await waitForRAF(page);
    assertSnapshot(ctx.events);
```

Mark-Fenng: Thanks for the suggestion! I will keep this in mind for later test cases.

Juice10 (approved):
Looks great @Mark-Fenng

PR #1087, opened 2023-01-10
fix: module error

### LINKED ISSUE #1056: [Bug]: ReferenceError: exports is not defined in ES module scope
### Preflight Checklist

rrweb

### Version

"rrweb": "^2.0.0-alpha.4",

### Expected Behavior

To start recording when the page is loaded.

### Actual Behavior

image

### Steps to Reproduce

https://stackblitz.com/edit/next-typescript-nfhzmu?file=pages%2Findex.tsx

### Testcase Gist URL

_No response_

### Additional Information

    "next": "^12.3.0",
    "react": "^17.0.2",
    "react-dom": "^17.0.2",
    "rrweb": "^2.0.0-alpha.4",

Mark-Fenng:
Fix #1056 

### Reason
In #976, a property `type: "module"` was added to the package.json. When it is set to "module", the type field allows a package to specify all .js files within are ES modules. If the "type" field is omitted or set to "commonjs", all .js files are treated as CommonJS. So all scripts under rrweb/lib/** are treated as esm format by JS bundlers unexpectedly.

### Fix
Change all suffixes of bundled scripts with commonjs module from 'js' to cjs.

### ⚠️Breaking Change⚠️:
The file suffix of all scripts under rrweb/lib/* is changed from **js** to **cjs**. This may cause programs that import rrweb through CDN links to fail.

### DIFF
--- a/packages/rrdom-nodejs/package.json
+++ b/packages/rrdom-nodejs/package.json
@@ -15,7 +15,7 @@
     "rrdom-nodejs"
   ],
   "license": "MIT",
-  "main": "lib/rrdom-nodejs.js",
+  "main": "lib/rrdom-nodejs.cjs",
   "module": "es/rrdom-nodejs.js",
   "typings": "es",
   "files": [
--- a/packages/rrdom/jest.config.js
+++ b/packages/rrdom/jest.config.js
@@ -1,5 +1,5 @@
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
-module.exports = {
+export default {
   preset: 'ts-jest',
   testEnvironment: 'node',
 };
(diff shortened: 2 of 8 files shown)


Juice10 (approved):
This is the way forward IMO, and 2.0 is the right moment to do it. I believe I did something similar in the vite.js branch, but that one isn’t finished yet so I’m happy we could get this in sooner

eoghanmurray:
> This may cause programs that import rrweb through CDN links to fail.

This is the case with jsdelivr I believe; e.g. the following fails as it's served as `application/node`:

```
<script src="https://cdn.jsdelivr.net/npm/rrweb-player@2.0.0-alpha.18/dist/rrweb-player.umd.cjs"></script>
```

PR #1088, opened 2023-01-10
feat: Guard against missing `window.CSSStyleSheet`

mydea:
We get a lot of reports from users of sentry that seem to originate from browsers where `window.CSSStyleSheet` is not available. See e.g. https://github.com/getsentry/sentry-javascript/issues/6534

This adds a simple guard that skips stylesheet observing when this is not defined.

### DIFF
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -536,6 +536,13 @@ function initStyleSheetObserver(
   { styleSheetRuleCb, mirror, stylesheetManager }: observerParam,
   { win }: { win: IWindow },
 ): listenerHandler {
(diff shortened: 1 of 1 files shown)


Mark-Fenng (approved):
Thanks for the contribution. The guard is good and I think can add it for sure. BTW, do you figure out the root cause of the error? I'm worried that this PR may not fix the bug from the root.

Juice10 (approved):
Im surprised this is needed since the CSSStyleSheet API seems to be supported by all browsers: https://caniuse.com/mdn-api_cssstylesheet

mydea:
Yeah, we thought the same, but kept getting bug reports for this. I'd say one of two things is happening:

Either A) it is somehow related to some rather obscure browser or B) something is messing with the global objects. 😬

Issue #941, opened 2022-07-20
[Bug]: ShadowHost can't be a string


pavlogavryliuk:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Expected Behavior

Some HTML Element `getRootNode().host` method returns a `String` type, but RRWeb expected type is `Element` or `null`.

![image](https://user-images.githubusercontent.com/46223649/179981855-e7575adb-1074-43f9-87dd-d97ef79797a2.png)

### How to reproduce?
1) [Bitbucket](https://bitbucket.org/dashboard/overview) -> Recent repositories -> Go to any repository
2) Same behavior on [CNN](https://edition.cnn.com/2022/07/18/africa/ghana-first-marburg-outbreak-intl/index.html) pages

### What's happening?
It's [mutation.ts](https://github.com/rrweb-io/rrweb/blob/f3064c1f2d0ae1f5519afeb7e7d1fe7bde014502/packages/rrweb/src/record/mutation.ts#L264) file:

`const shadowHost = n.getRootNode().host` returns a `String` type, but expected type is `Element` or `null`:

![image](https://user-images.githubusercontent.com/46223649/179981855-e7575adb-1074-43f9-87dd-d97ef79797a2.png)

`this.doc.contains(shadowHost)` - if `shadowHost` is a `String` type, this line is crashing rrweb recording:

![image](https://user-images.githubusercontent.com/46223649/179981893-1c4decb5-7a72-45ff-9b46-2c4f3e54681d.png)

PR #1092, opened 2023-01-13
fix: regression of issue: ShadowHost can't be a string (issue 941)

Mark-Fenng:
### Bug
image
This PR is to fix a bug which is a regression of #941.
The original one has been fixed but imported again in #1049 

### Reason
Given the text node of a detached HTMLAnchorElement, getRootNode() will return the anchor element itself and its host property is a string.
image

### Fix
The fix is like #1020. I added unit tests to cover all key codes and I hope this bug won't show up anymore.

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -27,6 +27,7 @@ import {
   isSerializedIframe,
   isSerializedStylesheet,
   inDom,
+  getShadowHost,
 } from '../utils';
 
 type DoubleLinkedListNode = {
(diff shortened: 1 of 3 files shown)


Juice10:
The problem here is that ShadowRoot and HTMLAnchorElement have colliding `host` properties. In the jsdom unit tests created this passes. But re-reading the code a couple times  I can’t see how the `host` collision issue is bypassed.

Mark-Fenng:
> The problem here is that ShadowRoot and HTMLAnchorElement have colliding `host` properties. In the jsdom unit tests created this passes. But re-reading the code a couple times I can’t see how the `host` collision issue is bypassed.

I think the problem in the left side is this line
image
https://github.com/rrweb-io/rrweb/pull/1092/files#diff-048da15f20b717b1ea5aa5346398dfed2fe4e24a7d38445b94cbe707640c9fc5L523
It retrieves the host property without checking the element type.

Here is the process how the error happens:
The input of inDom() is a text node in <a>
image
Then we try to get its rootShadowHost. 
image
Finally the string 'github.com' is returned as the "rootShadowHost".

bhavitsharma:
@Mark-Fenng I am a bit lost here, we do check if the `getRootNode().nodeType` is a Document Fragment so why did the check fail?

Mark-Fenng:
> @Mark-Fenng I am a bit lost here, we do check if the `getRootNode().nodeType` is a Document Fragment so why did the check fail?

https://github.com/rrweb-io/rrweb/pull/1092/files#diff-048da15f20b717b1ea5aa5346398dfed2fe4e24a7d38445b94cbe707640c9fc5L523
This line didn't check the document check

bhavitsharma:
Ah, I understand it better now. Thanks. Can we merge this PR then?

Mark-Fenng:
> Ah, I understand it better now. Thanks. Can we merge this PR then? 

We need at least one reviewer to approve this before merging it.

bhavitsharma (approved):

Juice10 (approved):

PR #1093, opened 2023-01-16
Fix cross origin iframe bugs

### LINKED ISSUE #1077: [Bug]: recordCrossOriginIframes: true does not work with pack/unpack fn
### Preflight Checklist

rrweb

### Version

latest from master

### Expected Behavior

I expect recordCrossOriginIframes will show iframe content when injected in to all iframes, with a pack/unpackfn specified.

### Actual Behavior

The mutations are captured for iframes but not processed unless I remove the packFn.  If I specify a packFn for only the top-level rrweb this also works, so it's just in the child to parent messaging that the packFn causes this behavior.

### Steps to Reproduce

Use the repl to get an IFrame site, and then add a packFn and see that it no longer renders cross-domain IFrame content.

I have an idea where to fix this that I'm looking at locally and I can submit a PR if I fix this behavior before anyone else does.

### Testcase Gist URL

_No response_

### Additional Information

_No response_

Mark-Fenng:
### Bug 1
image

Recorder generates error data while recording some websites are integrated with stripe
   These websites will usually have an iframe with src "https://js.stripe.com/v3/m-outer-xxx.html"

### Reason
The Stripe site has nested iframes and they all have a message proxy to forward messages from child iframe to parent iframe. So cross-origin rrweb messages are also forwarded by this hidden proxy and it makes the recorder generate redundant iframe full snapshots.
digram

### Fix
Add an `origin` property to each cross-origin rrweb message for the parent frame to check the integrity of messages.

### Bug 2
fixes #1077
As Justin's suggestion https://github.com/rrweb-io/rrweb/issues/1077#issuecomment-1373495719, I disabled packing for inter-iframe communication.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -30,7 +30,7 @@ jobs:
         run: yarn turbo run check-types
 
       - name: Run tests
-        run: xvfb-run --server-args="-screen 0 1920x1080x24" yarn test
+        run: PUPPETEER_HEADLESS=true xvfb-run --server-args="-screen 0 1920x1080x24" yarn test
 
       - name: Upload diff images to GitHub
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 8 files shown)


Juice10 (approved):
Great stuff @Mark-Fenng!

PR #1095, opened 2023-01-17
fix duplicated shadow dom

YunFeng0817:
### Bug
There are piles of duplicated nodes after a recorder takes a new full snapshot manually.
Duplicated node creation may cause unexpected errors in the replayer.

### Fix
1. Clean observers of shadow dom manager before taking new full snapshots.
2. Add a fallback on the replayer side to make it stable when data has duplicated nodes for other kinds of reasons.


bhavitsharma:
Hey @Mark-Fenng, why do you mean by taking fullsnapshot manually? and why are there duplicated nodes? Do you mean taking fullsnapshot by manually starting the recording again (i.e. after rrweb has already taken a full-snapshot)? Why do we need this workaround then? Shouldn't be we cleaning up all the observers in that case?

packages/rrweb/src/record/shadow-dom-manager.ts
```diff
@@ -135,6 +141,14 @@ export class ShadowDomManager {
 
   public clearCache() {
     this.shadowDoms = new WeakSet();
+    this.observerHandlers.forEach((handler) => {
```
bhavitsharma, line 144: OK, I think I understand it better now. Can we reuse the variable `restorePatches` here. Also, do you think in the index.ts file we should be calling the `reset` function instead of `clearCache` function? Otherwise, we'll be repatching the shadow-dom observer again.

bhavitsharma: i.e. in this PR https://github.com/rrweb-io/rrweb/pull/1086/files#diff-ad746ef2bb1aa2d230a3b8635d3fd20a4587dc176bb71f42065a857392b755b3, we should be calling the reset function and not the clearCache function.

packages/rrweb/src/record/shadow-dom-manager.ts
```diff
@@ -135,6 +141,14 @@ export class ShadowDomManager {
...
+      try {
+        handler();
+      } catch (e) {
+        //
+      }
+    });
+    this.observerHandlers = [];
   }
 
   public reset() {
```
bhavitsharma, line 154: Unrelated to this PR but should be putting the `try/catch` here as well? 
```
this.restorePatches.forEach((restorePatch) => {
     try { restorePatch() } catch (e) {} 
});
```

Mark-Fenng:
> Hey @Mark-Fenng, why do you mean by taking fullsnapshot manually? and why are there duplicated nodes? Do you mean taking fullsnapshot by manually starting the recording again (i.e. after rrweb has already taken a full-snapshot)? Why do we need this workaround then? Shouldn't be we cleaning up all the observers in that case? 

Taking fullsnapshot manually means call rrweb.takeFullSnapshot(true) after the recorder started.

bhavitsharma (approved):
Thank you for making suggestions. The PR looks great to me!

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -213,3 +219,30 @@ export function is2DCanvasBlank(canvas: HTMLCanvasElement): boolean {
...
+    a.type === NodeType.Comment ||
+    a.type === NodeType.Text ||
+    a.type === NodeType.CDATA
+  )
+    return a.textContent === (b as textNode).textContent;
+  else if (a.type === NodeType.Element)
+    return (
+      a.tagName === (b as elementNode).tagName &&
+      JSON.stringify(a.attributes) ===
+        JSON.stringify((b as elementNode).attributes) &&
```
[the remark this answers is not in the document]
Mark-Fenng: > This is quite a slow call to do, aren't you afraid this will slow down repeated `rrweb.takeFullSnapshot(true)`?

Yes you are right, this function is slow. It's in the rebuilt.ts so I don't think this can slow down the takeFullSnapshot function. If the data events are correct, the program won't go to this path. This comparison will only be used when rrweb ingests any problematic data.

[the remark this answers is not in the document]
Mark-Fenng: > Also what happens if a node gets a new attribute since the last time it was snapshotted? Will this node get added twice?

If the recorder works well, the incremental data will only change the node's attribute rather than totally recreate the node through this path. 
In case the program goes here, the node will get added twice as you said. But honestly, I don't know what kind of error event data can lead to this. I wrote this function as a fallback.
Do you have any suggestions for improving the code here?

Juice10 (approved):

PR #903, opened 2022-05-25
Speed up snapshotting of many new dom nodes

Juice10:
By avoiding reflow we shave about 15-25% off our snapshotting time.
By doing some other tweaks we can bring our savings up to about 40-50%.

Before:
image
image
image

After:
image
image
image

PR #1097, opened 2023-01-17
text masking settings apply to inputs

### LINKED ISSUE #874: maskTextClass and maskTextSelector not working
`maskTextClass` and `maskTextSelector` and `blockSelector` are not working on `<input type="text">` for some reason. Same thing with the default `rr-mask` class. 

But `maskAllInputs` and `maskInputOptions: {text: true}` and `rr-block` works fine.

### LINKED ISSUE #1096: [Feature Request]: text & input masking enhancements
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a feature request that matches the one I want to file, without success.

rrweb

### Problem Description

I'm looking to make enhancements to text & input masking (PR to follow, for your consideration). My team is getting ready for a rrweb deployment and we'd like to offer "privacy by default" to avoid collecting sensitive information. My understanding is that the current way to accomplish this is `maskTextSelector: '*'`, however, we would also like to offer the ability to unmask portions of the page that the customer deems safe.

### Proposed Solution

* add `maskAllText` option, mirroring the naming of `maskAllInputs`
* add `unmaskTextSelector` and `unmaskTextClass` to selectively unmask certain areas.
* inputs respect text masking options in addition to input masking options for finer control over which inputs are masked or unmasked.

### Alternatives Considered

I'm not aware of any current solutions for unmasking content with rrweb.

### Additional Information

_No response_

mdellanoce:
* inputs respect text masking options in addition to input masking options
* fixes `maskInputFn` not passed to initial snapshot
* fixes #874
* adds a fast return from needMaskingText if maskTextSelector is `*`

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -321,6 +321,7 @@ export function needMaskingText(
         ? (node as HTMLElement)
         : node.parentElement;
     if (el === null) return false;
+    if (maskTextSelector === '*') return true;
     if (typeof maskTextClass === 'string') {
       if (checkAncestors) {
         if (el.closest(`.${maskTextClass}`)) return true;
@@ -503,11 +504,14 @@ function serializeNode(
         keepIframeSrcFn,
         newlyAddedElement,
         rootId,
+        needsMask,
       });
     case n.TEXT_NODE:
       return serializeTextNode(n as Text, {
         needsMask,
         maskTextFn,
+        maskInputOptions,
+        maskInputFn,
         rootId,
       });
     case n.CDATA_SECTION_NODE:
@@ -538,16 +542,20 @@ function serializeTextNode(
   options: {
     needsMask: boolean | undefined;
     maskTextFn: MaskTextFn | undefined;
+    maskInputOptions: MaskInputOptions;
+    maskInputFn: MaskInputFn | undefined;
     rootId: number | undefined;
   },
 ): serializedNode {
-  const { needsMask, maskTextFn, rootId } = options;
+  const { needsMask, maskTextFn, maskInputOptions, maskInputFn, rootId } =
+    options;
   // The parent node may not be a html element which has a tagName attribute.
   // So just let it be undefined which is ok in this use case.
   const parentTagName = n.parentNode && (n.parentNode as HTMLElement).tagName;
   let textContent = n.textContent;
   const isStyle = parentTagName === 'STYLE' ? true : undefined;
   const isScript = parentTagName === 'SCRIPT' ? true : undefined;
+  const isTextarea = parentTagName === 'TEXTAREA' ? true : undefined;
   if (isStyle && textContent) {
     try {
       // try to read style sheet
@@ -577,6 +585,11 @@ function serializeTextNode(
       ? maskTextFn(textContent, n.parentElement)
       : textContent.replace(/[\S]/g, '*');
   }
+  if (isTextarea && textContent && maskInputOptions.textarea) {
+    textContent = maskInputFn
+      ? maskInputFn(textContent, n.parentNode as HTMLElement)
+      : textContent.replace(/[\S]/g, '*');
+  }
 
   return {
     type: NodeType.Text,
@@ -604,6 +617,7 @@ function serializeElementNode(
      */
     newlyAddedElement?: boolean;
     rootId: number | undefined;
+    needsMask?: boolean;
   },
 ): serializedNode | false {
   const {
@@ -619,6 +633,7 @@ function serializeElementNode(
     keepIframeSrcFn,
     newlyAddedElement = false,
     rootId,
+    needsMask,
   } = options;
   const needBlock = _isBlockedElement(n, blockClass, blockSelector);
   const tagName = getValidTagName(n);
@@ -682,6 +697,7 @@ function serializeElementNode(
         value,
         maskInputOptions,
         maskInputFn,
+        needsMask,
       });
     } else if (checked) {
       attributes.checked = checked;
@@ -1246,7 +1262,7 @@ function snapshot(
     inlineStylesheet?: boolean;
     maskAllInputs?: boolean | MaskInputOptions;
     maskTextFn?: MaskTextFn;
-    maskInputFn?: MaskTextFn;
+    maskInputFn?: MaskInputFn;
     slimDOM?: 'all' | boolean | SlimDOMOptions;
     dataURLOptions?: DataURLOptions;
     inlineImages?: boolean;
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -4,6 +4,7 @@ import {
   Mirror,
   getInputType,
   toLowerCase,
+  needMaskingText,
 } from 'rrweb-snapshot';
 import type { FontFaceSet } from 'css-font-loading-module';
 import {
@@ -420,6 +421,8 @@ function initInputObserver({
   maskInputFn,
   sampling,
   userTriggeredOnInput,
+  maskTextClass,
+  maskTextSelector,
 }: observerParam): listenerHandler {
   function eventHandler(event: Event) {
     let target = getEventTarget(event) as HTMLElement | null;
@@ -452,11 +455,19 @@ function initInputObserver({
     let isChecked = false;
     const type: Lowercase<string> = getInputType(target) || '';
 
+    const needsMask = needMaskingText(
+      target as Node,
+      maskTextClass,
+      maskTextSelector,
+      true,
+    );
+
     if (type === 'radio' || type === 'checkbox') {
       isChecked = (target as HTMLInputElement).checked;
     } else if (
       maskInputOptions[tagName.toLowerCase() as keyof MaskInputOptions] ||
-      maskInputOptions[type as keyof MaskInputOptions]
+      maskInputOptions[type as keyof MaskInputOptions] ||
+      needsMask
     ) {
       text = maskInputValue({
         element: target,
@@ -465,6 +476,7 @@ function initInputObserver({
         type,
         value: text,
         maskInputFn,
+        needsMask,
       });
     }
     cbWithDedup(
--- a/packages/rrweb/test/__snapshots__/integration.test.ts.snap
+++ b/packages/rrweb/test/__snapshots__/integration.test.ts.snap
@@ -4960,6 +4960,665 @@ exports[`record integration tests can use maskInputOptions to configure which ty
 ]"
 `;
 
+exports[`record integration tests can use maskTextSelector to configure which inputs should be masked 1`] = `
+"[
+  {
+    \\"type\\": 0,
+    \\"data\\": {}
+  },
+  {
+    \\"type\\": 1,
+    \\"data\\": {}
+  },
+  {
+    \\"type\\": 4,
+    \\"data\\": {
+      \\"href\\": \\"about:blank\\",
+      \\"width\\": 1920,
+      \\"height\\": 1080
+    }
+  },
+  {
+    \\"type\\": 2,
+    \\"data\\": {
+      \\"node\\": {
+        \\"type\\": 0,
+        \\"childNodes\\": [
+          {
+            \\"type\\": 1,
+            \\"name\\": \\"html\\",
+            \\"publicId\\": \\"\\",
+            \\"systemId\\": \\"\\",
+            \\"id\\": 2
+          },
+          {
+            \\"type\\": 2,
+            \\"tagName\\": \\"html\\",
+            \\"attributes\\": {
+              \\"lang\\": \\"en\\"
+            },
+            \\"childNodes\\": [
+              {
+                \\"type\\": 2,
+                \\"tagName\\": \\"head\\",
+                \\"attributes\\": {},
+                \\"childNodes\\": [
+                  {
+                    \\"type\\": 3,
+                    \\"textContent\\": \\"\\\\n    \\",
+                    \\"id\\": 5
+                  },
+                  {
+                    \\"type\\": 2,
+                    \\"tagName\\": \\"meta\\",
+                    \\"attributes\\": {
+                      \\"charset\\": \\"UTF-8\\"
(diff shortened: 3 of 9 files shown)


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -280,45 +280,110 @@ export function classMatchesRegex(
...
   if (el === null) return false;
 
-  if (typeof maskTextClass === 'string') {
-    if (el.classList.contains(maskTextClass)) return true;
-    if (el.closest(`.${maskTextClass}`)) return true;
-  } else {
-    if (classMatchesRegex(el, maskTextClass, true)) return true;
-  }
+  const maskDistance = distanceToMatch(el, maskTextClass, maskTextSelector);
+  const unmaskDistance = distanceToMatch(
```
Mark-Fenng, line 374: I think skipping this calculation when `maskAllText` is false could improve the performance.

Juice10:
Hi @mdellanoce, thanks for creating this Pull Request, I think this is a great idea to help give people more privacy and this should definitely be included in rrweb.

I did notice one potential issue with this implementation: the `needMaskingText` function has been a performance problem in the past because it gets called on every single node that gets recorded (sometimes tens-of-thousands or even hundreds of thousands of nodes get recorded at once), and because the new `needMaskingText` now compares distances for both masked and unmasked parents its doing twice as much work.
Maybe you could implement some escape hatches to simplify the work it has to do, for example if the is no masked parent, then there is no need to figure out what the unmasked distance is.

Thanks again for submitting this, I think if we can get this performant this would be a great addition to rrweb!

mdellanoce:
@Mark-Fenng @Juice10 thanks for the feedback, totally understand the performance concerns. I think the maskAllText suggestion will work, and I have a few other ideas. I'll see what I can do and hopefully ping you back in a few days.

packages/rrweb-snapshot/src/snapshot.ts
mdellanoce: updated this so that if maskAllText is true, and unmasking either is not configured (best case) or fails to find a match (worst case), then the masking distance is not computed.

mdellanoce:
@YunFeng0817 @Juice10 sorry for the delay, but I added 3 commits to address performance:

1. if maskAllText is true, and unmasking either is not configured (best case) or fails to find a match (worst case), then the masking distance is not computed, and needsMaskingText returns true early
2. I removed the default for `unmaskTextClass`, so that unmasking overhead won't "surprise" anyone with existing setups until they opt-in to use it.
3. I reworked the distance computation so that it is a single walk from the element to the root node. Each step will check for class match first, then check selectors. Also, in the masking distance computation, if the computed distance at a given step exceeds what was found for the unmasking distance, it'll stop walking upwards immediately and return a non-match.

Hope that all makes sense. Let me know what you think.

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -660,12 +758,21 @@ function serializeElementNode(
...
+        unmaskTextSelector,
+        maskAllText,
+      );
       attributes.value = maskInputValue({
         type: attributes.type,
         tagName,
         value,
         maskInputOptions,
         maskInputFn,
+        forceMask,
```
billyvg, line 775: Does this mean that `maskAllText` will also mask inputs? Do we need `(un)maskInput(Selector|Class)` as well so they can be selectively (un)masked?

mdellanoce: the masking and unmasking options in general apply to inputs, including maskAllText. So I was thinking unmaskSelector/Class already cover this.

the forceMask parameter is a little awkward, maybe there's a better way to do that. I'm evaluating the masking/unmasking settings outside of this function call to get a true/false value to pass to forceMask.

packages/rrweb-snapshot/src/snapshot.ts
Juice10: According to the documentation unmaskTextClass also works on conventional masking techniques. Maybe we need to check if a mask is applied first, before we check its unmask distance

Juice10:
@mdellanoce This PR is coming along really nicely. Since performance is such an important part of this it probably makes sense to re-run the benchmarks, and make some benchmarks of our own to understand the performance of this. 
Check out this PR, it should give you a handy guideline https://github.com/rrweb-io/rrweb/pull/903 specifically checkout packages/rrweb/test/benchmark/dom-mutation.test.ts it has the contents of these benchmarks

mdellanoce:
@Juice10 I added 2 new benchmarks for masking and unmasking.

Here are the results from master:

image
image
image
image

Note: the unmasking test on master is effectively the same as the masking test, since unmasking is unsupported there.

Here are the results from my branch:

image
image
image
image


(commits with changes to follow soon, currently in rebase hell)

mdellanoce:
@Juice10 ready for another look, i think

billyvg (commented):
@mdellanoce I meant to make these PR's to your fork, but I updated my local branches with master and it's causing bad diffs, so I'll just link the PRs from my fork -- if you update your branch with rrweb master, I can send out PRs

1) https://github.com/getsentry/rrweb/pull/97 -- This fixes tests not properly using `maskAllText`, this shows that unmasking does not work. I've also added `<textarea>` in `mask-text.html`, as it has a `textContent` prop that can conflict with its `value` attribute when masking. 

2) https://github.com/getsentry/rrweb/pull/98 - adds additional tests for dynamically added inputs and different configurations of maskAllInput. 

There are additional comments within these PRs that point out the failing/desired test results.

Let me know if you have any questions and if I can assist in any way!

mdellanoce:
@billyvg i synced branch with master, i'm looking into the unmasking issue you pointed out

packages/rrweb/test/utils.ts
```diff
@@ -597,6 +597,7 @@ export function generateRecordSnippet(options: recordOptions<eventWithTime>) {
     maskTextSelector: ${JSON.stringify(options.maskTextSelector)},
     maskAllInputs: ${options.maskAllInputs},
     maskInputOptions: ${JSON.stringify(options.maskAllInputs)},
+    maskInputFn: ${options.maskInputFn},
```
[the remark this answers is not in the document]
mdellanoce: fixed. also looks like the unmasking issue was the same (not passing unmaskTextSelector here), so I updated that as well. The test snapshots look correct to me now, but would definitely appreciate your eyes on them as well.

billyvg:
So here's a case with textareas that will cause a bug: https://github.com/pendo-io/rrweb/compare/md-mask-enhancements-1096..getsentry:rrweb:add-test-for-dynamically-added-inputs#diff-a414b2006f5f91cc65386e7241dbe659dac2ab113e09439483cc817373104a28R11120-R11145

test here: https://github.com/pendo-io/rrweb/compare/md-mask-enhancements-1096..getsentry:rrweb:add-test-for-dynamically-added-inputs#diff-185b8ff868973ccd68a879036d1e5f5abcd83ce0a908c833015b204222676214R650-R673

---

this also looks wrong: https://github.com/pendo-io/rrweb/compare/md-mask-enhancements-1096..getsentry:rrweb:add-test-for-dynamically-added-inputs#diff-a414b2006f5f91cc65386e7241dbe659dac2ab113e09439483cc817373104a28R15335-R15352

packages/rrweb/test/__snapshots__/integration.test.ts.snap
```diff
@@ -3391,6 +3391,349 @@ exports[`record integration tests can record node mutations 1`] = `
...
+                  {
+                    \\"type\\": 2,
+                    \\"tagName\\": \\"div\\",
+                    \\"attributes\\": {
+                      \\"class\\": \\"rr-unmask\\"
+                    },
+                    \\"childNodes\\": [
+                      {
+                        \\"type\\": 3,
+                        \\"textContent\\": \\"*******\\\\n        \\",
```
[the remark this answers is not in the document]
mdellanoce: i don't think so. one of the later commits makes unmaskTextClass empty by default, so since i didn't specify it in the test, this shouldn't be unmasked. I could remove the rr-unmask from the test file though, since I can see how that is confusing.

mdellanoce:
@billyvg 

1. okay, I see this, I must be missing some case for textarea, b/c it looks like it masked the value attribute, but not the textContent. Will fix that up.
2. This one is because maskAllInputs has precedence over the text masking settings. We decided to give the existing input masking options (maskAllInputs and maskInputOptions) highest priority here, so our customers don't end up accidentally unmasking something sensitive. Maybe I just need to update the docs to be more explicit about the masking precedence?

billyvg:
> @billyvg
> 
> 1. okay, I see this, I must be missing some case for textarea, b/c it looks like it masked the value attribute, but not the textContent. Will fix that up.
What I did on my fork was strip `textContent` completely from textarea, but my implementation of `maskAllText` was different because it was independent of `maskAllInputs`. So you could have maskAllText != maskAllInputs, and textareas could have only one of `value`/`textContent` masked, and the other unmasked. Not sure if this is the case with your impl of `maskAllText`

> 2. This one is because maskAllInputs has precedence over the text masking settings. We decided to give the existing input masking options (maskAllInputs and maskInputOptions) highest priority here, so our customers don't end up accidentally unmasking something sensitive. Maybe I just need to update the docs to be more explicit about the masking precedence?

Ah I see, that makes sense then, i was purely going from what I saw in the test snapshots (the snapshots make the tests hard to reason about). Documenting the masking precedence would definitely be helpful!

mdellanoce:
> but my implementation of maskAllText was different because it was independent of maskAllInputs. So you could have maskAllText != maskAllInputs, and textareas could have only one of value/textContent masked, and the other unmasked. Not sure if this is the case with your impl of maskAllText

Yeah, that's what is happening here too, as far as I can tell. The test sets maskAllInputs to true, which masks the value of the textarea, but the textContent is not masked because that is considered "not an input" and maskAllText is not set to true. Definitely seems incorrect.

mdellanoce:
@billyvg fixed the textarea masking. it can still be a little wonky with the default masking functions, b/c one (maskTextFn) excludes whitespace from masking, while the other (maskInputFn) masks every character regardless.

billyvg:
> @billyvg fixed the textarea masking. it can still be a little wonky with the default masking functions, b/c one (maskTextFn) excludes whitespace from masking, while the other (maskInputFn) masks every character regardless.

Should we have a default masking regex/function?

packages/rrweb/src/record/observer.ts
```diff
@@ -456,11 +462,21 @@ function initInputObserver({
...
+      maskAllText,
+    );
+
     if (type === 'radio' || type === 'checkbox') {
       isChecked = (target as HTMLInputElement).checked;
     } else if (
       maskInputOptions[tagName.toLowerCase() as keyof MaskInputOptions] ||
-      maskInputOptions[type as keyof MaskInputOptions]
+      maskInputOptions[type as keyof MaskInputOptions] ||
+      forceMask
```
billyvg, line 479: I don't think we can re-use `needMaskingText`/`forceMask` here because we could have an `unmaskTextSelector` that matches which means `forceMask` is false, but `maskInputOptions` is true.

`forceMask` needs to have 3 outcomes I suppose: 1) matches mask text class, 2) matches unmask, 3) does not match any masking OR unmasking

Only in the 3rd case should we fallback to checking `maskInputOptions`

mdellanoce: not sure i follow

> because we could have an unmaskTextSelector that matches which means forceMask is false, but maskInputOptions is true

in this case, it'd still be masked b/c maskInputOptions is true, and the input masking options have priority on inputs like we discussed above?

mdellanoce:
@Juice10 @YunFeng0817 something I've been playing with in a different branch:

https://github.com/pendo-io/rrweb/commit/6516f547e9b6e9d7a614305f5f87eea47f1547ce

I noticed when taking a snapshot or processing a large mutation, we'd be evaluating masking selectors for the same nodes over and over. This commit caches those results temporarily to avoid re-evaluating the selectors when the result for a given node is known already. In my testing, it seems to provide a minor improvement (~10ms faster for a 100ms snapshot). Thought it might provide a bigger boost, so that was a bit of a disappointment, though I've seen it be quite a bit faster in certain configurations.

Not sure my approach here is the best/cleanest, but figured I'd throw it out there in case performance concerns are still a sticking point on this PR.

eoghanmurray:
I haven't read this PR in detail, but one idea could be that `maskDistance` be passed down the call chain (and incremented at each level) like the boolean `needsMask` is passed down in #1349 — assuming that's merged

eoghanmurray:
So https://github.com/rrweb-io/rrweb/pull/1349 is merged now which I wrote before being aware of this PR; it fixes performance problems with the prior approach namely the constant looking back up the tree when recursing during full snapshot.
The approach from that PR should make moot the need for caching, as each node should only be checked once (with the exception of mutations, where you always need to look back up the tree after insertion)

I've tried to rebase this PR based on that, but the merge is non-trivial, but I might be able to do it given half a day.

Possibly it needs a new approach, maybe passing down counters for the 'mask/unmask distance' when recursing.

Another option as a first step would be to have the maskAllText + unmaskText be mutually exclusive from options which include the current 'maskText', in order to avoid any distance calculation.

(Also, it might be nice to first merge the `maskTextClass/Selector` options for simplicity, and/or combine all into a single 'maskTextOptions' object similar to MaskInputOptions)

mdellanoce:
@eoghanmurray i'll work on rebasing and modifying the approach

eoghanmurray:
If you wish to contribute the fix for #874 as a separate PR that might make review easier, we should be able to fasttrack that.

mdellanoce:
@eoghanmurray i rebased and updated this PR to only address #874. I removed all the new unmasking logic, since I can apply that through the maskText/InputFn callbacks now that they receive the element as a parameter.

I also added a check for `maskTextSelector == '*'` to return true faster from `needMaskingText` without needing to run matches or closest.

colingm:
@eoghanmurray it has been awhile since we commented on this. This PR has been cut down to just address #874 , could we get another look here?

PR #1098, opened 2023-01-18
Don't have requestAnimationFrame looping in background for Live Mode 

eoghanmurray:
Don't have requestAnimationFrame looping in background for Live Mode - instead restart the loop when a new action is added

### DIFF
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -718,14 +718,14 @@ export class Replayer {
           this.service.send('END');
           this.emitter.emit(ReplayerEvents.Finish);
         };
-        let finish_buffer = 50;  // allow for checking whether new events aren't just about to be loaded in
+        let finish_buffer = 50; // allow for checking whether new events aren't just about to be loaded in
         if (
           event.type === EventType.IncrementalSnapshot &&
           event.data.source === IncrementalSource.MouseMove &&
           event.data.positions.length
         ) {
           // extend finish event if the last event is a mouse move so that the timer isn't stopped by the service before checking the last event
-          finish_buffer += Math.max(0, -event.data.positions[0].timeOffset)
+          finish_buffer += Math.max(0, -event.data.positions[0].timeOffset);
         }
         setTimeout(finish, finish_buffer);
       }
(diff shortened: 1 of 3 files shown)


packages/rrweb/src/replay/timer.ts
```diff
@@ -10,64 +10,71 @@ export class Timer {
   public speed: number;
 
   private actions: actionWithDelay[];
-  private raf: number | null = null;
-  private liveMode: boolean;
+  private raf: number | boolean | null = null;
```
Mark-Fenng, line 13:
```suggestion
  private raf: number | true | null = null;
```
I think we can make this type more precise.

packages/rrweb/src/replay/index.ts
```diff
@@ -719,18 +718,16 @@ export class Replayer {
...
-          // defer finish event if the last event is a mouse move
-          setTimeout(() => {
-            finish();
-          }, Math.max(0, -event.data.positions[0].timeOffset + 50)); // Add 50 to make sure the timer would check the last mousemove event. Otherwise, the timer may be stopped by the service before checking the last event.
-        } else {
-          finish();
+          // extend finish event if the last event is a mouse move so that the timer isn't stopped by the service before checking the last event
+          finish_buffer += Math.max(0, -event.data.positions[0].timeOffset);
         }
+        setTimeout(finish, finish_buffer);
```
Mark-Fenng, line 730: With #1115, there will not be 'finish' event anymore in the liveMode. So IMO adding delay for all events won't be necessary.

Mark-Fenng (commented):
Most of code looks good to me.

packages/rrweb/src/replay/timer.ts
```diff
@@ -10,64 +10,71 @@ export class Timer {
...
+      this.raf = requestAnimationFrame(this.rafCheck.bind(this));
+    } else {
+      this.raf = true; // was active
+    }
   }
 
   public clear() {
     if (this.raf) {
-      cancelAnimationFrame(this.raf);
+      if (this.raf !== true) {
```
Juice10, line 75: `raf` can also be null
```suggestion
      if (typeof this.raf === "number") {
```

Juice10 (approved):

packages/rrweb/src/replay/index.ts
eoghanmurray: Was the removal of this setTimeout implemented in #1115?

PR #1104, opened 2023-01-31
[console-plugin] Feat: Record `unhandledrejection` event

jlalmes:
https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event

### DIFF
--- a/packages/rrweb/src/plugins/console/record/index.ts
+++ b/packages/rrweb/src/plugins/console/record/index.ts
@@ -133,7 +133,10 @@ function initLogObserver(
         if (window) window.removeEventListener('error', errorHandler);
       });
       const unhandledrejectionHandler = (event: PromiseRejectionEvent) => {
-        const error = event.reason as Error;
+        const error =
+          event.reason instanceof Error
+            ? event.reason
+            : new Error(`Uncaught (in promise) ${event.reason as string}`);
         const trace: string[] = ErrorStackParser.parse(
           error,
         ).map((stackFrame: StackFrame) => stackFrame.toString());
--- a/packages/rrweb/src/plugins/console/record/index.ts
+++ b/packages/rrweb/src/plugins/console/record/index.ts
@@ -132,6 +132,26 @@ function initLogObserver(
       cancelHandlers.push(() => {
         if (window) window.removeEventListener('error', errorHandler);
       });
+      const unhandledrejectionHandler = (event: PromiseRejectionEvent) => {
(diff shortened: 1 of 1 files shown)


packages/rrweb/src/plugins/console/record/index.ts
```diff
@@ -132,6 +132,26 @@ function initLogObserver(
       cancelHandlers.push(() => {
         if (window) window.removeEventListener('error', errorHandler);
       });
+      const unhandledrejectionHandler = (event: PromiseRejectionEvent) => {
+        const error = event.reason as Error;
```
Mark-Fenng, line 136: https://developer.mozilla.org/en-US/docs/Web/API/PromiseRejectionEvent/reason
image
According to the document, the `reason` is not always an Error object. It can even be `undefined` or `null`.
So maybe the reason object can be handled in a more proper way.

Mark-Fenng (changes requested):
Thanks for your great patch to console-plugin.

packages/rrweb/src/plugins/console/record/index.ts
jlalmes: Thanks @Mark-Fenng, how does the latest commit look?

Mark-Fenng (commented):
I made some changes in another branch https://github.com/rrweb-io/rrweb/commits/jlalmes-feat/console-plugin/add-unhandledrejection-handler. Could you please take a look at them?
[commit 1](https://github.com/rrweb-io/rrweb/commit/01165e81e507880b0679c5a08262b314546fab9e) :
change the window object to the passed in 'win'. I just noticed these problems in the old version. I'm trying to fix them in the PR by the way.

[commit 2](https://github.com/rrweb-io/rrweb/commit/1911f628b00143f463cb5b2d297a5ea0d0a24e7c)
improve serialized error messages

packages/rrweb/src/plugins/console/record/index.ts
Mark-Fenng: When I test the branch, I found that if the reason is an object, the final message would be like: 'Uncaught (in promise) [object Object]'.

jlalmes:
@Mark-Fenng both of these look good to me, I have merged your commits into this branch. 👍

Mark-Fenng (approved):

Mark-Fenng:
Thank you! I will merge this if no one reviews this by the end of this week.

PR #1033, opened 2022-10-21
Chore: Migrate build to vite

Juice10:
- [x] Move plugins to /packages/plugins/PLUGIN_NAME
- [x] Move testing of plugins to their own packages
- [x] Move packer to its own package
- [x] Get tests working again in all packages
- [x] Add umd bundling to rrweb: depends on https://github.com/vitejs/vite/pull/10609 (or use workaround by running multiple vite builds, one for each entry) https://github.com/vitejs/vite/discussions/1736#discussioncomment-4997467
- [x] Unify all tsconfig.json files by extending them from `/tsconfig.base.json`
  - [x] Fix any typescript code that fails due to unification
- [x] Update all `.md` files to reflect new file names and plugin locations
- [x] port bundling of rrweb-player to vite
- [x] fix errors from https://publint.dev/rrweb@2.0.0-alpha.4
- [x] setup `yarn dev` to build with --watch 
- [x] move `apply virtual style rules to node` tests to rrweb from rrdom
- [x] consider adding an `@rrweb/all` convenience package that bundles: record/replay/pack/unpack/plugins-console-record/plugins-console-replay
- [x] make sure repl and stream work
- [x] make sure benchmarks run
- [x] fix CI

Tests passing:
 - [x] rrdom/
 - [x] rrdom-nodejs/ ~`failing: needs to be updated to jest 29 like rrdom`~
 - [x] rrvideo/ ~`failing: needs to be updated to jest 29 like rrdom`~
 - [x] rrweb/ ~`failing`~
 - [x] ~rrweb-player/~ *no tests*
 - [x] rrweb-snapshot/ ~`2 failed tests "rrweb is not defined"`~
 - [x] @rrweb/packer
 - [x] ~@rrweb/types~ *no tests*
 - [x] ~web-extension/~ *no tests*
 - [x] ~plugins/rrweb-plugin-canvas-webrtc-record/~ *no tests*
 - [x] ~plugins/rrweb-plugin-canvas-webrtc-replay/~ *no tests*
 - [x] plugins/rrweb-plugin-console-record/ `failing locally some times; vite adds a console.debug message when included`
 - [x] ~plugins/rrweb-plugin-console-replay/~ *no tests*
 - [x] ~plugins/rrweb-plugin-sequential-id-record/~ *no tests*
 - [x] ~plugins/rrweb-plugin-sequential-id-replay/~ *no tests*

Build passing:
 - [x] rrdom/
 - [x] rrdom-nodejs/
[...]

PR #1105, opened 2023-01-31
[network-plugin] Feat: Capture network events

### LINKED ISSUE #552: XHR Capture
I'm sorry that I don't know typescript yet to more deeply contribute to the project.  In an effort to help with the groundwork to capture the XHR calls I've written the following gist:

https://gist.github.com/ucjonathan/9adc68ee74b331f6da5e0a13e1652dac

This is a vanilla JS piece of code that will collect the XHR request/response details into an object and then trigger an event that can be listened to.  I've mode provisions in the code to:

1) Allow for a custom function to perform redaction on the log record before further processing (masking of sensitive data that could be in the request/response.
2) Allow for a custom function to suppress a particular capture.  You don't really want to capture the XHR call you're making to store the events that RRWeb is collecting on your server.  That would create a nasty feedback loop.

I'm hopeful that this code could be used as the foundation to start adding that XHR capture support as a new event type similar to what was done with the console logging.

jlalmes:
Closes https://github.com/rrweb-io/rrweb/issues/552

This implementation records [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) and [`XMLHttpRequest`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) by patching their object & methods. We record document navigation using [`PerformanceNavigationTiming`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) and we use [`PerformanceResourceTiming`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) for recording everything else (script, img, link etc.) via [`PerformanceObserver`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver) API.

### Recording
```ts
type NetworkRequest = {
  url: string;
  method?: string;
  initiatorType: InitiatorType;
  status?: number;
  startTime: number;
  endTime: number;
  requestHeaders?: Headers;
  requestBody?: string | null;
  responseHeaders?: Headers;
  responseBody?: string | null;
};

rrweb.record({
  ...,
  plugins: [
    getRecordNetworkPlugin({
      initiatorTypes: ['fetch', 'xmlhttprequest'], // omit to record all types
      ignoreRequestFn: (request: NetworkRequest) => false
      recordHeaders: true, // defaults to false
      recordBody: true, // defaults to false
      recordInitialRequests: true, // defaults to false
    })
  ]
})
```

- `initiatorTypes` is an array specifying which request types to record
- `ignoreRequestFn` is a function that allow you to block recording an event for a request (you don't want to create a loop recording the request you send to store `rrweb` events on your server)
- `recordHeaders` will record an request/response headers for `fetch` and `xmlhttprequest` 
- `recordBody` will record an request/response bodies for `fetch` and `xmlhttprequest`
- `recordInitialRequests` will record an event for all requests prior to `rrweb.record()` being called

### Replaying
```ts
new rrweb.Replayer(events, {
  ...,
  plugins: [
    getReplayNetworkPlugin({
      onNetworkData: (data: { requests: NetworkRequest[], isInitial?: boolean }) => { // required
        // do something
      }, 
    })
  ]
})
```

- You **must** supply your own `onNetworkData` function.

### To-do
- Tests


jlalmes:
Just a thought, the `NetworkRequest` payload is quite large because the `PerformanceEntry` interface contains more information than we likely need. What are your thoughts on changing it to the following? 
```ts
type NetworkRequest = {
  requestUrl: string;
  requestMethod: string;
  requestInitiator: InitiatorType | 'navigation';
  requestHeaders?: Headers;
  requestBody?: Body;
  responseStatus: number;
  responseTimeMs: number;
  responseHeaders?: Headers;
  responseBody?: Body;
};
```

packages/rrweb/src/plugins/network/record/index.ts
```diff
@@ -0,0 +1,512 @@
...
+                networkRequest.responseHeaders![header] = value;
+              }
+            });
+          }
+          if (recordResponseBody) {
+            if (!xhr.response) {
+              networkRequest.responseBody = null;
+            } else {
+              try {
+                const objBody = JSON.parse(xhr.response as string) as object;
```
Mark-Fenng, line 306: It might not be safe to use the response as a string directly. The response can be other types, such as images, videos, etc. It's better to filter out these non-text responses as they can significantly increase the event size or leave the choice to the developers.

packages/rrweb/src/plugins/network/record/index.ts
```diff
@@ -0,0 +1,512 @@
...
+      before = win.performance.now();
+      networkRequest.responseStatus = res.status;
+      if (recordResponseHeaders) {
+        networkRequest.responseHeaders = {};
+        res.headers.forEach((value, header) => {
+          networkRequest.responseHeaders![header] = value;
+        });
+      }
+      if (recordResponseBody) {
+        const reqBody = await res.clone().text();
```
Mark-Fenng, line 420: Similarly, it's better to have a content type filter here.

packages/rrweb/src/plugins/network/record/index.ts
```diff
@@ -0,0 +1,512 @@
...
+type NetworkRequest = {
+  requestUrl: string;
+  requestMethod: string;
+  requestInitiator: InitiatorType;
+  requestHeaders?: Headers;
+  requestBody?: string | null;
+  responseStatus?: number;
+  responseDuration: number;
+  responseHeaders?: Headers;
+  responseBody?: string | null;
```
Mark-Fenng, line 89: I'm wondering whether it is possible to have a property that represents the starting time of requests. 
Since only with duration, it's hard to determine when the request happens. Specially, with the option `recordInitialRequests` switched on, it's harder to determine when the requests happened.

YunFeng0817 (commented):
This is a wonderful and exciting pull request, thank you very much. 
Most of code looks great. I'm also willing to add Chinese docs for it later.
It's better to have some test cases on the recorder side since this plugin is huge and importatnt.

packages/rrweb/src/plugins/network/record/index.ts
jlalmes: agreed 👍
```diff
- responseDuration: number;
+ startTime: number;
+ endTime: number;
```

jlalmes:
Thanks for your review @Mark-Fenng - I have just updated this branch with all of the changes you have suggested! I plan to add some tests shortly.

Out of interest do you know when we plan to release `2.0.0-alpha.5`?

packages/rrweb/src/plugins/network/record/index.ts
jlalmes: Sounds good, `recordBody` now has the following signature.

```ts
recordBody?: boolean | string[] | { 
  request: boolean | string[]; 
  response: boolean | string[] 
};
```

- `true` records everything
- `false` records nothing
- `string[]` records only requests/response bodies with matching `Content-Type` headers.

Example:
```ts
getRecordNetworkPlugin({
  recordBody: ['application/json'],
  ...,
})
```

Mark-Fenng: This thought is great but the structure of the option can be optimized IMO.
The current structure design is too complex for a single property `recordBody` and it's also hard to parse. For example, if I set `recordBody: ['application/json']`, the parsed `recordResponseBody` will always be true rather than a string array. You may have a try at it.
https://github.com/rrweb-io/rrweb/pull/1105/files#diff-7fb82041d48502b38e60b5d8e2901b6f9cb5ff213f8841e6f3b1f409c1a51ebaR229-R233
image


In my view, it's OK to have more optional properties rather than just one `recordBody`.

packages/rrweb/src/plugins/network/record/index.ts
```diff
@@ -0,0 +1,505 @@
...
+  url: string;
+  method?: string;
+  initiatorType: InitiatorType;
+  status?: number;
+  startTime: number;
+  endTime: number;
+  requestHeaders?: Headers;
+  requestBody?: unknown;
+  responseHeaders?: Headers;
+  responseBody?: unknown;
```
Mark-Fenng, line 81: We may use a more specific type definition here.

Mark-Fenng:
Thanks for applying all review suggestions. Although there are still a few small points that need to tweak, it's very close to the final version.

Here is a TODO list:

- [ ] improve the content type filter
- [ ] test cases
- [ ] Chinese docs (I will add them later)

I'm so sorry that, for the release of 2.0.0-alpha.5, I also don't have a specific date to tell. These days other team members are all busy with their work and some important pull requests for bug fixes are delayed.

packages/rrweb/src/plugins/network/record/index.ts
jlalmes: > `recordBody: ['application/json']`, the parsed `recordResponseBody` will always be `true`

> content type `"application/json; charset=UTF-8"` the current filter can't recognize

I have fixed the two bugs above, but have kept the signature of `recordBody` the same. I have added a function `shouldRecordBody` to simplify this complexity.

YunFeng0817 (dismissed):
With more tests, I think this pull request is good to go

Juice10:
This looks great! I agree with @Mark-Fenng, if we have some more tests this will be ready to merge. Also I'm happy to spend some time working on the 2.0.0-alpha.5 release soon

docs/recipes/network.md
```diff
@@ -0,0 +1,97 @@
...
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        events: [event],
+      }),
+    });
+  },
+  // customized record options
+  plugins: [
+    rrweb.getRecordConsolePlugin({
```
benjackwhite, line 37: Think this is a typo? 

```suggestion
    rrweb. getRecordNetworkPlugin({
```

benjackwhite:
@jlalmes need some help getting this across the finish line? This would be such a good improvement to the custom solution we have for network recording and things here look pretty solid, just needing some good tests it seems.

yaoyuanrong:
Is this plugin released?

yaoyuanrong:
I really need this feature。I think this is a very powerful feature

antunesgabriel (approved):

packages/rrweb/src/plugins/network/record/index.ts
```diff
@@ -0,0 +1,491 @@
...
+    };
+  }
+  const networkOptions = (options
+    ? Object.assign({}, defaultNetworkOptions, options)
+    : defaultNetworkOptions) as Required<NetworkRecordOptions>;
+
+  const cb: networkCallback = (data) => {
+    const requests = data.requests.filter(
+      (request) => !networkOptions.ignoreRequestFn(request),
+    );
```
benjackwhite, line 468: It would be great if this was both filtering and cleansing.

The most typical case for privacy controls is:
"I want to filter out requests to this place"
"I wan't to ensure certain headers aren't included"
"I want to filter out PII from the response bodies".

This would all doable with a slight modification of the function:
```suggestion
    const requests = data.requests.reduce(
      (list, request) => {
        const maskedRequest = networkOptions.maskRequestFn(request)
        
        return maskedRequest ? [...list, maskedRequest] : list
      }, []
    );
```

Which then allows the function to be something like

```ts
maskRequestFn: (request: NetworkRequest): NetworkRequest | undefined  => {
  if (request.url.contains("rrweb-collector-api.com") {
    return
 }
 delete request.requestHeaders["Authorization"]
 request.responseBody = maskTextFn(request.responseBody)
 return request
}

```

benjackwhite (changes requested):
Saw this is starting to get reviewed and wanted to throw in a thought regarding `ignoreRequestFn` to see if we can't make it more flexible, as well as more inline with the text and input style approach.

packages/rrweb/src/plugins/network/record/index.ts
benjackwhite: In theory this could be added in a follow up PR as well. Happy to do that if that is the decision.

packages/rrweb/src/plugins/network/record/index.ts
```diff
@@ -0,0 +1,491 @@
...
+  initiatorType: string,
+  url: string,
+  after?: number,
+  before?: number,
+  attempt = 0,
+): Promise<PerformanceResourceTiming> {
+  if (attempt > 10) {
+    throw new Error('Cannot find performance entry');
+  }
+  const urlPerformanceEntries = win.performance.getEntriesByName(
```
pauldambra, line 212: If I'm understanding correctly I think we'd need to make this optional... 

If people already have their own code (or another dependency) tracking performance entries they might be calling https://developer.mozilla.org/en-US/docs/Web/API/Performance/clearResourceTimings already so we'd have a race. 

Since this usage comes after the response if someone is clearing the buffer as soon as they're able to then I guess that this code would lose the race.

I think it's cheap enough to have the observer running no matter what that you could have `recordRequestPerformance` and return early here so the signature becomes `Promise<PerformanceRequestTiming | null>`

pauldambra: My mistake - 🤯 I've just read back through this and seen that the thrown error (and any other error here) are captured and swallowed where the returned promise is handled.

My preference would be to return null so that the flow is more clear but at least the error here doesn't bubble out

packages/rrweb/src/plugins/network/record/index.ts
```diff
@@ -0,0 +1,491 @@
...
+      requests: initialPerformanceEntries.map((entry) => ({
+        url: entry.name,
+        initiatorType: entry.initiatorType as InitiatorType,
+        status: 'responseStatus' in entry ? entry.responseStatus : undefined,
+        startTime: Math.round(entry.startTime),
+        endTime: Math.round(entry.responseEnd),
+      })),
+      isInitial: true,
+    });
+  }
```
pauldambra, line 141: Was playing with this...

If I understand correctly...

This entire block only exists so `isInitial: true` can be added to the callback payload. If we weren't adding that property then when creating the observer immediately below we could observe the entry types individually with `buffer: options.recordInitialRequests`

I can't think of the use-case for knowing a request was "initial" when playing back. (tbh I can't think why I'd not want to capture "initial" requests) But in the interests of starting with the smallest API because it's easier to add things. I'd be tempted to say we shouldn't record `initial` on the payload.

packages/rrweb/src/plugins/network/record/index.ts
```diff
@@ -0,0 +1,491 @@
...
+  method?: string;
+  initiatorType: InitiatorType;
+  status?: number;
+  startTime: number;
+  endTime: number;
+  requestHeaders?: Headers;
+  requestBody?: Body;
+  responseHeaders?: Headers;
+  responseBody?: Body;
+};
```
pauldambra, line 92: I was experimenting with this again...

We already capture Network timing and the `PerformanceEntry` spec uses `name` where this `NetworkRequest` timing uses `url`

It is more natural for the typing here to be something like

```
export type CapturedNetworkRequest = Omit<PerformanceEntry, 'toJSON'> & {
    method?: string
    initiatorType?: InitiatorType
    status?: number
    startTime?: number
    endTime?: number
    requestHeaders?: Headers
    requestBody?: Body
    responseHeaders?: Headers
    responseBody?: Body
    // was this captured before fetch/xhr could have been wrapped
    isInitial?: boolean
}
```

That way you can extend this to capture `PerformanceEntry`s and/or headers and/or body and only use one type

pathik-appachhi:
Hello,

I have recently started using rrrweb-io via the [cdn](https://cdn.jsdelivr.net/npm/rrweb@2.0.0-alpha.11/) . It would be very helpful to know when this plugin will be available on the CDN link.

ShayMalchi:
@jlalmes @YunFeng0817 Seems like this PR was approved long ago, can you please share what is the current blocker for merging? Many people in this thread are interested in it, and seems like most suggestions could be handled in followup PRs. Thank you!

pauldambra:
I'll just re-up this comment https://github.com/rrweb-io/rrweb/pull/1105#discussion_r1402360817

It's easier for me if we have `name` not `URL`. Since the performance observer uses `name` and emits more than just things with URLs.

sharan-ravi:
@jlalmes @pauldambra 
Thank you so much for this!

I was running into an issue with some of my requests not being recorded with this plugin, after digging deeper, turns out in case of certain requests, if the server doesn't return a [Timing-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Timing-Allow-Origin) response header, we wouldn't be able to track the performance metrics. Given this plugin throws an error when performance metrics are not available, these requests were not being recorded. 

While the performance metrics are not available, Other data on these requests are still available, so in my case I had to do something like this to still record these requests. 

```
getRequestPerformanceEntry(win, 'fetch', req.url, after, before)
                    .then((entry) => {
                        if (_isNull(entry)) {
                            const requests = prepareRequestWithoutPerformance(req, networkRequest)
                            cb({ requests })
                        } else {
                            const requests = prepareRequest(entry, req.method, res?.status, networkRequest)
                            cb({ requests })
                        } })

function prepareRequestWithoutPerformance(
    req: Request,
    networkRequest: Partial<CapturedNetworkRequest>,
): CapturedNetworkRequest[] {

    const reqJson = {
        url: req.url,
        method: req.method,
        requestHeaders: networkRequest.requestHeaders,
        requestBody: networkRequest.requestBody,
        responseHeaders: networkRequest.responseHeaders,
        responseBody: networkRequest.responseBody,
    }

    return [reqJson]
}
```

Is this expected?

Juice10:
> My plan is to let it settle in our product for a week or two. And then I'll open a PR based on this one that would let us inject the things we'd varied - which might also be things other people would want to vary...
> 
> It won't be massively different - and of course the version that works for us might not be preferable generally
@pauldambra would you still be interested to open a PR based on the version you have running in production?

FYI: https://github.com/rrweb-io/rrweb/pull/1033 just got finished which moves the all of the plugins to the `packages/plugins/...` directory. Might be worth considering creating the PR based on that as well

akhsri:
Hi @jlalmes , When can we see this in rrweb package? it would be very powerful and helpful feature!

pauldambra:
@Juice10 definitely... our implementation has been pretty stable for a while now... it's just time that's stopping me 🙈 

I think the "tricky" thing will be separating our config/mapping needs out of the plugin but that's pretty mechanical non-ground-breaking stuff

Marlonzao:
Hello, is there a guide on how to download this plugin?

Marlonzao:
I have figuret it, you have to clone this repo https://github.com/jlalmes/rrweb/tree/feat/network-plugin then ckeckout to the branch "feat/network-plugin" and run "yarn && yarn run build:all" the files needed will be in the directory packages/rrweb/dist

akshanshkmr:
Hi,
Is there any plan to merge this?

crutch12:
@pauldambra may we somehow merge it and publish new packages to npm? It would be really helpful 🙏

crutch12:
I've updated the code and opened a new PR:
https://github.com/rrweb-io/rrweb/pull/1689

packages/rrweb/src/plugins/network/record/index.ts
crutch12: Fixed here: https://github.com/rrweb-io/rrweb/pull/1689

packages/rrweb/src/plugins/network/record/index.ts
crutch12: Fixed here: https://github.com/rrweb-io/rrweb/pull/1689

packages/rrweb/src/plugins/network/record/index.ts
crutch12: Fixed here: https://github.com/rrweb-io/rrweb/pull/1689

Juice10:
Thanks for contributing this @jlalmes! We just got it merged via https://github.com/rrweb-io/rrweb/pull/1689, thanks again for taking the time to do this and apologies that it took so long for us to merge it. Big thanks to everyone here that contributed as well!

PR #1106, opened 2023-02-03
fix: Ensure CSS support is checked more robustly

mydea:
Noticed this while debugging tests in sentry-javascript. it seems jsdom has a weird support for `CSSMediaRule`, where it is defined but if you look at `CSSMediaRule.prototype` it is:

```
CSSRule {
        parentRule: null,
        parentStyleSheet: null,
        constructor: [Function: CSSMediaRule],
        type: 4
      }
```

I guess this could also be the case for other "weird" browsers or engines. I think it's safer to make sure everything we need is actually there at this point.

ref https://github.com/getsentry/rrweb/pull/42

### DIFF
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -61,17 +61,14 @@ function isCSSStyleSheetMonkeyPatchable(
     | 'CSSSupportsRule'
     | 'CSSConditionRule',
 ): boolean {
-  /* eslint-disable @typescript-eslint/ban-ts-comment,@typescript-eslint/unbound-method */
   return Boolean(
     typeof window[prop] !== 'undefined' &&
       // Note: Generally, this check _shouldn't_ be necessary
       // However, in some scenarios (e.g. jsdom) this can sometimes fail, so we check for it here
       window[prop].prototype &&
-      // @ts-ignore ensure it is actually set
-      window[prop].prototype.insertRule &&
-      window[prop].prototype.deleteRule,
+      'insertRule' in window[prop].prototype &&
+      'deleteRule' in window[prop].prototype,
   );
-  /* eslint-enable @typescript-eslint/ban-ts-comment,@typescript-eslint/unbound-method */
 }
 
 const isCSSGroupingRuleSupported = isCSSStyleSheetMonkeyPatchable(
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -54,10 +54,34 @@ type WindowWithAngularZone = IWindow & {
 export const mutationBuffers: MutationBuffer[] = [];
 export const processedNodeManager = new ProcessedNodeManager();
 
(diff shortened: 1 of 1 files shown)


packages/rrweb/src/record/observer.ts
```diff
@@ -54,10 +54,36 @@ type WindowWithAngularZone = IWindow & {
...
+  return Boolean(
+    typeof window[prop] !== 'undefined' &&
+      // Note: Generally, this check _shouldn't_ be necessary
+      // However, in some scenarios (e.g. jsdom) this can sometimes fail, so we check for it here
+      window[prop].prototype &&
+      // @ts-ignore ensure it is actually set
+      window[prop].prototype.insertRule &&
+      window[prop].prototype.deleteRule,
+  );
+  /* eslint-enable @typescript-eslint/ban-ts-comment,@typescript-eslint/unbound-method */
```
Juice10, line 74: `ts-ignore` is a little aggressive, this would be a much cleaner way of getting the same result:
```suggestion
  return Boolean(
    typeof window[prop] !== 'undefined' &&
      // Note: Generally, this check _shouldn't_ be necessary
      // However, in some scenarios (e.g. jsdom) this can sometimes fail, so we check for it here
      window[prop].prototype &&
      "insertRule" in window[prop].prototype &&
      "deleteRule" in window[prop].prototype,
  );
```

Juice10:
@mydea Thanks for submitting this PR, I had no idea rrweb was being used to record virtual pages with JSDOM, very interesting use case!
I added a [small suggestion](https://github.com/rrweb-io/rrweb/pull/1106#discussion_r1095704909), if you go ahead and merge that suggestion, I'd be happy to approve this PR and get it merged.

mydea:
> @mydea Thanks for submitting this PR, I had no idea rrweb was being used to record virtual pages with JSDOM, very interesting use case! I added a [small suggestion](https://github.com/rrweb-io/rrweb/pull/1106#discussion_r1095704909), if you go ahead and merge that suggestion, I'd be happy to approve this PR and get it merged.

Yeah, I only really found it due to tests using jsdom. But it showed that apparently some older CSS spec had this different, which may also happen in other browsers I guess - see: https://github.com/NV/CSSOM/issues/105

I commited your suggestion 👍 thanks!

mydea:
Oops sorry accidentally deleted this branch 😬 restored it and reopened the PR!

Mark-Fenng (approved):

Mark-Fenng:
@Juice10  Is is OK to merge this one?

mydea:
OK, actually we ran into a problem with this, so I updated this PR to ensure we do not access `window` in module scope.

Now, this is only checked on-demand. I also split this into to methods, one to check when we only want to access this, and one when we want to monkey patch this.

Juice10 (approved):

Juice10:
@mydea thanks for the refactor and the bugfix, I’ve merged the PR

PR #1107, opened 2023-02-03
feat: Allow to pass `errorHandler` as record option

mydea:
We've been playing around quite a lot with ways to handle errors that originate somewhere from within rrwebs deep callback stack. This PR adds an option to add a `errorHandler` option to `record`, which can be used to either swallow errors or handle them in different ways:

```js
record({
  errorHandler: (error) => {
    // return `true` to swallow the error and stop propagating
    // Or log the error somewhere, etc.
  }
});
```

I tried to wrap all places that are called as (async) callbacks somewhere. I _may_ have missed a place, but IMHO this should give a pretty decent coverage.

Note that in order to do that, I had to rewrite the monkey patching we are doing on e.g. `CSSStyleSheet` to use a proxy. Otherwise, retaining the `this` context correctly turned out quite tricky. IMHO this is a nicer approach anyhow, and all supported browsers (> IE11) support proxies, so should be OK I think.

### DIFF
--- a/.changeset/pretty-plums-rescue.md
+++ b/.changeset/pretty-plums-rescue.md
@@ -1,0 +1,5 @@
+---
+'rrweb': minor
+---
+
+feat: Allow to pass `errorHandler` as record option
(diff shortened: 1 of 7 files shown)


packages/rrweb/src/record/errorHandler.ts
```diff
@@ -0,0 +1,37 @@
+import type { ErrorHandler } from '../types';
```
Juice10, line 1: We’re using kebab-case for file names in rrweb. Could you rename this file to `error-handler.ts`?

mydea: ah, sure thing!

YunFeng0817 (approved):
I feel like this is a good mechanism. Using "Proxy" constructor will make the browser compatibility worse. Otherwise, it's a good way to do patching things.
Could you please solve the merge conflicts and add a change log for this one?

mydea:
I resolved the merge commits and added a changeset!

mydea:
> Sorry, one more thing before merging this. Could you also add the description of this config in the rrweb doc?

Ah sure, of course! I added this to the guide doc as well.

YunFeng0817:
Yes, I resolved it in another PR.

PR #1109, opened 2023-02-05
add option to record on DOMContentLoaded event

### LINKED ISSUE #1090: [Bug]: Recorder initializes very late in page load lifecycle
### Preflight Checklist

rrweb

### Version

v1.1.3

### Expected Behavior

I am using RRWeb in a Chrome extension to record activity on a website. 
I inject RRWeb library in the webpage on `document_start` and call the `record()` function as soon as possible, so recording may start earliest in page load lifecycle and should not wait for complete page to load (like images). To start capturing DOM state, it should only wait for `DOMContentLoaded` event.

Along with DOM state, it should also start capturing console logs as early as possible.

### Actual Behavior

The RRWeb recorder seems to initialize only when window `load` event is fired which means the webpage has loaded completely including resources like images.

I was looking at the [source code](https://github.com/rrweb-io/rrweb/blob/master/packages/rrweb/src/record/index.ts#L578), according to which if the webpage is still in `loading` state, that is, `document.readyState="loading"`, it would wait for window `load` event which is fired after all resources have been loaded.
In my case, the whole website loads except for few images which are delayed for almost a minute and recording could not start until then.
IMHO, it should rather wait for `DOMContentLoaded` event and start observing DOM as soon as it is ready.

### Steps to Reproduce

Covered above.

### Testcase Gist URL

_No response_

### Additional Information

_No response_

Mark-Fenng:
Fixes #1090

- Add an option 'recordAfter' which can customize the initialization time of the recorder.
- Add doc for the option
- Add an integration test case

Helpful info: 
1. When `DOMContentLoaded` is emitted, the readyState of the document is `interactive`. 
2. When `load` is emitted, the readyState of the document is 'complete'.


jlalmes (approved):
🙌

Juice10 (approved):

PR #1115, opened 2023-02-09
[fix] Don't trigger Finish event when in liveMode

Juice10:
ReplayerEvents.Finish is used to set the rrweb-player to pause. This shouldn't trigger as long as live mode is enabled as new events may still come in


Mark-Fenng (approved):
You finally come back, hero!

PR #1119, opened 2023-02-09
Automate NPM package releases

Juice10:
Use [changesets](https://github.com/changesets/changesets) to automate all npm package releases.

Todo before of after merge
 - [x] Add NPM authentication token as a secret to the repository with the name `NPM_TOKEN`
 - [x] Add [Changesets GitHub bot](https://github.com/apps/changeset-bot) to this repository

 
_Edit: for more info on how this works see https://github.com/rrweb-io/rrweb/pull/1119#discussion_r1103520338 or https://github.com/changesets/changesets_


.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+          node-version: lts/*
+
+      - name: Install Dependencies
+        run: yarn
+
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          publish: yarn run release
```
Mark-Fenng, line 30: I just want to make sure this command will publish the package directly or create a pull request first?
I feel that it's safer to create a pull request first rather than directly publish packages.

Mark-Fenng:
Changesets will really help us publish packages much faster. Currently, our publishing period is much longer than the average of npm packages. I believe this tool will change the situation.

.github/workflows/release.yml
Juice10: It will create a PR first, once that PR is accepted it will publish the packages on NPM. Also if a PR already exists and more code is added to master the release command will update the PR with the new changes.

Mark-Fenng: I saw Matt's youtube video and I found that he only build the package here rather than publishing them. So I'm guessing ChangeSets will execute changesets publish for us rather than we manually do it?

Juice10: I tried that myself on https://GitHub.com/Juice10/rrweb and it didn’t work. Also looking at the changesets documentation and some projects using changesets this seems like the way to go. Looks like a mistake in Matt’s video

Mark-Fenng: Thanks for the explanation!

Mark-Fenng (approved):

Yuyz0112:
I think I've done the TODOs, feel free to ping me on slack or here if there is something missing.

packages/rrdom-nodejs/src/document-nodejs.ts
```diff
@@ -36,15 +36,16 @@ export class RRWindow {
 
 export class RRDocument
   extends BaseRRDocumentImpl(RRNode)
-  implements IRRDocument {
+  implements IRRDocument
+{
```
eoghanmurray, line 40: These are formatting changes unrelated to changesets?

PR #1122, opened 2023-02-12
Chore: add change log and update readme

YunFeng0817:
- Added change logs since 2.0.0-alpha.4
- Update outdated info in README.md


Juice10 (approved):
Thanks for adding this Yun!

PR #1128, opened 2023-02-13
fix: Exclude scripts loaded with `rel=modulepreload` from snapshots

billyvg:
We were only excluding `<link rel="preload" as="script" />` but we should include `rel="modulepreload"` as well.


Juice10 (approved):
Thanks for contributing this @billyvg

PR #1129, opened 2023-02-13
Record pointerType on clicks

eoghanmurray:
- could be useful for displaying e.g. a circle rather than a point during replay

touchscreen laptops can have both mouse and touch within the same session

### DIFF
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -240,13 +240,14 @@ function initMouseInteractionObserver({
       let pointerType: PointerTypes | null = null;
       let e = event;
       if ('pointerType' in e) {
-        Object.keys(PointerTypes)
-          .forEach((pointerKey: keyof typeof PointerKeys) => {
+        Object.keys(PointerTypes).forEach(
+          (pointerKey: keyof typeof PointerKeys) => {
             if ((e as PointerEvent).pointerType === pointerKey.toLowerCase()) {
               pointerType = PointerTypes[pointerKey];
               return;
             }
-          });
+          },
+        );
         if (pointerType === PointerTypes.Touch) {
           if (MouseInteractions[eventKey] === MouseInteractions.MouseDown) {
             // we are actually listening on 'pointerdown'
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -9,7 +9,7 @@ import {
   getWindowHeight,
   getWindowWidth,
   isBlocked,
-  isTouchEvent,
+  legacy_isTouchEvent,
   patch,
   StyleSheetMirror,
 } from '../utils';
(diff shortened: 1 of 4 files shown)


packages/rrweb/src/record/observer.ts
```diff
@@ -239,6 +239,7 @@ function initMouseInteractionObserver({
         id,
         x: clientX,
         y: clientY,
+        pointerType: e.pointerType, // touch / pen / mouse
```
Juice10, line 242: This looks great Eoghan, I would suggest a slight refactor since line 236 is already doing some destructuring

```ts
      const { clientX, clientY, pointerType } = e;
      mouseInteractionCb({
        type: MouseInteractions[eventKey],
        id,
        x: clientX,
        y: clientY,
        pointerType,  // touch / pen / mouse
      });
```

eoghanmurray: I think it needs some more work as `pointerType` is not always available.

Juice10:
You have some typescript typing that needs changing, and then a bunch of tests will need to be updated. But this is a welcome addition to rrweb IMO!

packages/rrweb/src/record/observer.ts
eoghanmurray: This is fixed now in latest force push

Juice10 (approved):

eoghanmurray:
@Juice10 I'm deploying the changes to a 'canary' test environment so feel free to merge in a few weeks if I haven't done so by then (will be away over Easter hols)

eoghanmurray:
@Juice10 any idea why `"pointerType": "mouse"` doesn't show up in regards to the `can record clicks` test?
 - I know the x/y values are stripped out by test/utils.ts 
 - in the live-stream view, the pointerType shows up — is livestream using the same puppeteer set up as testing?

eoghanmurray:
> @Juice10 any idea why `"pointerType": "mouse"` doesn't show up in regards to the `can record clicks` test?
> 
>     * I know the x/y values are stripped out by test/utils.ts
> 
>     * in the live-stream view, the pointerType shows up — is livestream using the same puppeteer set up as testing?

I guess it could be because our version of Puppeteer doesn't support PointerType?

Juice10:
@eoghanmurray the setup is mostly the same. Big difference is that click events in puppeteer aren't 100% the same as "real" click events.

From the puppeteer docs:
> Note: The mouse events trigger synthetic MouseEvents. This means that it does not fully replicate the functionality of what a normal user would be able to do with their mouse.
> 
> For example, dragging and selecting text is not possible using page.mouse. Instead, you can use the `DocumentOrShadowRoot.getSelection()` functionality implemented in the platform.

https://pptr.dev/api/puppeteer.mouse

Juice10:
@eoghanmurray would you like to merge this one or wait till you get the tests dialed in?

eoghanmurray:
I've gone ahead and merged; if we change up from Puppeteer in future it should be discoverable what the correct `pointerType` should be for the results of tests.

PR #1130, opened 2023-02-14
Fix relative path detection in stylesheet URLs

Equlnox:
There can URLs with protocols other than HTTP / HTTPs / FTP

Example:
If you have loom installed in your browser, it adds some stylesheets that refer to `chrome-extension` protocol

image

For such URLs, rrweb is pre-pending the webpage origin in the CSS.

This PR is meant to make the relative URL detection more generic.
The regex I have added has been taken from stack overflow - https://stackoverflow.com/a/19709846
It checks for the existence of a protocol in the URL.

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -72,7 +72,7 @@ let canvasCtx: CanvasRenderingContext2D | null;
 
 const URL_IN_CSS_REF = /url\((?:(')([^']*)'|(")(.*?)"|([^)]*))\)/gm;
 const URL_PROTOCOL_MATCH = /^(?:[a-z+]+:)?\/\//i;
-const URL_WWW_MATCH = /^www\..*/i
+const URL_WWW_MATCH = /^www\..*/i;
 const DATA_URI = /^(data:)([^,]*),(.*)/i;
 export function absoluteToStylesheet(
   cssText: string | null,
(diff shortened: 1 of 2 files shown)


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -92,7 +93,7 @@ export function absoluteToStylesheet(
       if (!filePath) {
         return origin;
       }
-      if (!RELATIVE_PATH.test(filePath)) {
+      if (URL_PROTOCOL_MATCH.test(filePath) || URL_WWW_MATCH.test(filePath) {
```
Juice10, line 96: You have a syntax error here
```suggestion
      if (URL_PROTOCOL_MATCH.test(filePath) || URL_WWW_MATCH.test(filePath)) {
```

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -71,7 +71,8 @@ let canvasService: HTMLCanvasElement | null;
 let canvasCtx: CanvasRenderingContext2D | null;
 
 const URL_IN_CSS_REF = /url\((?:(')([^']*)'|(")(.*?)"|([^)]*))\)/gm;
-const RELATIVE_PATH = /^(?!www\.|(?:http|ftp)s?:\/\/|[A-Za-z]:\\|\/\/|#).*/;
+const URL_PROTOCOL_MATCH = /^(?:[a-z+]+:)?\/\//i;
+const URL_WWW_MATCH = /^www\..*/i
```
YunFeng0817, line 75:
```suggestion
const URL_WWW_MATCH = /^www\..*/i;
```

YunFeng0817 (approved):
Thanks for the contribution.

PR #1132, opened 2023-02-14
some minor changes for releasing the extension and revert failed publish PR #1123

YunFeng0817:
1. The vite-plugin-zip don't support enough option to config.
2. Skip adding recorder to same-origin iframes.
3. revert failed PR #1123


Juice10 (approved):
Looks good!

PR #1133, opened 2023-02-14
Replay fix for Firefox — add <head> and <body> separately

eoghanmurray:
This PR solves a problem in Firefox where css transitions are incorrectly being applied upon rebuild.  Presumably FF doesn't finished parsing the styles in time, and applies e.g. a default margin:0 to elements which have a non-zero margin set in CSS, along with a transition on them.

I've included a test file which can't be automated at the moment as we use Chrome for repl and live-stream.

To recreate, you'd need to make a `repl` recording against the transition.html file, save the events to html and then open the file in Firefox (without this PR applied).  You'll see the red box very slowly move across the screen, which is incorrect as the styles never had a transition from margin-left:0 to margin-left:400px — it was at the latter from initial page load.


YunFeng0817:
Repro: open this link in a firefox browser.
https://rrwebdebug.com/play/index.html?url=http%3A%2F%2Fjsonblob.com%2F1075033034780327936&version=1.0.0-alpha.4&play=on

eoghanmurray:
I've no idea _how_ this fixes the problem.

I've created a bug report for Firefox here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1816672

eoghanmurray:
@YunFeng0817 wondering whether we need to also apply this fix when rebuilding a diff via rrdom, but I can't pinpoint the bit of code in rrdom that applies the diff... do you have any pointers?

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -437,6 +437,27 @@ export function buildNodeWithSN(
...
+          // margin set in CSS, along with a transition on them
+          const bodyP = bodys[0].parentElement;
+          const tmpBody = bodyP.removeChild(bodys[0]);
+          // append <head> and <style>s
+          node.appendChild(childNode);
+          // now append <body>
+          bodyP.appendChild(tmpBody);
+        } else {
+          node.appendChild(childNode);
+        }
```
YunFeng0817, line 460:
```suggestion
      } else if (
        n.type === NodeType.Document &&
        childN.type == NodeType.Element
      ) {
        const htmlElement = childNode as HTMLElement;
        let body: HTMLBodyElement | null = null;
        htmlElement.childNodes.forEach((child) => {
          if (child.nodeName === 'BODY') body = child as HTMLBodyElement;
        });
        if (body) {
          // this branch solves a problem in Firefox where css transitions are incorrectly
          // being applied upon rebuild.  Presumably FF doesn't finished parsing the styles
          // in time, and applies e.g. a default margin:0 to elements which have a non-zero
          // margin set in CSS, along with a transition on them
          htmlElement.removeChild(body);
          // append <head> and <style>s
          node.appendChild(childNode);
          // now append <body>
          htmlElement.append(body);
        } else {
          node.appendChild(childNode);
        }
```

YunFeng0817 (commented):
I rewrite your code to make it fit the rrdom and also make it easier to understand.  
It will be better to add a change log for this.

YunFeng0817 (approved):

PR #1136, opened 2023-02-15
fix: Recursive logging bug with console recording

benjackwhite:
Originally reported [here](https://github.com/PostHog/posthog-js/issues/517)

## Problem
Thanks to some debugging with a customer, we found an interesting "edge" case that can trigger an infinite loop via console logs. The context is this was a Vue.js app that was logging some warnings which contained objects within that warning with Proxied methods that triggered logs themselves.

The situation is like this (**WARNING**: If you run this with console log recording enabled your browser will freeze).

```js
let target = {
  foo: "bar"
};

const handler = {
  get(target, prop, receiver) {
    if (prop === "foo") {
      console.warn("foo was accessed. This shouldn't be done... Context:", target)
    }
    return Reflect.get(...arguments);
  },
};

const proxy = new Proxy(target, handler);
target.proxy = proxy

console.log(proxy); 
```

1. We log the `proxy` object
2. The console log is intercepted by our code
3. It creates a "copy" of the [object via stringifying](https://github.com/rrweb-io/rrweb/blob/master/packages/rrweb/src/plugins/console/record/index.ts#L166)
4. This triggers the proxied `get` call which itself is logging a warning
5. ... which triggers Step 2 and now we have a loop 🙈 

## Solution

* Flag when we enter the block to stringify and callback the code. If any console logs occur in that time, we don't capture them to avoid infinite loops

### DIFF
--- a/.changeset/violet-melons-itch.md
+++ b/.changeset/violet-melons-itch.md
@@ -1,0 +1,5 @@
+---
+"rrweb": patch
+---
+
+fix: Recursive logging bug with console recording
--- a/.changeset/violet-melons-itch.md
+++ b/.changeset/violet-melons-itch.md
@@ -1,5 +1,5 @@
 ---
-"rrweb": patch
+'rrweb': patch
 ---
 
 fix: Recursive logging bug with console recording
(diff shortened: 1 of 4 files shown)


YunFeng0817 (approved):
This a very interesting edge case. Thanks for your great description and fix.

YunFeng0817:
Could you please help us add the change log? Seems like I  don't have the permission to add it on your branch.

YunFeng0817:
@benjackwhite Yes, you did. I will merge it once CI passes.

PR #1139, opened 2023-02-16
Fix: Switch to real dom before rebuilding fullsnapshot

YunFeng0817:
This pull request makes rrdom optimization robust.

1. Normally rebuilding full snapshots should not be under a virtual dom environment.
But if the order of data events has some issues, it might be possible.
Adding a check to avoid any potential issues.

2. Fix the bug: If an RRNode appends the same child twice, the child nodes will become an infinite linked list. This is a bug imported in #1127

### DIFF
--- a/.changeset/eight-terms-hunt.md
+++ b/.changeset/eight-terms-hunt.md
@@ -1,0 +1,5 @@
+---
+'rrweb': patch
+---
+
+Fix: Switch from virtual dom to real dom before rebuilding fullsnapshot
--- a/.changeset/real-masks-explode.md
+++ b/.changeset/real-masks-explode.md
@@ -1,0 +1,5 @@
+---
+'rrdom': patch
+---
+
+Fix: If RRNode appends a single child twice, children of the node will become an infinite link list.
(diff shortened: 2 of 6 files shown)


Juice10:
By setting usingVirtualDom to false, aren’t you essentially turning off all rrdom/virtual dom optimizations whenever you hit a full snapshot?

YunFeng0817:
> By setting usingVirtualDom to false, aren’t you essentially turning off all rrdom/virtual dom optimizations whenever you hit a full snapshot?

Yes, because if a new fullsnapshot is rebuilt, the status of the existing virtual dom doesn't mean anything as it represents the previous snapshot. `usingVirtualDom` is an internal switch that determines the current dom environment.  It's different from the replayer's configuration `useVirtualDOm`.

eoghanmurray:
For future work: I feel like the `usingVirtualDom` global is dangerous; I'd prefer that this variable be passed down the call chain to each function rather than relying on global state.  I've come across a situation in last few weeks where after an error, more errors were generated because it's state wasn't reset.

YunFeng0817:
Your suggestion can also help. I think if this pull request is merged, there are not many serious bugs left.

Juice10 (approved):

PR #1140, opened 2023-02-16
Render full snapshot on scrub to zero

eoghanmurray:


Juice10:
Tests are failing, maybe you could add a test for this use case if the current failing tests don't cover it. Apart from that this looks good and can be approved/merged

eoghanmurray:
Tests are failing as this PR makes things more complicated!
e.g. 'will start actions when play' now has 2 less actions queued up as it has already 'sync executed' the first action which is a DomContentLoaded as well as a subsequent FullSnapshot.
Not sure whether/how to fix this up, I could just subtract 2 from the number of events expected?

PR #1145, opened 2023-02-17
Extend the suppression of warnings to take into account anscestors

eoghanmurray:
Take account that a prior removal may not have been against the immediate parent of a subsequent removal, but rather some anscestor

### DIFF
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -1361,14 +1361,20 @@ export class Replayer {
     const mirror = this.usingVirtualDom ? this.virtualDom.mirror : this.mirror;
     type TNode = typeof mirror extends Mirror ? Node : RRNode;
 
+    d.removes = d.removes.filter((mutation) => {
+      // warn of absence from mirror before we start applying each removal
+      // as earlier removals could remove a tree that includes a later removal
+      if (!mirror.getNode(mutation.id)) {
+        this.warnNodeNotFound(d, mutation.id);
+        return false;
+      }
+      return true;
+    });
     d.removes.forEach((mutation) => {
       const target = mirror.getNode(mutation.id);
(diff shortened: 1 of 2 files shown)


packages/rrweb/src/replay/index.ts
```diff
@@ -1361,14 +1361,20 @@ export class Replayer {
...
 
+    d.removes = d.removes.filter((mutation) => {
+      // warn of absence from mirror before we start applying each removal
+      // as earlier removals could remove a tree that includes a later removal
+      if (!mirror.getNode(mutation.id)) {
+        this.warnNodeNotFound(d, mutation.id);
+        return false;
+      }
+      return true;
+    });
```
YunFeng0817, line 1372:
```suggestion
```

packages/rrweb/src/replay/index.ts
```diff
@@ -1361,14 +1361,20 @@ export class Replayer {
...
     d.removes.forEach((mutation) => {
       const target = mirror.getNode(mutation.id);
       if (!target) {
-        if (d.removes.find((r) => r.id === mutation.parentId)) {
-          // no need to warn, parent was already removed
-          return;
-        }
-        return this.warnNodeNotFound(d, mutation.id);
+        // no need to warn here, an ancestor may have already been removed
+        return;
```
YunFeng0817, line 1377:
```suggestion
        return this.warnNodeNotFound(d, mutation.id);
```
I feel like the new changes are equivalent to this.

eoghanmurray: I don't think so, did you miss that the first 'loop' filters the removes so that only nodes present at the start are dealt with in the second forEach loop?

Juice10 (approved):

YunFeng0817 (approved):
Looks great

PR #1146, opened 2023-02-18
chore: web-extension vite.config.ts needs these changes to work for node 14

neilhan:
I am using node 14 to build this project. web-extension fails to build. vite.config.ts needs 2 lines to be backward compatible.

yarn test at root fails. with node 14.19.3.


YunFeng0817 (approved):
Thanks for your contribution.

PR #1150, opened 2023-02-21
fix: First record is checkout

mydea:
I am honestly not 100% sure if this is a bug, or somehow intended.
Currently, the initial full snapshot will never have `isCheckout === true`. If I understand the `isCheckout` flag correctly, I wonder why that is the case, and why this isn't `true` for the initial full snapshot? Maybe there is a good reason for that, or I am misunderstanding something, but IMHO it makes more sense to mark this as a checkout in that case?

### DIFF
--- a/.changeset/metal-eels-tie.md
+++ b/.changeset/metal-eels-tie.md
@@ -1,0 +1,5 @@
+---
+'rrweb': patch
+---
+
+fix: Ensure isCheckout is true on initial snapshots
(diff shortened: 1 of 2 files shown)


YunFeng0817:
I think this is not a bug. The checkout feature corresponds to two config options: `checkoutEveryNth` and `checkoutEveryNms `. The feature allows building full snapshots after the recorder starts. And the `isCheckout` flag is used to distinguish the first fullsnapshot and later checked-out snapshots.

mydea:
Hmm OK, so the idea is that `isCheckout` is _only_ ever set/triggered when `checkoutEveryXXX` is set, and other than this it is never set - I see. Then maybe we have to find a different way to handle this 🤔

PR #1155, opened 2023-02-27
Feat: Add support for replaying :defined pseudo-class of custom elements

YunFeng0817:
Some websites use [:defined](https://developer.mozilla.org/en-US/docs/Web/CSS/:defined) pseudo-class for custom elements, e.g. https://store.google.com/us/product/pixel_7?hl=en-US
In the previous implementation, we don't define any custom elements during the playback so all :defined styles are not applied.
In this pull request, I add the support for replaying :defined pseudo-class of custom elements.

Replayer before this PR:
image

Replayer after this PR:
image


Juice10:
This is great @YunFeng0817! One thing I’m wondering is if [CustomElementRegistry.define()](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define) was never called during recording time this might lead to unexpected behavior. Maybe we should monkeypatch [CustomElementRegistry.define()](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define) to make sure these are accurately captured

YunFeng0817:
Good suggestion. Do you think it's OK to change the data structure SerializedElement to this
```
export type elementNode = {
  type: NodeType.Element;
  tagName: string;
  attributes: attributes;
  childNodes: serializedNodeWithId[];
  isSVG?: true;
  needBlock?: boolean;
  isCustom?: boolean; // determine whether this is a defined custom element
};
```

Juice10:
@YunFeng0817 yes I think that works

Juice10:
@YunFeng0817 I think I've found another edge case for you that currently isn't handled by this PR

- An element gets added to the dom eg. `special-element` (not yet `:defined` or `isCustom`
- After a couple seconds `CustomElementRegistry.define()` gets called and `special-element` is officially defined, triggering `:defined` styles

Juice10 (approved):
This is really great! Thanks @YunFeng0817!

Juice10:
Looks like you have a failing type check. When that's fixed feel free to merge this

bruno-garcia:
Hey folks, any blockers to merging this? It does help! Sentry has cherry picked this fix into our fork.

nafees87n:
Any blockers to this PR? We really need this fix

packages/rrweb/src/record/observer.ts
```diff
@@ -1169,6 +1170,41 @@ function initSelectionObserver(param: observerParam): listenerHandler {
...
+        constructor: CustomElementConstructor,
+        options?: ElementDefinitionOptions,
+      ) {
+        try {
+          customElementCb({
+            define: {
+              name,
+            },
+          });
+        } catch (e) {}
```
Juice10, line 1200: Fix eslint issue
```suggestion
        } catch (e) {
          console.warn(`Custom element callback failed for ${name}`);
        }
```

packages/rrweb/src/record/observer.ts
```diff
@@ -1169,6 +1170,43 @@ function initSelectionObserver(param: observerParam): listenerHandler {
   return on('selectionchange', updateSelection);
 }
 
+function initCustomElementObserver({
+  doc,
+  customElementCb,
+}: observerParam): listenerHandler {
+  const win = doc.defaultView as IWindow;
+  if (!win || !win.customElements) return () => {};
```
Juice10, line 1178:
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-empty-function
  if (!win || !win.customElements) return () => {};
```

PR #1156, opened 2023-02-27
fix the statement which is getting changed by Microbundle

Code-Crash:
While testing the cross-origin iframe feature, I realized that with the compressed version, it's not working and without compression, it's working, after digging into it, I found that, the microbundle is not converting a single line statement properly, please see the attached screenshot.

I will also add this issue to the microbundle issue list as well, however, I think it should be good if we could change the statement properly, or else anyway it will crash, and catch will handle it for cross-origin iframes.

We can see in the below image, `window.parent.document` is there as a normal statement, which gets trimmed somehow as `window` in the next compressed version of the image. 

uncompress

We can see here, it's getting replaced as `window` only, which is causing cross-origin iframe not able to work as expected.

compress

So, if we use this as a proper statement as per this pull request, it's working fine, please see if we can merge it @Juice10

### DIFF
--- a/.changeset/chatty-cherries-train.md
+++ b/.changeset/chatty-cherries-train.md
@@ -1,5 +1,5 @@
 ---
-"rrweb": patch
+'rrweb': patch
 ---
 
 Fix the statement which is getting changed by Microbundle
--- a/.changeset/chatty-cherries-train.md
+++ b/.changeset/chatty-cherries-train.md
@@ -1,0 +1,5 @@
+---
+"rrweb": patch
+---
+
+Fix the statement which is getting changed by Microbundle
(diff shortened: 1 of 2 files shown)


Code-Crash:
Sure, thanks.

Juice10 (approved):
Thanks @Code-Crash

YunFeng0817 (approved):

Code-Crash:
FYI: @Juice10 

In Microbundle, it's expected behavior, please see the thread : https://github.com/developit/microbundle/issues/1029

Juice10:
Thanks for sharing @Code-Crash

PR #1157, opened 2023-02-27
fix: Explicitly handle `null` attribute values

mydea:
When an attribute is removed via `el.removeAttribute('my-attr')`, this triggers the mutation observer with a `null` value for the attribute. This is currently not handled by the code/types.

This PR adjusts this to handle this case explicitly.


YunFeng0817 (approved):
Great catch! Thanks for your contribution.

Issue #1118, opened 2023-02-09
[Bug]: record failed


zhaobosky:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

v.2.0.0 alpha4

### Expected Behavior

record.emit send correct dom data,when page load complete

### Actual Behavior

1. record.emit don't send dom data,when page load complete
2. rebuild failed in some pages

### Steps to Reproduce

1.open page ,record.emit don't send data
https://www.gmgitc.com/Bid/BidList

zhaobosky:
https://www.gmgitc.com/Bid/BidList
https://bd.ispacechina.com/retrieve.do?typflag=1

use try catch in snapshot.ts ,the page can snapshot success

<img width="594" alt="image" src="https://user-images.githubusercontent.com/22151144/218902652-84d18105-cf89-49d3-8dab-abd526c25a3e.png">

<img width="946" alt="image" src="https://user-images.githubusercontent.com/22151144/218903030-57433677-2a55-49b4-b3d8-cd89b9f19e96.png">
<img width="561" alt="image" src="https://user-images.githubusercontent.com/22151144/218901300-19fa2298-f73b-4ed8-b37f-ac804b24073c.png">

YunFeng0817:
After some debug, the reason is that in your demo page,  you polyfill the classList method for all HTML elements but the implementation seems to be not right. 
<img width="978" alt="image" src="https://user-images.githubusercontent.com/27533910/220096605-5c2f6325-05c3-4253-82da-678495f4b034.png">

zhaobosky:
> After some debug, the reason is that in your demo page, you polyfill the classList method for all HTML elements but the implementation seems to be not right. <img alt="image" width="978" src="https://user-images.githubusercontent.com/27533910/220096605-5c2f6325-05c3-4253-82da-678495f4b034.png">

https://www.gmgitc.com/Bid/BidList

<img width="1277" alt="image" src="https://user-images.githubusercontent.com/22151144/220301637-874d0686-3f74-42c4-a6ff-d19e20f1b05a.png">


The list data is fetched by ajax，but rrweb don't monitor the dom changes
<img width="955" alt="image" src="https://user-images.githubusercontent.com/22151144/220303634-0161ce73-01ae-4300-ba40-8cd4ed717f95.png">

PR #1163, opened 2023-03-01
improve: some websites rebuild imcomplete

zhaobosky:
fixed https://github.com/rrweb-io/rrweb/issues/1118

Some websites, addedSet in emit function is not empty, but the result converted from Array.from is empty

Some websites polyfill classList functions of HTML elements. Their implementation may throw errors and cause the snapshot to fail. I add try-catch statements to make the code robust.

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -332,7 +332,7 @@ export default class MutationBuffer {
       this.mirror.removeNodeFromMap(this.mapRemoves.shift()!);
     }
 
-    for (const n of Array.from(this.movedSet.values())) {
+    for (const n of this.movedSet) {
       if (
         isParentRemoved(this.removes, n, this.mirror) &&
         !this.movedSet.has(n.parentNode!)
@@ -342,7 +342,7 @@ export default class MutationBuffer {
       pushAdd(n);
     }
 
-    for (const n of Array.from(this.addedSet.values())) {
+    for (const n of this.addedSet) {
       if (
         !isAncestorInSet(this.droppedSet, n) &&
         !isParentRemoved(this.removes, n, this.mirror)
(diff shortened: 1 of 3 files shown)


packages/rrweb/src/utils.ts
```diff
@@ -228,11 +228,15 @@ export function isBlocked(
       : node.parentElement;
   if (!el) return false;
 
-  if (typeof blockClass === 'string') {
-    if (el.classList.contains(blockClass)) return true;
-    if (checkAncestors && el.closest('.' + blockClass) !== null) return true;
-  } else {
-    if (classMatchesRegex(el, blockClass, checkAncestors)) return true;
+  try {
```
YunFeng0817, line 231: Thanks for your fix. Actually, I've already fixed this in #1148. Would you like to take a look at it?

packages/rrweb/src/record/mutation.ts
```diff
@@ -332,7 +332,7 @@ export default class MutationBuffer {
       this.mirror.removeNodeFromMap(this.mapRemoves.shift()!);
     }
 
-    for (const n of Array.from(this.movedSet.values())) {
+    for (const n of this.movedSet) {
```
YunFeng0817, line 335: `movedSet` is a Set object, so I don't think it's correct to iterate a set in this way.

packages/rrweb/src/utils.ts
zhaobosky: I modified the file utils.ts, not snapshot.ts

packages/rrweb/src/record/mutation.ts
zhaobosky: Set object support iterate in this way

YunFeng0817 (approved):
Thank you for your contribution.

PR #1164, opened 2023-03-02
feat: Better masking of option/radio/checkbox values

mydea:
This PR improves masking via `maskAllInputs` for:

* `<option>` elements as part of `<select>`
* `<input type="checkbox">`
* `<input type="radio">`

## What is masked

For radio/checkbox, we only mask the `value="xxx"` attribute. So for example:

```html
<input type="radio" value="my-val">
```

becomes 

```html
<input type="radio" value="******">
```

If no value is specified, nothing is added.

For `option` tags, we mask both the `value` attribute on them (so they also match the value of the `select`), as well as the content of the option - this is IMHO more in line with text input masking than to leave the displayed text.

```html
<select value="val1">
  <option value="val1">Value 1</option>
  <option value="val2">Value 2</option>
</select>
```

becomes

```html
<select value="****">
  <option value="****">*******</option>
  <option value="****">*******</option>
</select>
```

### DIFF
--- a/packages/rrweb-snapshot/src/utils.ts
+++ b/packages/rrweb-snapshot/src/utils.ts
@@ -174,17 +174,21 @@ export function maskInputValue({
 }: {
   element: HTMLElement;
   maskInputOptions: MaskInputOptions;
-  tagName: string;
-  type: string | null;
+  tagName: Uppercase<string>;
+  type: Lowercase<string> | null;
   value: string | null;
   maskInputFn?: MaskInputFn;
 }): string {
   let text = value || '';
-  const actualType = type && toLowerCase(type);
+
+  // Handle `option` like `select
+  if (tagName === 'OPTION') {
+    tagName = 'SELECT';
+  }
 
   if (
     maskInputOptions[tagName.toLowerCase() as keyof MaskInputOptions] ||
-    (actualType && maskInputOptions[actualType as keyof MaskInputOptions])
+    (type && maskInputOptions[type as keyof MaskInputOptions])
   ) {
     if (maskInputFn) {
       text = maskInputFn(text, element);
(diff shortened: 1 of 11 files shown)


packages/rrweb-snapshot/src/utils.ts
```diff
@@ -167,6 +167,12 @@ export function maskInputValue({
   maskInputFn?: MaskInputFn;
 }): string {
   let text = value || '';
+
+  // Handle `option` like `select
+  if (tagName.toLowerCase() === 'option') {
+    tagName = 'select';
+  }
+
```
Juice10, line 175: This method always gets an upper case tagName, keeping it upper case would save us one `toLowerCase` call, slightly improving performance
```suggestion
  // Handle `option` like `select
  if (tagName === 'OPTION') {
    tagName = 'SELECT';
  }

```

YunFeng0817 (changes requested):
I don't understand why this pull request can lead to the change of this snapshot

packages/rrweb-snapshot/src/utils.ts
mydea: Yeah, I wasn't quite sure here myself, as sometimes tagNames are passed as lowercase, sometimes as uppercase strings 😅

mydea: Oops, actually it is _not_ always lowercase (which nicely the tests caught!). 
I've actually changed the type to use `Uppercase` to ensure that the correct case is passed in. This is supported since TS 4.1, is this OK?

YunFeng0817 (commented):
Sorry for the late review. There are so many changes to the Jest snapshot files so I took some time to check them.
I may find an issue with masking radio/checkbox elements. It would be nice to check them. I'm willing to merge this PR after then.

packages/rrweb/test/__snapshots__/integration.test.ts.snap
```diff
@@ -7955,17 +8174,26 @@ exports[`record integration tests should not record input values if maskAllInput
     \\"type\\": 3,
     \\"data\\": {
       \\"source\\": 5,
-      \\"text\\": \\"off\\",
+      \\"text\\": \\"radio-on\\",
```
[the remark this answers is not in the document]
mydea: found the culprit and fixed it! 🎉

mydea:
I rebased & updated this PR on latest master!

PR #1170, opened 2023-03-08
feat: Ensure password inputs are masked when switching type

mydea:
This PR improves masking of `<input type="password">` elements in the case that their `type` is switched to `text`. This is quite common in modern UI, where often you have a button to "show password" while typing it. We should still mask this content in that case as if it was a password, in order to prevent accidentally capturing this unmasked.

This is done by adding an `data-rr-is-password` attribute to the input in the case where the type is changed from `password` to something else.

Note: While at it, I also changed the code to actually accommodate uppercase type, as e.g. `<input type="PASSWORD">`  is valid but would not have been masked previously, as we never converted the type to lowercase.


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -674,8 +674,13 @@ function serializeElementNode(
       attributes.type !== 'button' &&
       value
     ) {
+      const type: string | null = n.hasAttribute('data-rr-is-password')
```
YunFeng0817, line 677: Could you make this ternary operation a common function like "getInputType"  and add some comments about this attribute?

mydea: I extracted this into a util function!

YunFeng0817 (approved):
Thank you!

PR #1173, opened 2023-03-10
fix: Cannot set property attributeName of #<MutationRecord> which has only a getter

wfk007:
1. fix MutationRecord type error: [Properties in MutationRecord are all ReadOnly](https://dom.spec.whatwg.org/#interface-mutationrecord) which will throw a runtime type error
2. fix slimDOM type error in `rrweb-snapshot`

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -486,8 +486,9 @@ export default class MutationBuffer {
       }
       case 'attributes': {
         const target = m.target as HTMLElement;
-        let value = (m.target as HTMLElement).getAttribute(m.attributeName!);
-        if (m.attributeName === 'value') {
+        let attributeName = m.attributeName as string;
+        let value = (m.target as HTMLElement).getAttribute(attributeName);
+        if (attributeName === 'value') {
           value = maskInputValue({
             maskInputOptions: this.maskInputOptions,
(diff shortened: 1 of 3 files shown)


packages/rrweb/src/record/mutation.ts
```diff
@@ -509,15 +509,9 @@ export default class MutationBuffer {
         if (
           target.tagName === 'IFRAME' &&
           m.attributeName === 'src' &&
-          !this.keepIframeSrcFn(value as string)
```
YunFeng0817, line 512: The other changes all look good. But what's the motivation for this change?

wfk007: > The other changes all look good. But what's the motivation for this change?

If the attributeName is read-only, this line seems useless, so in order to keep code and logic clean, I delete this line.
But if it is necessary to keep it, I can restore this change.

YunFeng0817: This line is important because some people may rely on this configuration. Do you know how to keep its original logic and keep mutation unchanged?

wfk007: > This line is important because some people may rely on this configuration. Do you know how to keep its original logic and keep mutation unchanged?

@YunFeng0817 updated

PR #1183, opened 2023-03-20
fix: Ensure attributes are lowercased when checking

mydea:
This is currently not enforced, and could lead to bugs.

I tried to add a test but it seems that `absoluteToDoc()` always returns `''` in the tests (for whatever reason), making this rather hard to test. Happy about any pointers how to test this nicely!

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -21,6 +21,7 @@ import {
   isNativeShadowDom,
   getCssRulesString,
   getInputType,
+  toLowerCase,
 } from './utils';
 
 let _id = 1;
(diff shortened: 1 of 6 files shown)


YunFeng0817 (approved):

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -629,7 +631,7 @@ function serializeElementNode(
       attributes[attr.name] = transformAttribute(
         doc,
         tagName,
-        attr.name,
+        attr.name.toLowerCase() as Lowercase<string>,
```
YunFeng0817, line 634: Could you please help us fix the Eslint error here:
https://github.com/rrweb-io/rrweb/actions/runs/4466789945/jobs/7845419305?pr=1183#step:6:58
and https://github.com/rrweb-io/rrweb/actions/runs/4466789945/jobs/7845419305?pr=1183#step:6:62

mydea: sorry, kind of forgot about this 😅 I updated this, hopefully fixing linting issues! I extracted a type-safe `toLowerCase` util out, which preserves what was passed in (stupid that the built-in `.toLowerCase()` does not do that).

PR #1186, opened 2023-03-22
Fix: processed-node-manager is created even in the environment that doesn't need a recorder

YunFeng0817:
### changes:
1. fix the issue: processed-node-manager is created even in the environment that doesn't need a recorder

image

In the built scripts, the new ProcessedNodeManager() statement is bundled into the **global** environment, which means it will be constructed whenever the script is loaded.
Currently, I'm trying to create a rrweb player page with the SSR (server-side render) and it doesn't have the RFA API on the server side. So it keeps throwing the error: `requestAnimationFrame isn't defined`.

2. solve ESLint errors introduced in #1170

### DIFF
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -615,6 +615,7 @@ function record<T = eventWithTime>(
     }
     return () => {
       handlers.forEach((h) => h());
+      processedNodeManager.destroy();
       recording = false;
       unregisterErrorHandler();
     };
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -4,11 +4,7 @@ import {
   SlimDOMOptions,
   createMirror,
 } from 'rrweb-snapshot';
-import {
-  initObservers,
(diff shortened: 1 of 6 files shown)


packages/rrweb/src/record/index.ts
```diff
@@ -306,6 +303,8 @@ function record<T = eventWithTime>(
       });
   }
 
+  const processedNodeManager = new ProcessedNodeManager();
```
Juice10, line 306: Maybe this should also be added to the list of handlers so it can get turned off once recording stops

Juice10 (commented):
This is great, one things I'm afraid of now we've moved it to the record function is that we might get multiple ProcessedNodeManager raf loops if someone stops recording, and starts again. I think we should add it to the list of handlers that get switched off once recording stops.

packages/rrweb/src/record/index.ts
YunFeng0817: Nice catch! I will change it soon

Juice10 (approved):

PR #1187, opened 2023-03-23
Perf: optimize the performance of processMutations

wfk007:
`processMutations` contains two phases:

1. **processMutation**
2. **emit**

and they have bad performance in some cases.

### case

In the case below: (reproduce link: https://github.com/wfk007/test-processMutation)

we create 1000 ElementNode and insert them into `#target`, each Node is the parent of the following Node

```html
<html>
    <body>
        <button id="add-nodes">Add multi nodes</button>

        <div id="target"></div>

        <script src="https://cdn.jsdelivr.net/npm/rrweb@latest/dist/record/rrweb-record.min.js"></script>
        <script>
            const addNodes = document.querySelector('#add-nodes');
            const target = document.querySelector('#target');

            addNodes.addEventListener('click', () => {
                const total = 1000;
                let parent = target;
                for (let i = 0; i < total; i++) {
                    const el = document.createElement('div');
                    el.textContent = `el-${i + 1}`;
                    parent.append(el);
                    parent = el;
                }
            });
            rrwebRecord({
                sampling: {
                    mouseInteraction: false,
                    mousemove: false,
                    scroll: 300,
                    input: 'last',
                },
                slimDOMOptions: true,
                emit(event) {
                    console.log(event);
                },
            });
        </script>
    </body>
</html>
```

### performance

WITHOUT `record`, it takes about **50ms** to render

image

---

WITH `record` it takes **655ms** and processMutation takes **604ms**

image

### reason

The dom structure below can be created in many ways:
```
body
    E1
       E2
```

Approach 1:
1. create E1
2. create E2
3. E1 append E2
4. body append E1

Browser Mutation observer callback result:
```
[
    {
        target: body,
        addedNodes: [E1],
        type: "childList",
    },
]
```

approach 2:
1. create E1
2. body append E1
3. create E2
4. E1 append E2

Browser Mutation observer callback result:
```
[
    {
        target: body,
        addedNodes: [E1],
        type: "childList",
    },
    {
        target: E1,
        addedNodes: [E2],
        type: "childList",
    },
]
```

In order to cover different Mutation observer callback results and track adds/removes Nodes, RRweb loops all mutations and its' `addedNodes`.
Each Node in `addedNodes` will still travel all its' childNodes.

Considering approach2 result:
1. loop array index 0:
travel E1
travel E2(which is the childNode of E1) 
2. loop array index 1:
travel E2(no need to travel E2 and its' childNodes again)

### solution

In `genAdds`, if the node is added to `addedSet` or `movedSet`, there is no need to travel it and its' children again

after optimizing, it takes **160ms** and processMutation takes **77ms**

image

### case
(case is on the way)

### performance

when the size of addedSet is huge(2w+ Nodes), the for-loop is very slow with time complexity O(n^2) to find a valid Node

image

### reason

image
There are 20000 nodes in addList, and index 900 is the first valid Node from right to left, but its' previous is not a valid Node. we loop from right to left to find the first valid Node

1. addList.get(19999): O(N) time complexity and it is very slow
2. addList.get(19998): O(N) time complexity and it is slow too
3. ...
4. addList(900): find valid Node
5. 899 is selected as candidate Node
6. 899 is invalid
7. so we loop from right to left
8. repeat 1/2/3...

It is very very slow(takes about 16s).

### solution

I create a linkedNodeList based on DoubleLinkedList to accelerate node query with array index, and query time complexity decrease from O(n) to O(1) with a huge performance boost

image

image

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -358,6 +358,16 @@ export default class MutationBuffer {
     }
 
     let candidate: DoubleLinkedListNode | null = null;
+
+    // create linkedNodeList based on DoubleLinkedList to accelerate node query with array index
+    // time complexity O(n) decrease to O(1)
+    let linkedNodeList: DoubleLinkedListNode[] | null = [];
+    let headNode = addList.head;
+    while (headNode) {
+      linkedNodeList.push(headNode);
+      headNode = headNode.next;
+    }
+
     while (addList.length) {
       let node: DoubleLinkedListNode | null = null;
       if (candidate) {
@@ -368,10 +378,10 @@ export default class MutationBuffer {
         }
       }
       if (!node) {
-        for (let index = addList.length - 1; index >= 0; index--) {
-          const _node = addList.get(index);
+        for (let index = linkedNodeList.length - 1; index >= 0; index--) {
+          const _node = linkedNodeList[index];
           // ensure _node is defined before attempting to find value
-          if (_node) {
+          if (_node && _node.value.__ln) {
             const parentId = this.mirror.getId(_node.value.parentNode);
             const nextId = getNextId(_node.value);
 
@@ -417,7 +427,7 @@ export default class MutationBuffer {
       addList.removeNode(node.value);
       pushAdd(node.value);
     }
-
+    linkedNodeList = null;
     const payload = {
       texts: this.texts
         .map((text) => ({
@@ -654,6 +664,9 @@ export default class MutationBuffer {
   private genAdds = (n: Node, target?: Node) => {
     // this node was already recorded in other buffer, ignore it
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
+
+    // if n is added to set, there is no need to travel it and its' children again
+    if (this.addedSet.has(n) || this.movedSet.has(n)) return;
 
     if (this.mirror.hasNode(n)) {
       if (isIgnored(n, this.mirror)) {
--- a/.changeset/flat-foxes-juggle.md
+++ b/.changeset/flat-foxes-juggle.md
@@ -1,0 +1,5 @@
+---
+'rrweb': patch
+---
+
+Optimize performance of the processMutations


wfk007:
The dom structure below can be created in many ways:
```
body
    E1
       E2
```

Approach 1:
1. create E1
2. create E2
3. E1 append E2
4. body append E1

Browser Mutation observer callback result:
```
[
    {
        target: body,
        addedNodes: [E1],
        type: "childList",
    },
]
```

approach 2:
1. create E1
2. body append E1
3. create E2
4. E1 append E2

Browser Mutation observer callback result:
```
[
    {
        target: body,
        addedNodes: [E1],
        type: "childList",
    },
    {
        target: E1,
        addedNodes: [E2],
       type: "childList",
    },
]
```

In order to cover different Mutation observer callback results and track adds/removes Nodes, RRweb loops all mutations and its' `addedNodes`.
Each Node in `addedNodes` will still travel to all childNodes.

Considering approach2 result:
1. loop array index 0:
travel E1
travel E2(which is the childNode of E1) 
2. loop array index 1:
travel E2(no need to travel E2 and its' childNodes again)

Juice10:
Hey @wfk007 this is a great performance improvement, thank you so much for putting all the hard work into this.
One thing I'm missing is making sure your hard work doesn't get reverted by accident in the future.
We have a benchmark suite in the test suite that would be a great candidate for this, especially in combination with a GitHub action (that we don't have yet) https://github.com/benchmark-action/github-action-benchmark that'll make sure performance gets tested on each pull request.

eoghanmurray:
I haven't looked too closely but the change is nice and small and seems to make sense.
I'm putting this out to a 'canary' environment today to see if any problems arise.
Note: The ESLint failure can be solved by rebasing upon the current trunk.

wfk007:
> Hey @wfk007 this is a great performance improvement, thank you so much for putting all the hard work into this. One thing I'm missing is making sure your hard work doesn't get reverted by accident in the future. We have a benchmark suite in the test suite that would be a great candidate for this, especially in combination with a GitHub action (that we don't have yet) https://github.com/benchmark-action/github-action-benchmark that'll make sure performance gets tested on each pull request.

@Juice10 @eoghanmurray I have been trapped at work and complex trivial things in life recently, and I will add a benchmark test later, thanks for your review and great suggestions.

wfk007:
@Juice10 @eoghanmurray Hi, I plan to separate this PR into two, thanks for your great suggestions again

part one: https://github.com/rrweb-io/rrweb/pull/1214

part two: is doing

packages/rrweb/src/record/mutation.ts
```diff
@@ -358,6 +358,16 @@ export default class MutationBuffer {
     }
 
     let candidate: DoubleLinkedListNode | null = null;
+
+    // create linkedNodeList based on DoubleLinkedList to accelerate node query with array index
+    // time complexity O(n) decrease to O(1)
+    let linkedNodeList: DoubleLinkedListNode[] | null = [];
```
YunFeng0817, line 364: This is a good idea. The optimization idea is to use an index array to accelerate fetching an element from a linked list. Cause getting an element from an array is O(1). I would suggest that you could have the variable name 'addedNodeIndexArr' to explain the optimisation thought.

packages/rrweb/src/record/mutation.ts
```diff
@@ -417,7 +427,7 @@ export default class MutationBuffer {
       addList.removeNode(node.value);
```
YunFeng0817, line 427: I'm not sure if there is a bug here so it would be better to have you take a look at this.
`addList` removes an element here and the element indexes would get changed in the next loop. But the index array `linkedNodeList` is not updated at the same time. Would this possibly cause inconsistency and potential bugs?

PR #1188, opened 2023-03-23
feat: Added support maskInputFn with HTMLElement

benjackwhite:
Commonly people want to have the safety of knowing all inputs are masked but also the ability to selectively unmask (perhaps for a search box for example).

This PR extends the `MaskInputFn` to have a second argument which is the `HTMLElement`. This way users can easily define their own logic to decide whether the input should be masked or not.


benjackwhite:
@Juice10 don't want to bother but what would be the best way to get attention for PRs that are good to go?

Juice10 (approved):

Juice10:
@benjackwhite Thanks for submitting this PR. At the moment we go through all the open pull requests once a month as a team, or ourselves whenever we have time. We're hoping to attract some more sponsorship for the project which will allow us to free up some more time to do this more often.

PR #1191, opened 2023-03-28
Fix: Trigger mouse movement & hover with mouse up/down in sync mode

Juice10:
Trigger mouse movement and hover with mouse up and mouse down events when replayer.pause(...) is called.

Currently mouseUp/mouseDown wasn't triggering any mouse movement or hover in isSync (pause/seek) mode.


eoghanmurray:
What if replayer is scrubbing/skipping over the hover/unhover — will many events get dispatched?

Juice10:
> What if replayer is scrubbing/skipping over the hover/unhover — will many events get dispatched?

Only one hover event will be applied on Flush once all events have been applied

eoghanmurray (approved):

PR #1198, opened 2023-04-05
fix: Set finished=false in goto instead of handleProgressClick

charliegracie:
If the player finishes a replaying a video the Controller's `finished` boolean is set to true. This allows the video to be restarted at the beginning if the controller is toggled to play. If a user clicks on the progress bar the `finished` boolean is reset so when `toggle` is called it starts at the right place. If a user programatically calls `goto`, which is the underlying API for handling the progress bar click, the next invocation of `toggle` will start the video at the beginning instead of the proper location.

### DIFF
--- a/.changeset/controller-finish-flag.md
+++ b/.changeset/controller-finish-flag.md
@@ -1,0 +1,6 @@
+---
+"rrweb-player": patch
+"rrweb": patch
+---
+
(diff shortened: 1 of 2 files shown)


charliegracie:
Looking into creating the changeset

YunFeng0817 (approved):
Good catch! Thanks for your contribution.

PR #1184, opened 2023-03-20
fix: Fix input.type check

mydea:
Actually I noticed that `el.type` returns `text` when type is not explicitly set, so this is slightly incorrect.

This was slightly incorrect in https://github.com/rrweb-io/rrweb/pull/1170

PR #1200, opened 2023-04-11
fix: change default value of input type from null to text

re-fort:
according to [W3C](https://www.w3.org/TR/2011/WD-html5-20110525/the-input-element.html), when omitting the type of input(e.g. `<input />`), the type must be treated as text.
> The missing value default is the [Text](https://www.w3.org/TR/2011/WD-html5-20110525/states-of-the-type-attribute.html#text-state-and-search-state) state.

after #1170, that case is treated as [null](https://github.com/rrweb-io/rrweb/pull/1170/files#diff-048da15f20b717b1ea5aa5346398dfed2fe4e24a7d38445b94cbe707640c9fc5R578) and is not masked correctly.
this PR resolves this.


re-fort:
cc: @mydea

YunFeng0817:
This looks good to me. @mydea Do you think it's OK to set 'text' as the default value?

mydea:
This should not be necessary, because `input.type` will already return `text` for `<input />` - I actually fixed this in a previous PR.

IMHO I would not do this change, because right now it will return `null` only when called on non-input elements, which IMHO is safer than to have e.g. `getInputType(divElement) === 'text'`?

I fixed this here: https://github.com/rrweb-io/rrweb/pull/1184 -  maybe we can/should just add a comment indicating that `input.text` will default to `text`?

re-fort:
@mydea 
ah, sorry.
i couldn't have checked your previous PR properly.
i've just added a comment in this PR.

and it would be helpful if you could also consider releasing rrweb@2.0.0-alpha.8

YunFeng0817 (approved):

YunFeng0817:
@re-fort 2.0.0-alpha.8 has been released!

PR #1201, opened 2023-04-12
Pointer type fixups

eoghanmurray:
These should have been part of #1129  as I didn't realize the tests were failing


Juice10 (approved):

packages/rrweb/test/__snapshots__/integration.test.ts.snap
```diff
@@ -1847,7 +2135,8 @@ exports[`record integration tests can record form interactions 1`] = `
     \\"data\\": {
       \\"source\\": 2,
       \\"type\\": 1,
-      \\"id\\": 27
+      \\"id\\": 27,
+      \\"pointerType\\": 0
```
YunFeng0817, line 2139: I'm wondering if we can hide this property by default if the pointer type is mouse. 
On the one hand, it can keep most of the data unchanged. On the other hand, most data is triggered by mouse so it can help reduce event size.

eoghanmurray: Note: I've merged the pull request as this is a fixup for #1129 which is already merged

eoghanmurray: > hide this property by default if the pointer type is mouse

I thought of that and am still open to it, some considerations:
 - it's also possible (older browsers I guess) that the PointerType info missing, in which case it will be absent (if we went with your suggestion then we could add a `PointerTypes.Unknown` for this case)
 - we don't tend to do this for other events ... for example at Statcounter I measured the most common fields/values and by far the worst offender in terms of space was Mutations, so we stripped out the data.source and empty attributes/removes/adds/texts on those for the biggest win.  TouchUp/Down/MouseUp/Down/Click events are hardly as common as scroll events etc.

eoghanmurray: I guess I could remove pointerType from MouseDown/MouseUp/TouchDown/TouchUp as these are implicit, will do so in a new PR

eoghanmurray: After https://github.com/rrweb-io/rrweb/pull/1206 the pointerType is only on Click events .... I'm also thinking that "most data is triggered by mouse" is an incorrect assumption, as most web browsing is from mobile devices.

PR #1212, opened 2023-04-27
Feat/improve custom mask text selector and mask text fn

FANCaipei:
Add customMaskTextRule in recordOptions.
'customMaskTextRule' is an array which it's item contains cssSelector & maskFn. With this option, user can config multi rules for text masking.

'maskTextSelector' & 'maskTextFn'  are reserved, but with lower priority than customMaskTextRule.
Once one of customMaskTextRule match, maskTextSelector & maskTextFn won't be applied.


eoghanmurray:
Is this PR still relevant?  Does #1310 provide anything of what you are looking for?

PR #1225, opened 2023-05-15
fix: Rrror parser throw

benjackwhite:
The error parser can throw if it fails to parse the error which is very meta. This ends up being logged / caught in Sentry which is unnecessary. 

## Changes
Rather than wrapping all of the places it is used, simply changes the function to return cleanly with a warning


YunFeng0817 (approved):
This change makes sense.

Issue #860, opened 2022-03-11
Architecture discussion: Deferred events


Juice10:
## Summary

rrweb currently doesn't have a preferred way of dealing with events that take a while to process before they are fully captured. Let's discuss strategies to deal with these types of events.

### Handling async commands

Some commands are only available async so we have to wait till they are complete before we're able to save the event.

### Keeping event order
In some cases event order is important.
Example of two events happening within the timespan of one second:
1. SLOW EVENT: adds element A (capture takes 10 seconds to complete)
2. FAST EVENT: does something with element A (takes 1ms to complete)
If we were to emit these events when they are completed processing they'll be saved out of order, breaking the replay.

### Performance / web workers
Some heavy synchronous tasks we do (e.g. canvas.toDataURL, JSON.stringify base64 encoding) cause [jank](https://developer.mozilla.org/en-US/docs/Glossary/Jank). Moving this to a web worker would mean having a wait longer for the results but a much smoother user experience during recording.

### Image inlining
When using `fetch` to request images to inline we have to wait for the request to complete. Most of the time this will be pretty fast as the browser probably still has the image in cache, but this is an async operation.
More info: https://github.com/rrweb-io/rrweb/pull/836#issuecomment-1047173966

### Canvas mutation serialization
Serializing some arguments of canvas calls is only available async. Examples: [Blob#arrayBuffer()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer) [Canvas#toBlob()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob)
With canvas recording order is incredibly important, if two events get recorded out of order the whole recording would (silently) break yielding an empty canvas (which is a PITA to debug).

Juice10:
# Potential solutions

## Event

### Events wrapped in a promise
When events are emitted they are wrapped in a promise and its up to the end user to `await` for them to finish.

### Attributes wrapped in a promise
For example, attribute pending:
```js
{
  timestamp: 1234,
  data: {
    content: [Promise]
  }
}
```
When finished it could overwrite the attribute:
```js
{
  timestamp: 1234,
  data: {
    content: 'base64encodedstring'
  }
}
```
## Recording API
### Two `emit` methods on record
One for the events at the moment they occur and one for the events when they are completed.

### One `emit` method: emit in order of occurrence
[...]

PR #1239, opened 2023-06-08
Add Asset event type and capture assets

Juice10:
Asset events allow for asynchronous events containing (image or potentially other) assets.
See https://github.com/rrweb-io/rrweb/issues/860 for more info.

These are async events that get emmited after the original dom mutation events and can be used to inject these assets after the fact in the player or via post processing.

Todo:
- [x] create asset event for assets that failed to load
- [x] implement asset in player
- ~[ ] add flag to log assets being captured~
- [x] add documentation for recording config
- [x] [docs] explain how its different to inline images
- [x] decided whether or not to move the capturing and encoding to a web worker (out of scope)
- [x] use canvas webgl code to do blob encoding
- [x] move more types from rrweb-snapshot to @rrweb/types to circumvent circular dependencies
- [x] add meta data to meta event including what asset origins are going to be cached, 
  - [x] also filter hooks in Asset Manager listening to those origins
- [x] add support for all elements
  - [x] video
  - [x] audio
  - [x] embed
  - [x] source
  - [x] track
  - [x] input#type=image
  - [x] img#srcset
  - [x] svg
  - [x] table#background
- ~[ ] Remove `inlineImages` in favour of `assetCapture`~
- [x] Forward `inlineImages` config to `assetCapture`
- [x] get it working in `yarn live-stream`
- [x] hide `src` attributes while asset is being awaited
- [x] work with rrdom
- [x] add tests for svg elements, ~also update asset manager to work with setAttributeNS~ _setAttributeNS was needed for `xlink:href`, but all modern browsers rewrite that to `href` which is already supported out of the box._
- [x] add implementation for `srcset`
- [x] rename `replay/asset/index.ts` to `replay/asset-manager/index.ts`


eoghanmurray (commented):
I'm wondering whether it's worth taking [this commit](fa2b165cc85678ceb016bd1a3db9be7da86cb534) out and merging in a separate PR?
Fine if not; maybe in future this sort of thing could be merged in earlier while rest of a PR is ongoing.

Juice10:
> I'm wondering whether it's worth taking [this commit](https://github.com/rrweb-io/rrweb/pull/1239/commits/fa2b165cc85678ceb016bd1a3db9be7da86cb534) out and merging in a separate PR? Fine if not; maybe in future this sort of thing could be merged in earlier while rest of a PR is ongoing.

If we weren't on an alpha release that type of commit would also require us to do a major version change. Not sure if it makes sense now, it's bound to break some tests and those got fixed later in subsequent commits. But in general I do agree with you, this type of thing would be better in a separate PR

docs/recipes/assets.md
```diff
@@ -0,0 +1,40 @@
+# Asset Capture Methods & Configuration in rrweb
+
+[rrweb](https://rrweb.io/) is a JavaScript library that allows you to record and replay user interactions on your website. It provides various configuration options for capturing assets (such as images) during the recording process. In this document, we will explore the different asset capture methods and their configuration options in rrweb.
+
+## Inline Images (Deprecated)
+
+The `inlineImages` configuration option is deprecated and should not be used anymore. The previous implementation had some issues, namely rewriting events that are already emitted which might make you miss the inlined image if the event has already been sent to the server. Currently it just turns on `captureAssets`, so use please directly use the `captureAssets` option to configure asset capture.
```
Yuyz0112, line 7: Forwarding to the new API is a good idea, but since it changes the behavior of the old config, shall we add breaking change notes to the changeset?

packages/rrdom-nodejs/package.json
```diff
@@ -29,27 +29,29 @@
...
     "eslint": "^8.15.0",
-    "jest": "^27.5.1",
-    "puppeteer": "^9.1.1",
+    "jest": "^29.7.0",
+    "jest-environment-jsdom": "^29.7.0",
+    "puppeteer": "^21.6.1",
     "rollup": "^2.56.3",
     "rollup-plugin-terser": "^7.0.2",
-    "rollup-plugin-typescript2": "^0.31.2",
+    "rollup-plugin-typescript2": "^0.36.0",
```
Yuyz0112, line 44: this upgrade is great:)

packages/rrweb-snapshot/package.json
```diff
@@ -42,22 +43,27 @@
     "url": "https://github.com/rrweb-io/rrweb/issues"
   },
   "homepage": "https://github.com/rrweb-io/rrweb/tree/master/packages/rrweb-snapshot#readme",
+  "dependencies": {
+    "@rrweb/types": "^2.0.0-alpha.11"
```
Yuyz0112, line 47: we put types as dependencies instead of devDependencies because we import runtime enums?

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -331,3 +329,54 @@ export function getInputType(element: HTMLElement): Lowercase<string> | null {
...
+      // Since we don't know how long the URL will be, we'll assume it's everything
+      // after the first space.
+      urls.push(trimmed.substring(0, spaceIndex));
+    }
+  }
+  return urls;
+}
+
+export const CACHEABLE_ELEMENT_ATTRIBUTE_COMBINATIONS = new Map([
+  ['IMG', new Set(['src', 'srcset'])],
```
Yuyz0112, line 353: It looks like the element types are not configurable in the captureAssets API.

I don't know if a user wants to capture the image src also wants to capture all video/audio elements.

But since the PR's scope is already big, we can leave this as a discussion and do it in the following PRs if needed.

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -331,3 +329,54 @@ export function getInputType(element: HTMLElement): Lowercase<string> | null {
       toLowerCase(type)
     : null;
 }
+
+export function getUrlsFromSrcset(srcset: string): string[] {
+  const urls: string[] = [];
+  const parts = srcset.split(',');
+  for (let i = 0; i < parts.length; i++) {
+    const trimmed = parts[i].trim();
+    const spaceIndex = trimmed.indexOf(' ');
```
Yuyz0112, line 338: shall we add some comments to inline two examples for the following if/else conditions

packages/rrweb-snapshot/tsconfig.json
```diff
@@ -14,5 +15,9 @@
   },
   "exclude": ["test"],
   "include": ["src"],
-  "references": []
+  "references": [
+    {
+      "path": "../types"
```
Yuyz0112, line 20: why we need this config?

packages/rrweb/src/replay/index.ts
```diff
@@ -2173,9 +2206,6 @@ export class Replayer {
 
   private backToNormal() {
     this.nextUserInteractionEvent = null;
-    if (this.speedService.state.matches('normal')) {
```
Yuyz0112, line 2176: why we are removing this?

packages/rrweb/src/replay/index.ts
```diff
@@ -993,16 +1012,25 @@ export class Replayer {
     }
   }
 
+  /**
+   * Process all asset events and preload them
+   */
+  private async preloadAllAssets(timestamp: number): Promise<void[]> {
```
Yuyz0112, line 1018: What will happen if users use the `addEvent` API to put more new asset events?

packages/rrweb/src/replay/asset-manager/index.ts
```diff
@@ -0,0 +1,243 @@
...
+
+          node.setAttribute(attribute, status.url);
+        }),
+      );
+    }
+
+    return Promise.all(promises);
+  }
+
+  public reset(config?: captureAssetsParam | undefined): void {
```
Yuyz0112, line 226: Why do we have to reset when a meta event is captured?

docs/recipes/assets.md
Juice10: Good idea! Here is the changeset for it: https://github.com/rrweb-io/rrweb/pull/1239/commits/0115cb65895e0bceb38b4bf36273b7d221a0d4b0

packages/rrweb-snapshot/src/utils.ts
Juice10: Yes that would be a nice configuration option I think (which I would indeed prefer to do separately from this PR). And I think adding a max-asset size would also be a useful configuration option as well.

packages/rrweb-snapshot/src/utils.ts
Juice10: Good idea, I've added that here: https://github.com/rrweb-io/rrweb/pull/1239/commits/4484a259a00120c690d7e890903b7f865429d2f2

packages/rrweb/src/replay/asset-manager/index.ts
Juice10: This is for live-mode/addEvent. Once a meta event is triggered its an indication that a page navigation has happened. That tells us that we don't have to keep waiting for new asset events to show up, hence the `reset` being called.

packages/rrweb/src/replay/index.ts
Juice10: If live mode is triggered we have a slightly different way of working.
In that case we wait for the asset to be loaded before we add its `src` attribute with the final url
https://github.com/rrweb-io/rrweb/blob/a6a1c9307b33d20a4bf928db52aba59ec09ca6a5/packages/rrweb/src/replay/asset-manager/index.ts#L192-L221

packages/rrweb-snapshot/package.json
Juice10: I think this is why I did it, but to be honest I don't remember so I'll move it to `devDependencies` and see if that breaks anything, maybe it will be fine.

Juice10: Just moved this to `devDependencies` and it seems to work fine. So we can leave it that way AFAIK

packages/rrweb/src/replay/asset-manager/index.ts
Yuyz0112: I see. But do we need to store the capture config in the meta event, and reload it as `this.config = config` when a meta event is emitted?

Or we can think the capture config is the same across a session and also call reset when a meta event is emitted, but without storing a copy of the capture params in the meta event?

packages/rrweb/src/record/observers/asset-manager.ts
```diff
@@ -0,0 +1,196 @@
...
+              data: [
+                {
+                  rr_type: 'ArrayBuffer',
+                  base64, // base64
+                },
+              ],
+            };
+
+            this.capturedURLs.add(url);
+            this.capturingURLs.delete(url);
```
Yuyz0112, line 161: how about move this line and the same line in the error block into a `finally` block?

Yuyz0112:
Just reviewed the last two files: the asset manager observer and the replay side manager, really cool implementation!

Leave some small comments, and we should be able to merge this in the following days.

eoghanmurray:
Not sure if this is a goal or a non-goal, but should we be collecting all assets for the `srcset` attribute? Shouldn't we be able to tell which one of the images is the one that will need to be rendered?

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -331,3 +329,61 @@ export function getInputType(element: HTMLElement): Lowercase<string> | null {
...
+      // Since we don't know how long the URL will be, we'll assume it's everything
+      // before the first space.
+      // eg. `image.jpg 2x`
+      urls.push(trimmed.substring(0, spaceIndex));
+    }
+  }
+  return urls;
+}
+
+export const CACHEABLE_ELEMENT_ATTRIBUTE_COMBINATIONS = new Map([
```
eoghanmurray, line 359:
```suggestion
export const CAPTURABLE_ELEMENT_ATTRIBUTE_COMBINATIONS = new Map([
```

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -331,3 +329,61 @@ export function getInputType(element: HTMLElement): Lowercase<string> | null {
...
+  ['TH', new Set(['background'])],
+  ['TBODY', new Set(['background'])],
+  ['THEAD', new Set(['background'])],
+  ['image', new Set(['href'])],
+  ['feImage', new Set(['href'])],
+  ['cursor', new Set(['href'])],
+]);
+
+export function isAttributeCacheable(n: Element, attribute: string): boolean {
+  const acceptedAttributesSet = CACHEABLE_ELEMENT_ATTRIBUTE_COMBINATIONS.get(
```
eoghanmurray, line 382:
```suggestion
  const acceptedAttributesSet = CACHEABLE_ELEMENT_ATTRIBUTE_COMBINATIONS.get(
```

eoghanmurray (changes requested):
Rename cacheable -> capturable

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -23,6 +25,8 @@ import {
   stringifyStylesheet,
   getInputType,
   toLowerCase,
+  getUrlsFromSrcset,
+  isAttributeCacheable,
```
eoghanmurray, line 29:
```suggestion
  isAttributeCapturable,
```

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -617,20 +662,31 @@ function serializeElementNode(
...
+      const value = (attributes[attr.name] = transformAttribute(
         doc,
         tagName,
         toLowerCase(attr.name),
         attr.value,
-      );
+      ));
+
+      // save assets offline
+      if (value && onAssetDetected && isAttributeCacheable(n, attr.name)) {
```
eoghanmurray, line 683:
```suggestion
      if (value && onAssetDetected && isAttributeCapturable(n, attr.name)) {
```

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -331,3 +329,61 @@ export function getInputType(element: HTMLElement): Lowercase<string> | null {
...
+  ['TR', new Set(['background'])],
+  ['TH', new Set(['background'])],
+  ['TBODY', new Set(['background'])],
+  ['THEAD', new Set(['background'])],
+  ['image', new Set(['href'])],
+  ['feImage', new Set(['href'])],
+  ['cursor', new Set(['href'])],
+]);
+
+export function isAttributeCacheable(n: Element, attribute: string): boolean {
```
eoghanmurray, line 381:
```suggestion
export function isAttributeCapturable(n: Element, attribute: string): boolean {
```

packages/rrweb-snapshot/test/utils.test.ts
```diff
@@ -1,9 +1,8 @@
 /**
  * @jest-environment jsdom
  */
-import { NodeType, serializedNode } from '../src/types';
-import { isNodeMetaEqual } from '../src/utils';
-import { serializedNodeWithId } from 'rrweb-snapshot';
+import { isAttributeCacheable, isNodeMetaEqual } from '../src/utils';
```
eoghanmurray, line 4:
```suggestion
import { isAttributeCapturable, isNodeMetaEqual } from '../src/utils';
```

packages/rrweb-snapshot/test/utils.test.ts
```diff
@@ -147,4 +146,50 @@ describe('utils', () => {
       expect(isNodeMetaEqual(element2, element3)).toBeFalsy();
     });
   });
+
+  describe('isAttributeCacheable()', () => {
```
eoghanmurray, line 150:
```suggestion
  describe('isAttributeCapturable()', () => {
```

packages/rrweb-snapshot/test/utils.test.ts
```diff
@@ -147,4 +146,50 @@ describe('utils', () => {
...
+      ['input', ['href']],
+      ['iframe', ['href']],
+      ['object', ['href']],
+    ] as const;
+
+    validAttributeCombinations.forEach(([tagName, attributes]) => {
+      const element = document.createElement(tagName);
+      attributes.forEach((attribute) => {
+        it(`should correctly identify <${tagName} ${attribute}> as cacheable`, () => {
+          expect(isAttributeCacheable(element, attribute)).toBe(true);
```
eoghanmurray, line 181:
```suggestion
          expect(isAttributeCapturable(element, attribute)).toBe(true);
```

packages/rrweb-snapshot/test/utils.test.ts
```diff
@@ -147,4 +146,50 @@ describe('utils', () => {
...
+        });
+      });
+    });
+
+    invalidAttributeCombinations.forEach(([tagName, attributes]) => {
+      const element = document.createElement(tagName);
+      attributes.forEach((attribute) => {
+        it(`should correctly identify <${tagName} ${attribute}> as NOT cacheable`, () => {
+          expect(isAttributeCacheable(element, attribute)).toBe(false);
```
eoghanmurray, line 190:
```suggestion
          expect(isAttributeCapturable(element, attribute)).toBe(false);
```

packages/rrweb/src/record/mutation.ts
```diff
@@ -616,12 +622,26 @@ export default class MutationBuffer {
...
+          const transformedValue = (item.attributes[attributeName] =
+            transformAttribute(
+              this.doc,
+              toLowerCase(target.tagName),
+              toLowerCase(attributeName),
+              value,
+            ));
+          if (
+            transformedValue &&
+            this.assetManager.isAttributeCacheable(target, attributeName)
```
eoghanmurray, line 634:
```suggestion
            this.assetManager.isAttributeCapturable(target, attributeName)
```

packages/rrweb/src/record/observers/asset-manager.ts
```diff
@@ -0,0 +1,196 @@
...
+  eventWithTime,
+  listenerHandler,
+} from '@rrweb/types';
+import type { assetCallback } from '@rrweb/types';
+import { encode } from 'base64-arraybuffer';
+
+import { patch } from '../../utils';
+
+import type { recordOptions } from '../../types';
+import { isAttributeCacheable } from 'rrweb-snapshot';
```
eoghanmurray, line 13:
```suggestion
import { isAttributeCapturable } from 'rrweb-snapshot';
```

packages/rrweb/src/record/observers/asset-manager.ts
```diff
@@ -0,0 +1,196 @@
...
+        });
+
+        this.failedURLs.add(url);
+        this.capturingURLs.delete(url);
+      });
+
+    return { status: 'capturing' };
+  }
+
+  public isAttributeCacheable(n: Element, attribute: string): boolean {
```
eoghanmurray, line 193:
```suggestion
  public isAttributeCapturable(n: Element, attribute: string): boolean {
```

packages/rrweb/src/record/observers/asset-manager.ts
```diff
@@ -0,0 +1,196 @@
...
+    return isAttributeCacheable(n, attribute);
```
eoghanmurray, line 194:
```suggestion
    return isAttributeCapturable(n, attribute);
```

packages/rrweb/src/replay/asset-manager/index.ts
```diff
@@ -0,0 +1,243 @@
+import type {
+  RebuildAssetManagerFinalStatus,
+  RebuildAssetManagerInterface,
+  RebuildAssetManagerStatus,
+  assetEvent,
+  captureAssetsParam,
+} from '@rrweb/types';
+import { deserializeArg } from '../canvas/deserialize-args';
+import { getSourcesFromSrcset, isAttributeCacheable } from 'rrweb-snapshot';
```
eoghanmurray, line 9:
```suggestion
import { getSourcesFromSrcset, isAttributeCapturable } from 'rrweb-snapshot';
```

packages/rrweb/src/replay/asset-manager/index.ts
```diff
@@ -0,0 +1,243 @@
...
+      status: 'unknown',
+    };
+  }
+
+  public isCacheable(
+    n: RRElement | Element,
+    attribute: string,
+    value: string,
+  ): boolean {
+    if (!isAttributeCacheable(n as Element, attribute)) return false;
```
eoghanmurray, line 128:
```suggestion
    if (!isAttributeCapturable(n as Element, attribute)) return false;
```

packages/rrweb/src/replay/asset-manager/index.ts
```diff
@@ -0,0 +1,243 @@
...
+    return cachedOrigins;
+  }
+
+  public async manageAttribute(
+    node: RRElement | Element,
+    nodeId: number,
+    attribute: string,
+  ): Promise<unknown> {
+    const originalValue = node.getAttribute(attribute);
+    if (!originalValue || !this.isCacheable(node, attribute, originalValue))
```
eoghanmurray, line 166:
```suggestion
    if (!originalValue || !this.isCapturable(node, attribute, originalValue))
```

packages/types/src/index.ts
```diff
@@ -694,3 +735,136 @@ export type TakeTypedKeyValues<Obj extends object, Type> = Pick<
...
+export declare abstract class RebuildAssetManagerInterface {
+  constructor(
+    playerConfig: { liveMode: boolean },
+    assetManagerConfig?: captureAssetsParam | undefined,
+  );
+  abstract add(event: assetEvent): Promise<void>;
+  abstract get(url: string): RebuildAssetManagerStatus;
+  abstract whenReady(url: string): Promise<RebuildAssetManagerFinalStatus>;
+  abstract reset(config?: captureAssetsParam | undefined): void;
+  abstract isCacheable(n: Element, attribute: string, value: string): boolean;
```
eoghanmurray, line 762:
```suggestion
  abstract isCapturable(n: Element, attribute: string, value: string): boolean;
```

packages/rrdom/src/diff.ts
```diff
@@ -322,6 +323,11 @@ function diffProps(
       };
     } else if (newTree.tagName === 'IFRAME' && name === 'srcdoc') continue;
     else oldTree.setAttribute(name, newValue);
+
+    if (assetManager?.isCacheable(oldTree, name, newValue)) {
```
eoghanmurray, line 327:
```suggestion
    if (assetManager?.isCapturable(oldTree, name, newValue)) {
```

packages/rrdom/test/diff.test.ts
```diff
@@ -460,6 +462,40 @@ describe('diff algorithm for rrdom', () => {
       diff(element, rrIframe, replayer);
       expect(element.getAttribute('srcdoc')).toBe(null);
     });
+
+    describe('with asset manager', () => {
+      let assetManager: RebuildAssetManagerInterface;
+      beforeEach(() => {
+        assetManager = fromPartial({
+          manageAttribute: jest.fn(),
+          isCacheable: jest.fn(),
```
eoghanmurray, line 471:
```suggestion
          isCapturable: jest.fn(),
```

packages/rrdom/test/diff.test.ts
```diff
@@ -460,6 +462,40 @@ describe('diff algorithm for rrdom', () => {
...
+        const node = document.createElement(tagName);
+        const sn = Object.assign({}, elementSn, { tagName });
+        mirror.add(node, sn);
+
+        const rrDocument = new RRDocument();
+        const rrNode = rrDocument.createElement(tagName);
+        const sn2 = Object.assign({}, elementSn, { tagName });
+        rrDocument.mirror.add(rrNode, sn2);
+
+        (assetManager.isCacheable as jest.Mock)
```
eoghanmurray, line 487:
```suggestion
        (assetManager.isCapturable as jest.Mock)
```

packages/rrweb/src/replay/asset-manager/index.ts
```diff
@@ -0,0 +1,243 @@
...
+        status: 'failed',
+      };
+    }
+
+    return {
+      status: 'unknown',
+    };
+  }
+
+  public isCacheable(
```
eoghanmurray, line 123:
```suggestion
  public isCapturable(
```

eoghanmurray (changes requested):
more cacheable -> capturable

(This is now a commit)

packages/rrweb/src/replay/asset-manager/index.ts
```diff
@@ -0,0 +1,243 @@
...
+      // In live mode we removes the attribute while it loads so it doesn't show the broken image icon
+      if (this.liveMode && nodeId > 0) {
+        let hijackedAttributes = this.nodeIdAttributeHijackedMap.get(nodeId);
+        if (!hijackedAttributes) {
+          hijackedAttributes = new Map();
+          this.nodeIdAttributeHijackedMap.set(nodeId, hijackedAttributes);
+        }
+        hijackedAttributes.set(attribute, originalValue);
+        if (node.tagName === 'IMG' && attribute === 'src') {
+          node.setAttribute('src', '//:0');
```
eoghanmurray, line 200:
```suggestion
          // special value to prevent a broken image icon while asset is being loaded
          node.setAttribute('src', '//:0');
```

package.json
```diff
@@ -42,7 +43,7 @@
     "test": "yarn run concurrently --success=all -r -m=1 'yarn workspaces-to-typescript-project-references --check' 'yarn turbo run test --concurrency=1 --continue'",
     "test:watch": "yarn turbo run test:watch",
     "test:update": "yarn turbo run test:update",
-    "check-types": "yarn turbo run check-types --continue",
+    "check-types": "yarn run concurrently --success=all -r -m=1 'yarn workspaces-to-typescript-project-references --check' 'yarn turbo run check-types --continue'",
```
eoghanmurray, line 46: Is this a regression from the `vite` way to the old way?
this longer version looks to have been added Nov 27 in f986245deb17854f6c499b4ee7baf1d4c0f12baf

Juice10: Just rebased, should be fixed now

Juice10:
Superseeded by https://github.com/rrweb-io/rrweb/pull/1239

PR #1243, opened 2023-06-27
fix: when inlineImages is true, image is cors, the image tag is broken

xujiujiu:
inlineImages is true，image.src is cors, when start record，the image tag is broken

ccb.png is cors

not start
image

started
image


eoghanmurray:
It looks like we are the ones adding the `crossorigin` attribute on the image.
This is presumably a prerequisite to access the image as a blob.
As this can cause failure of image loading as you have discovered, we likely should do this in a separate `new Image()` instead.  Or at least only add it if we have checked `document.location.origin === (new URL(image.src)).origin`

Also note that Justin is working on some async asset code loading in #1239 which will have the side effect of deprecate this inlineImages approach.

xujiujiu:
> It looks like we are the ones adding the `crossorigin` attribute on the image. This is presumably a prerequisite to access the image as a blob. As this can cause failure of image loading as you have discovered, we likely should do this in a separate `new Image()` instead. Or at least only add it if we have checked `document.location.origin === (new URL(image.src)).origin`
> 
> Also note that Justin is working on some async asset code loading in #1239 which will have the side effect of deprecate this inlineImages approach.

Will there be any configuration after #1239  that can instead of inlineImages? After all, we don’t want to affect the picture display speed due to the network speed during playback

xujiujiu:
Tracking in PR of #1359

PR #1247, opened 2023-06-29
Fix types in rrwebPlayer

Juice10:
Was using DeprecatedMirror in its types, that shouldn't be used anymore.
Also `playRange` wasn't being defined but was documented.


YunFeng0817 (approved):
Good catch!

Issue #1250, opened 2023-07-06
[Bug]: Events repeat generated after the iframe appeared


wfk007:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

2.0.0-alpha.9

### Expected Behavior

Resize and MediaInteraction events do not repeat generated after the iframe appeared

### Actual Behavior

Resize and MediaInteraction events repeat generated after the iframe appeared

### Steps to Reproduce

1. git clone git@github.com:wfk007/test-rrweb-repeat-event.git
2. pnpm install
3. pnpm run dev
4. click `[onStart]` button to start record
5. try to play and pause the video
6. try to resize the page
7. click `[toggle-iframe]` button to show iframe
8. repeat steps 5 and 6 above
9. click `[onStop]` button
10. review the generated events in the console, you will see the duplicate Resize and MediaInteraction events

### Testcase Gist URL

https://github.com/wfk007/test-rrweb-repeat-event

PR #1251, opened 2023-07-06
fix: Resize and MediaInteraction events repeat generated after the iframe appeared

wfk007:
fix: https://github.com/rrweb-io/rrweb/issues/1250


Juice10 (approved):
Thanks for finding and fixing this issue @wfk007!

Issue #1208, opened 2023-04-17
[Bug]: Safari does not escape `:` in CSS rule


billyvg:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

v1 + v2

### Expected Behavior

Chrome: https://rrwebdebug.com/play/index.html?url=https%3A%2F%2Fgist.githubusercontent.com%2Fbillyvg%2F36b1028063eef73337958955fb494cf0%2Fraw%2Fbfd605b8244fa4e8431046d8aef8a2a45f8f3130%2Fchrome.json&version=2.0.0-alpha.7&play=on

"loading" text should not appear in replay

### Actual Behavior

Safari: https://rrwebdebug.com/play/index.html?url=https%3A%2F%2Fgist.githubusercontent.com%2Fbillyvg%2F36b1028063eef73337958955fb494cf0%2Fraw%2Fe4a6ac08edc7f09828444a175332b2b9e765d04e%2Fsafari.json&version=2.0.0-alpha.7&play=on

Notice red box + loading text

### Steps to Reproduce

CSS with `:` does not get escape during recording in the Safari browser.

e.g.

```
[foo:bar] {
  display: none;
}
```

In Chrome this gets escaped to `[foo\\:bar]`, but in Safari it is untouched: `[foo:bar]`. During replay (any browser), the `[foo:bar]` rule is ignored.

### Additional Information

Note this is a 5 year old bug in Safari: https://bugs.webkit.org/show_bug.cgi?id=184604

PR #1253, opened 2023-07-07
fix: Fix CSS rules captured in Safari

mydea:
Safari does not capture some css selectors correctly, so we need to fix them, or else playback fails for them.

Closes https://github.com/rrweb-io/rrweb/issues/1208

ref https://github.com/getsentry/rrweb/pull/86

### DIFF
--- a/.changeset/thirty-baboons-punch.md
+++ b/.changeset/thirty-baboons-punch.md
@@ -1,0 +1,5 @@
+---
+'rrweb-snapshot': patch
+---
+
+Fix CSS rules captured in Safari
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -22,6 +22,7 @@ import {
   getCssRulesString,
   getInputType,
   toLowerCase,
+  validateStringifiedCssRule,
 } from './utils';
 
 let _id = 1;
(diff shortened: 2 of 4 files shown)


Juice10 (approved):

Juice10:
@mydea thanks for submitting this, could you add a Changeset then I can merge this issue

mydea:
> @mydea thanks for submitting this, could you add a Changeset then I can merge this issue?

Added a changeset 👍

eoghanmurray:
In light of performance efforts in #1271 I wonder whether it's desirable to run a browser check before employing the `validateStringifiedCssRule` function?

I believe (correct me if I'm wrong) that the `if (cssStringified.includes(':')) {` check will succeed for nearly 100% of css strings, so the regex will be needlessly run in non-safari browsers.  Is that correct?

eoghanmurray:
@mydea I've done a slight rewrite of the test in #1280
if you have a chance, could you validate that that still does the right thing in Safari?

JonasBa:
> In light of performance efforts in #1271 I wonder whether it's desirable to run a browser check before employing the `validateStringifiedCssRule` function?
> 
> I believe (correct me if I'm wrong) that the `if (cssStringified.includes(':')) {` check will succeed for nearly 100% of css strings, so the regex will be needlessly run in non-safari browsers. Is that correct?

Correct, + regexp.replace will always allocate a new string for you, so you are also performing an unnecessary copy

PR #1254, opened 2023-07-07
fix: Handle case where `event` is null/undefined

mydea:
We got one or two bugs that failed due to event not being defined. So we guard against this here.

ref https://github.com/getsentry/rrweb/pull/32

### DIFF
--- a/.changeset/tiny-buckets-love.md
+++ b/.changeset/tiny-buckets-love.md
@@ -1,0 +1,5 @@
+---
+'rrweb': patch
+---
+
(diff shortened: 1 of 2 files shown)


Juice10:
Looks good @mydea! Could you add a Changeset, then I can go ahead and merge this issue

mydea:
> Looks good @mydea! Could you add a Changeset, then I can go ahead and merge this issue

oops, always forget about this, added a changeset!

PR #1257, opened 2023-07-12
feat: Allow for masking of attributes

billyvg:
~Note: this is more of an RFC~
Ideally this would allow users to be able to mask attributes of DOM nodes. An example of this is `placeholder` which can be a bit suspicious to our users when our default state is that everything is masked.

This currently adds a new API, `maskAttributesFn: (key: string, value: string, el: HTMLElement) => string`, that is used as a callback in `transformAttribute`. I prefer this API as it gives more flexibility for users (though it may need to pass the el node for most flexibility), but it is a bit inconsistent with `maskTextFn` and `maskInputFn`.

other options:

~* Rename this to something else (open to ideas)~
* Change this to pass value, and dom element (similar to MaskInputFn) to customize masking instead of decision maker of when to mask and introduce a simpler declarative API for what attributes to mask
~* ???~


colingm:
Just wanted to say that it is awesome to see this, we were going to be bringing up attribute masking from Pendo pretty soon anyways 😅 

So in your solution you opt for just always calling the `maskAttributeFn` instead of having another option that specifies a list of attributes to mask (that could take a wildcard or something) and then only call the function on those particular attributes. Would an option like that provide much performance benefit? I guess it means we would save calling to the function on elements that never even have the attribute in the first place.

Another thought would be if we don't want a list of attributes as an option would it instead be better to just call the `maskAttributeFn` once per element and then let the function decide to mask multiple attributes all at once?

PR #1258, opened 2023-07-17
feat: Pass parent element or node to `maskTextFn`

billyvg:
Allows `maskTextFn` to access to parent element or node so that it can make masking decisions based on the element. This brings the interface to be similar to `maskInputFn`.

### DIFF
--- a/.changeset/wet-snails-rhyme.md
+++ b/.changeset/wet-snails-rhyme.md
@@ -1,0 +1,6 @@
+---
+'rrweb-snapshot': patch
+'rrweb': patch
+---
+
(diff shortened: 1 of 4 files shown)


eoghanmurray:
Looks good
(I presume you've checked passing in a function which doesn't deal with the second parameter? - that would be fine, right?)

benjackwhite:
Only just noticed this PR exists! I just did the same thing but with some slightly stricter typing and tests. 

https://github.com/rrweb-io/rrweb/pull/1310

billyvg:
Closing in favor of #1310

PR #1262, opened 2023-07-24
feat: Add `ignoreSelector` option

billyvg:
Similar to `ignoreClass`, but accepts a CSS selector so that you can use any CSS selector.


eoghanmurray (approved):
Looks fine to me

eoghanmurray:
Could you add a quick changeset please?  See message from changeset-bot in this thread ... I think 'patch' (rather than 'minor') is fine https://semver.org/

billyvg:
@eoghanmurray Added!

PR #1268, opened 2023-07-27
Compact style mutation fixes and improvements

eoghanmurray:
**1)**: CSSOM was failing badly with shorthand + variables as reported in #1246 e.g. in

```
elem.style = "background: var(--my-bg);"
```
the resultant mutation had no record of the variable.  This is a deficiency in the spec for CSSOM as evidenced here:
https://bugs.chromium.org/p/chromium/issues/detail?id=1218159#c_ts1631714802

In this case we can fallback to the regular method before style attributes were special cased in #464, and which is still supported in replay.

**2)**: I realized we can switch between the #464 method and the regular attribute method for a given mutation depending on which is longer.  This prevents us converting the following:

```
elem.style = "background:black;"
```
to the much more verbose:

```
'style': {
        'background-color': 'black',
        'background-image': false,
        'background-position-x': false,
        'background-position-y': false,
        'background-size': false,
        'background-repeat-x': false,
        'background-repeat-y': false,
        'background-attachment': false,
        'background-origin': false,
        'background-clip': false
    }
```

Whilst still preserving those cases where the #464 method is more terse.  We achieve this by executing both methods and comparing string lengths of the two methods.
Comparison of the two methods upon emission is also the way we verify that the variable functions are being preserved.

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -562,7 +562,10 @@ export default class MutationBuffer {
           target.setAttribute('data-rr-is-password', 'true');
         }
 
-        if (attributeName === 'style') {
+        if (attributeName === 'style' &&
+            (value as string).indexOf('var(') === -1 &&
+            typeof item.attributes.style !== 'string'
+           ) {
           const old = unattachedDoc.createElement('span');
           if (m.oldValue) {
             old.setAttribute('style', m.oldValue);
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -18,7 +18,6 @@ import type {
   attributeCursor,
   removedNodeMutation,
   addedNodeMutation,
-  styleOMValue,
   Optional,
 } from '@rrweb/types';
 import {
@@ -441,18 +440,18 @@ export default class MutationBuffer {
         .map((attribute) => {
           const { attributes } = attribute;
           if (typeof attributes.style === 'string') {
-            const somAsStr = JSON.stringify(attribute.styleOM);
-            const unchangedAsStr = JSON.stringify(attribute.styleOMUnchanged);
-            // check if compact style mutation is actually shorter than string style mutation
-            // CSSOM fails badly when var() is present on shorthand properties, so ensure that there are the
-            // same number of variable strings present as compared with the style version before allowing it's use
-            if (
-              somAsStr.length < attributes.style.length &&
-              (somAsStr + unchangedAsStr).split('var(').length ===
-                attributes.style.split('var(').length
-            ) {
-              // use the compact style mutation format #464
-              attributes.style = attribute.styleOM;
+            const diffAsStr = JSON.stringify(attribute.styleDiff);
+            const unchangedAsStr = JSON.stringify(attribute._unchangedStyles);
+            // check if the style diff is actually shorter than the regular string based mutation
+            // (which was the whole point of #464 'compact style mutation')
+            if (diffAsStr.length < attributes.style.length) {
+              // also: CSSOM fails badly when var() is present on shorthand properties, so only proceed with
+              // the compact style mutation if these have all been accounted for
+              if ((diffAsStr + unchangedAsStr).split('var(').length ===
+                  attributes.style.split('var(').length
+                 ) {
+                attributes.style = attribute.styleDiff;
+              }
             }
           }
           return {
@@ -566,8 +565,8 @@ export default class MutationBuffer {
           item = {
             node: m.target,
             attributes: {},
-            styleOM: {},
-            styleOMUnchanged: {},
+            styleDiff: {},
+            _unchangedStyles: {},
           };
           this.attributes.push(item);
         }
@@ -603,19 +602,19 @@ export default class MutationBuffer {
                 newPriority !== old.style.getPropertyPriority(pname)
               ) {
                 if (newPriority === '') {
-                  item.styleOM[pname] = newValue;
+                  item.styleDiff[pname] = newValue;
                 } else {
-                  item.styleOM[pname] = [newValue, newPriority];
+                  item.styleDiff[pname] = [newValue, newPriority];
                 }
               } else {
                 // for checking
-                item.styleOMUnchanged[pname] = [newValue, newPriority];
+                item._unchangedStyles[pname] = [newValue, newPriority];
               }
             }
             for (const pname of Array.from(old.style)) {
               if (target.style.getPropertyValue(pname) === '') {
                 // "if not set, returns the empty string"
-                item.styleOM[pname] = false; // delete
+                item.styleDiff[pname] = false; // delete
               }
             }
           }
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -439,7 +439,7 @@ export default class MutationBuffer {
         .filter((text) => this.mirror.has(text.id)),
       attributes: this.attributes
         .map((attribute) => {
-          const attributes = attribute.attributes;
+          const { attributes } = attribute;
           if (typeof attributes.style === 'string') {
             const somAsStr = JSON.stringify(attribute.styleOM);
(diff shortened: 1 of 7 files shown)


packages/rrweb/src/record/mutation.ts
```diff
@@ -438,10 +438,26 @@ export default class MutationBuffer {
         // text mutation's id was not in the mirror map means the target node has been removed
         .filter((text) => this.mirror.has(text.id)),
       attributes: this.attributes
-        .map((attribute) => ({
-          id: this.mirror.getId(attribute.node),
-          attributes: attribute.attributes,
-        }))
+        .map((attribute) => {
+          const attributes = attribute.attributes;
```
Juice10, line 442: Object destructuring is neater here
```suggestion
          const { attributes } = attribute;
```

packages/rrweb/test/__snapshots__/integration.test.ts.snap
```diff
@@ -2933,24 +2933,14 @@ exports[`record integration tests can record node mutations 1`] = `
...
           }
         },
         {
           \\"id\\": 70,
           \\"attributes\\": {
-            \\"style\\": {
-              \\"display\\": false
-            }
+            \\"style\\": \\"\\"
           }
```
Juice10, line 2944: This looks like undesired behaviour

packages/rrweb/test/integration.test.ts
```diff
@@ -209,6 +209,55 @@ describe('record integration tests', function (this: ISuite) {
...
+    await page.goto('about:blank');
+    await page.setContent(getHtml.call(this, 'blank.html'), {
+      waitUntil: 'networkidle0',
+    });
+
+    // goal here is to ensure var(--mystery) ends up in the mutations (CSSOM fails in this case)
+    await page.evaluate(
+      'document.body.setAttribute("style", "background: var(--mystery)")',
+    );
+    await page.waitForTimeout(10);
```
Juice10, line 223: Wait for requestAnimationFrame is more reliable than using timeout
```suggestion
    await waitForRAF(page);
```

packages/rrweb/test/integration.test.ts
```diff
@@ -209,6 +209,55 @@ describe('record integration tests', function (this: ISuite) {
...
+    // whereas this case, it's shorter to record the entire string than the longhands for margin
+    await page.evaluate(
+      'document.body.setAttribute("style", "color:var(--mystery-color);display:block;margin:10px")',
+    );
+    await page.waitForTimeout(10);
+    // and in this case, it's shorter to record just the change to the longhand margin-left;
+    await page.evaluate(
+      'document.body.setAttribute("style", "color:var(--mystery-color);display:block;margin:10px 10px 10px 0px;")',
+    );
+    await page.waitForTimeout(10);
```
Juice10, line 249:
```suggestion
    await waitForRAF(page);

    await page.evaluate('document.body.setAttribute("style", "display:block")');
    await waitForRAF(page);
    // following should be recorded as an update of `{ color: 'var(--mystery-color)' }` without needing to include the display
    await page.evaluate(
      'document.body.setAttribute("style", "color:var(--mystery-color);display:block")',
    );
    await waitForRAF(page);
    // whereas this case, it's shorter to record the entire string than the longhands for margin
    await page.evaluate(
      'document.body.setAttribute("style", "color:var(--mystery-color);display:block;margin:10px")',
    );
    await waitForRAF(page);
    // and in this case, it's shorter to record just the change to the longhand margin-left;
    await page.evaluate(
      'document.body.setAttribute("style", "color:var(--mystery-color);display:block;margin:10px 10px 10px 0px;")',
    );
    await waitForRAF(page);
```

packages/rrweb/test/integration.test.ts
```diff
@@ -209,6 +209,55 @@ describe('record integration tests', function (this: ISuite) {
...
+    // see what happens when we manipulate the style object directly (expecting a compact mutation with just these two changes)
+    await page.evaluate(
+      'document.body.style.marginTop = 0; document.body.style.color = null',
+    );
+
+    const snapshots = (await page.evaluate(
+      'window.snapshots',
+    )) as eventWithTime[];
+    assertSnapshot(snapshots);
+  });
```
Juice10, line 259: We should make sure the snapshots have been updated
```suggestion
    await page.evaluate(
      'document.body.style.marginTop = 0; document.body.style.color = null',
    );
    await waitForRAF(page);

    const snapshots = (await page.evaluate(
      'window.snapshots',
    )) as eventWithTime[];
    assertSnapshot(snapshots);
  });
```

packages/types/src/index.ts
```diff
@@ -292,13 +292,15 @@ export type styleValueWithPriority = [string, string];
 export type attributeCursor = {
   node: Node;
   attributes: {
-    [key: string]: string | styleAttributeValue | null;
+    [key: string]: string | styleOMValue | null;
   };
+  styleOM: styleOMValue;
```
Juice10, line 297: Can we call this one `styleOMChanged` ?
I had to read through the code a couple times to understand what the differences are with `styleOMUnchanged` so maybe a slightly different name can help add clarity.

packages/rrweb/src/record/mutation.ts
```diff
@@ -438,10 +438,26 @@ export default class MutationBuffer {
...
+          if (typeof attributes.style === 'string') {
+            const somAsStr = JSON.stringify(attribute.styleOM);
+            const unchangedAsStr = JSON.stringify(attribute.styleOMUnchanged);
+            if (
+              somAsStr.length < attributes.style.length &&
+              (somAsStr + unchangedAsStr).split('var(').length ===
+                attributes.style.split('var(').length
+            ) {
+              // use the compact style mutation format #464
+              // (CSSOM fails badly when var() is present on shorthand properties)
```
Juice10, line 452: I was a little confused, think this might clarify things for future me
```suggestion
            // check if compact style mutation is actually shorter than string style mutation
            // CSSOM fails badly when var() is present on shorthand properties, check if this is the case here
            if (
              somAsStr.length < attributes.style.length &&
              (somAsStr + unchangedAsStr).split('var(').length ===
                attributes.style.split('var(').length
            ) {
              // use the compact style mutation format #464
```

Juice10 (commented):
Great stuff @eoghanmurray and @okeminja!

I had a few tweaks for clarity, and I've spotted a potential bug, but overall this PR looks great!

packages/rrweb/src/record/mutation.ts
eoghanmurray: +0
It's actually quite confusing as there is a parent property called `attributes` and each attribute has an `attributes` property, but that should be solved by renaming one of them so no objection to this change.

packages/rrweb/test/__snapshots__/integration.test.ts.snap
eoghanmurray: It's actually correct but well spotted as I did spend a lot of time investigating this one to ensure it was correct and thought I had made a note about it along with the commit message, but that seems to not have made it here.

Basically in the previous display:false, the `false` means 'remove this style attribute', and as there is only one attribute left ('display', it's returning the 'style' to have no attributes set, which is equivalent to the empty string.

Juice10: Aaah, ok that makes sense indeed, thanks for the explanation!

Juice10 (approved):

packages/rrweb/src/record/mutation.ts
```diff
@@ -438,10 +437,29 @@ export default class MutationBuffer {
...
-          id: this.mirror.getId(attribute.node),
-          attributes: attribute.attributes,
-        }))
+        .map((attribute) => {
+          const { attributes } = attribute;
+          if (typeof attributes.style === 'string') {
+            const diffAsStr = JSON.stringify(attribute.styleDiff);
+            const unchangedAsStr = JSON.stringify(attribute._unchangedStyles);
+            // check if the style diff is actually shorter than the regular string based mutation
+            // (which was the whole point of #464 'compact style mutation')
```
Juice10, line 446:
```suggestion
            // (which was the whole point of #464 'compact style mutation').
```

PR #1269, opened 2023-07-27
Mutation (attribute & text) duplicate info elimination

eoghanmurray:
Demonstrated a test case with duplicated information in the mutation and updated the mutation.emit() method to filter out the redundant info.


packages/rrweb/test/record.test.ts
```diff
@@ -651,6 +652,49 @@ describe('record', function (this: ISuite) {
...
+          div.parentNode?.removeChild(div as HTMLElement);
+        }
+      }, 15);
+      setTimeout(() => {
+        // 'unfreeze' happens upon a user event
+        // however, we expect none of the above mutations to produce any effect
+        document.body.click();
+      }, 20);
+    });
+    await ctx.page.waitForTimeout(50); // wait till setTimeout is called
```
Juice10, line 685: `waitForTimeout` can be pretty flaky. Other tests I've made more reliable by wrapping in a promise and returning that in the eval, making sure that all timing is handled inside the `eval` call.

Example:
```js
    await ctx.page.evaluate(() => {
      return new Promise((resolve) => {
       const { record, freezePage } = (window as unknown as IWindow).rrweb;
       record({
         emit: (window as unknown as IWindow).emit,
       });
       freezePage();
       setTimeout(() => {
         const div = document.createElement('div');
         div.setAttribute('id', 'here-and-gone');
         document.body.appendChild(div);
       }, 0);
       setTimeout(() => {
         const div = document.getElementById('here-and-gone');
         if (div) {
           div.setAttribute('data-test', 'x');
         }
       }, 10);
       setTimeout(() => {
         const div = document.getElementById('here-and-gone');
         if (div) {
           div.parentNode?.removeChild(div as HTMLElement);
         }
       }, 15);
       setTimeout(() => {
         // 'unfreeze' happens upon a user event
         // however, we expect none of the above mutations to produce any effect
         document.body.click();
       }, 20);
       setTimeout(resolve, 30);
       });
     });
```

eoghanmurray: I've made this change I think

PR #1271, opened 2023-07-30
fix(perf): improve perf of mutation and snapshot

JonasBa:
This PR builds on what @Juice10 had [found](https://github.com/rrweb-io/rrweb/pull/903) - I basically identified the same thing using different tools.

**Motivation:**
We started collecting [browser profiles](https://docs.sentry.io/platforms/javascript/profiling/) on some Sentry sites and noticed rrweb being the bottleneck when diffing large trees (common for loading screens which populate a large DOM tree). 

Example profile collected by the SDK from unknown arch (platform is macos):
image

Profile for matching route collected locally (arm64 macos)
image

By looking at the code, I managed to identify a couple of bottlenecks:
- isBlocking has a runtime of On^2 when checking for ancestors - I tried swapping that out for el.matches(.block *), I think it works faster, but I need to run benchmarks. The other approach could be to keep a cache of blocked elements and use [compareDocumentPosition](https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition), but that should be benchmarked before.
- There is some inefficiencies around Array.from(nodelist) when iterating over nodes and their children. This is just unnecessary as we are creating and discarding arrays where we could be using [NodeList.forEach](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach)
- genAdds is a recursive fn with possible deep recursion - I converted this to an iterative procedure. It causes the snapshot tests to fail as the procedure now runs in breadth first iteration (the data is there, just in different places and id's are different)
- some functions are just inefficient, they can either return early or avoid extra calls/allocations with simple modifications (example of refactoring to for loop vs doing .map().filter()). There are small changes like hoisting static regular expressions, calling .test instead of .match when discarding match groups and avoid creating a lot of new strings via lowerIfExists). Similar case with calling while(queue.length) queue.shift(). Since shift is On, it means the runtime for that is actually O(n^2)

**Benchmark results:**
```
master:#c6600e7
'append 70 x 70 x 70 elements' │   3   │ 1691.3333333333333 │ '1701, 1683, 1690'
'append 1000 elements and reverse their order' │   3   │   350    │ '342, 358, 350'
'real events recorded on bugs.chromium.org' │   3   │   3116   │ '3090, 3140, 3118'
'create 1000x10 DOM nodes' │ 'benchmark-dom-mutation.html' │ 'window.workload()' │  10   │  116.5   │ '114, 144, 119, 120, 111, 110, 127, 102, 112, 106'
'create 1000x10x2 DOM nodes and remove a bunch of them' │ 'benchmark-dom-mutation-add-and-remove.html' │ 'window.workload()' │  10   │  470.9   │ '526, 658, 455, 462, 415, 461, 415, 465, 441, 411'
'create 1000 DOM nodes and append into its previous looped node' │ 'benchmark-dom-mutation-multiple-descendant-add.html' │ 'window.workload()' │   5   │   69.2   │ '67, 62, 64, 89, 64'
'create 10000 DOM nodes and move it to new container' │ 'benchmark-dom-mutation-add-and-move.html' │ 'window.workload()' │   5   │  205.4   │ '208, 196, 214, 216, 193'
```

```
#bad990b
'append 70 x 70 x 70 elements' │   3   │ 1722.3333333333333 │ '1753, 1716, 1698'
'append 1000 elements and reverse their order' │   3   │ 358.6666666666667 │ '357, 341, 378'
'real events recorded on bugs.chromium.org' │   3   │ 3128.3333333333335 │ '3219, 3061, 3105
'create 1000x10 DOM nodes' │ 'benchmark-dom-mutation.html' │ 'window.workload()' │  10   │  121.2   │ '124, 168, 110, 130, 112, 107, 131, 115, 106, 109'
'create 1000x10x2 DOM nodes and remove a bunch of them' │ 'benchmark-dom-mutation-add-and-remove.html' │ 'window.workload()' │  10   │  196.6   │ '182, 201, 207, 167, 166, 210, 166, 170, 318, 179'
'create 1000 DOM nodes and append into its previous looped node' │ 'benchmark-dom-mutation-multiple-descendant-add.html' │ 'window.workload()' │   5   │   43.2   │ '44, 46, 48, 40, 38'
'create 10000 DOM nodes and move it to new container' │ 'benchmark-dom-mutation-add-and-move.html' │ 'window.workload()' │   5   │  223.6   │ '214, 225, 247, 220, 212'
```

Main performance improvement seems to be for the following cases (I would like to re-run these after review in case there are mistakes in my PR):
`create 1000x10x2 DOM nodes and remove a bunch of them` which decreased from ~500ms to ~200ms
`create 1000 DOM nodes and append into its previous looped node` which decreased from ~70ms to ~40ms

I suspect that by being a bit smarter about operations should also decrease memory pressure and avoid GC sweeps, but that is hard to quantify (basing this off my experience)

Feedback/critics on the PR are much appreciated. For what it's worth, I am not very familiar with rrweb codebase so take this with a grain of salt, I may have invalidated some assumptions the lib relies on and gotten invalid benchmark results. If I did, please point it out and I will address it. If someone is interested, feel free to cherry pick my changes as this PR has gotten quite large and risky. Hopefully it inspires someone to further improve the performance of operations (I am also happy to split it myself)

While this does address performance of the current implementation, I think the largest win in this case would be to process genAdds in an async loop so that we avoid blocking the main thread (I lack context on everything that might break?). Besides that, removing try/catch statements around hot code paths like isBlocking should be reevaluated as it is a common optimization bailout reason in v8 (are those really required?)


packages/rrweb/src/utils.ts
```diff
@@ -240,17 +242,14 @@ export function isBlocked(
 
   try {
     if (typeof blockClass === 'string') {
-      if (el.classList.contains(blockClass)) return true;
-      if (checkAncestors && el.closest('.' + blockClass) !== null) return true;
-    } else {
-      if (classMatchesRegex(el, blockClass, checkAncestors)) return true;
+      return el.classList.contains(blockClass) || (checkAncestors && el.matches(`.${blockClass} *`));
```
eoghanmurray, line 245: this is clever; but is slightly harder to understand — profiling would be good.
Also perhaps `el.matches(\`.${blockClass} ${el.tagName}\`)` could give you a slight speed increase and also be slightly clearer.

JonasBa: Yup, that makes sense. I had a simple benchmark with a similar approach and it seemed to be ~10% faster (I did not save it). I think results here will vary depending on the DOM structure though, but higher specificity should yield better results

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -52,13 +52,17 @@ function getValidTagName(element: HTMLElement): Lowercase<string> {
...
-        )
-        .join('')
-    : '';
+  if(sheet.cssRules.length === 0) {
+    return "";
+  }
+
+  const buffer: string[] = new Array<string>(sheet.cssRules.length);
+
+  for(let i = 0; i < sheet.cssRules.length; i++) {
```
eoghanmurray, line 61: "a live [CSSRuleList](https://developer.mozilla.org/en-US/docs/Web/API/CSSRuleList) which provides a real-time, up-to-date list of every CSS rule"

I'm wondering if `Array.from` is "more atomic" in converting the list to an array, whereas the manual iteration here could e.g. read the same item twice if an earlier rule is deleted during iteration?
If this is on the main thread then maybe that live-update is prevented?

JonasBa: Wouldn't that mean that it is less accurate today since it might miss that exact case?

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -147,7 +151,7 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
 
   function collectCharacters(regEx: RegExp) {
     let chars: string;
-    const match = regEx.exec(attributeValue.substring(pos));
```
eoghanmurray, line 150: apparently I'm the author of this line although I didn't remember as it's a copy of https://github.com/albell/parse-srcset/blob/master/src/parse-srcset.js

According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec, the `exec` here (unlike `match`) maintains some state, so I'd like to run the tests on that other repository with this change in place and see if anything breaks before going with this change.

packages/rrweb-snapshot/src/snapshot.ts
JonasBa: @eoghanmurray I could store sheet.cssRules.length in a variable such that new rules being added would not be parsed, but replaced rules could still have their new values reflected in the end.

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -823,24 +842,40 @@ function serializeElementNode(
...
-  } else if (sn.type === NodeType.Element) {
+  }
+  
+  if (sn.type === NodeType.Element) {
+    /* eslint-disable */
+    const snAttributeName: string = sn.attributes.name
+      ? // @ts-ignore
+        sn.attributes.name.toLowerCase()
+      : '';
+    const snAttributeRel: string = sn.attributes.name
```
eoghanmurray, line 869: this should be `sn.attributes.rel` (and next line)

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -823,24 +842,40 @@ function serializeElementNode(
...
+
+  if (sn.type === NodeType.Comment && slimDOMOptions.comment) {
     // TODO: convert IE conditional comments to real nodes
     return true;
-  } else if (sn.type === NodeType.Element) {
+  }
+  
+  if (sn.type === NodeType.Element) {
+    /* eslint-disable */
+    const snAttributeName: string = sn.attributes.name
```
eoghanmurray, line 865: having to write this inlined code to list what you will be checking further down kind of mystifies things, and discourages casual additions, but otherwise this is sound from a performance pov

packages/rrweb/src/record/mutation.ts
```diff
@@ -426,24 +431,36 @@ export default class MutationBuffer {
...
+      }
+      texts.push({
+        id: this.mirror.getId(this.texts[i].node),
+        value: this.texts[i].value,
+      });
+    }
+
+    const attributes = [];
+    for(let i = 0; i < this.attributes.length; i++){
+      if(!this.mirror.getId(this.attributes[i].node)){
```
eoghanmurray, line 450: previous version doesn't call `getId` twice; should be easy to fix with an extra let?

eoghanmurray: (also for the this.texts loop)

JonasBa: Ah yes, explains the missmatched IDs in test failures!

packages/rrweb-snapshot/src/snapshot.ts
JonasBa: I was too eager here, removing

packages/rrweb-snapshot/src/snapshot.ts
JonasBa: 🤦🏼

packages/rrweb-snapshot/src/snapshot.ts
JonasBa: @eoghanmurray this might be easier to write if we used sets. For example 
```
slimDOMOptions.headMetaRobots &&
        (snAttributeName === 'robots' ||
          snAttributeName === 'googlebot' ||
          snAttributeName === 'bingbot')
```

would just become 
```
slimDOMOptions.headMetaRobots && set.has(lowerIfExists(sn.attributes.name))
```
With the added benefit that we defer lowerIfExists until that branch is hit

JonasBa: Tbh I would expect these strings to be fairly short and not the bottleneck, so whatever is easier to read (provided we are smarter about lowerIfExists)

JonasBa:
I re-ran the benchmark after [5d59621](https://github.com/rrweb-io/rrweb/pull/1271/commits/5d59621f3a0d10c988b19d221db6cf6bfad9b9f2)

```
append 70 x 70 x 70 elements' │   3   │ 1669.3333333333333 │ '1688, 1666, 1654
append 1000 elements and reverse their order' │   3   │   343    │ '347, 343, 339
real events recorded on bugs.chromium.org' │   3   │ 3020.6666666666665 │ '3128, 2970, 2964
create 1000x10 DOM nodes' │ 'benchmark-dom-mutation.html' │ 'window.workload()' │  10   │  117.6   │ '115, 120, 133, 112, 119, 112, 121, 114, 109, 121
create 1000x10x2 DOM nodes and remove a bunch of them' │ 'benchmark-dom-mutation-add-and-remove.html' │ 'window.workload()' │  10   │   179    │ '187, 209, 163, 165, 201, 163, 163, 211, 164, 164
create 1000 DOM nodes and append into its previous looped node' │ 'benchmark-dom-mutation-multiple-descendant-add.html' │ 'window.workload()' │   5   │   40.6   │ '41, 42, 44, 38, 38
create 10000 DOM nodes and move it to new container' │ 'benchmark-dom-mutation-add-and-move.html' │ 'window.workload()' │   5   │  218.8   │ '210, 217, 236, 219, 212
```

The timings are consistent and we can still see a boost for create + remove and create + append into previous looped node

JonasBa:
Actually @eoghanmurray, running the [benchmark](https://jsbench.me/7ilkpmi7rn/1) here and testing matches with the .tagName combination and it seems to be even slower. I've looked at the blink implementation of matches (just because I was curious) and it seems that each individual selector is parsed and tested in a left to right order. I suspect the dip here is because the engine can short circuit the search if the selector it first checks after .block is a * selector as anything will match

eoghanmurray:
@Juice10 is gonna get the performance/benchmark tests in main shortly so will hold off on committing #1272 until we can do a nice demonstration of that in action, and hopefully apply the same to this PR

JonasBa:
@eoghanmurray makes sense, just keep me updated and let me know how I can help further. 

On a small side note, I looked into the blink implementation of matches and it confirms the difference between closest and matches implementations - closest applies the same check as ::matches, but it does so for each parent element. I suspect that `*` selector could be faster as the engine has to only compare the parent -> child relationship between nodes and not match against the actual selector as * will match anything.

```diff
bool SelectorDataList::matches(Element& targetElement) const
{
    if (m_needsUpdatedDistribution)
        targetElement.updateDistribution();

    unsigned selectorCount = m_selectors.size();
    for (unsigned i = 0; i < selectorCount; ++i) {
        if (selectorMatches(*m_selectors[i], targetElement, targetElement))
            return true;
    }

    return false;
}

Element* SelectorDataList::closest(Element& targetElement) const
{
    if (m_needsUpdatedDistribution)
        targetElement.updateDistribution();

    unsigned selectorCount = m_selectors.size();
+    for (Element* currentElement = &targetElement; currentElement; currentElement = currentElement->parentElement()) {
        for (unsigned i = 0; i < selectorCount; ++i) {
            if (selectorMatches(*m_selectors[i], *currentElement, targetElement))
                return currentElement;
        }
+    }
    return nullptr;
}
```

eoghanmurray:
> closest applies the same check as ::matches, but it does so for each parent element

Ya but in this case you are also modifying the selector so it's not a straight comparison... but `closest` does do an unnecessary check on the element itself (we've already checked that via the classList check, so I wonder if an optimisation to remove the classList check would work just as well)

> I suspect that `*` selector could be faster as the engine has to only compare the parent -> child relationship between nodes and not match against the actual selector as * will match anything.

Performance guestimation is always counter-intuitive to me!

eoghanmurray:
I've merged a recent PR that is now on main

JonasBa:
@eoghanmurray would you like me to rerun the benchmarks?

JonasBa:
> Ya but in this case you are also modifying the selector so it's not a straight comparison... but `closest` does do an unnecessary check on the element itself (we've already checked that via the classList check, so I wonder if an optimisation to remove the classList check would work just as well)

I updated the benchmark https://jsbench.me/7ilkpmi7rn/1 - it seems that matches is still faster (benchmark is pushing to an array in order to escape optimizations)

JonasBa:
It seems that the changes from master broke lint rules

eoghanmurray:
> It seems that the changes from master broke lint rules

cannot find name `addedIds`

I did the merge conflict resolution in the github editor and messed it up!

JonasBa:
I'll reset and rebase this correctly, no worries

eoghanmurray:
Hi again Jonas, would it be possible to refactor out the pushAdds change from recursive to iterative into it's own PR?
I'd like to get another core team member reviewing it in isolation!

JonasBa:
Yes, not a problem

JonasBa:
@eoghanmurray I reverted the genAdds change and updated snapshots, lmk if you need me to change something else

JonasBa:
@eoghanmurray something I just noticed is that the SDK has a default blockClass option set to "rr-block". I'm not sure why that is, maybe you have an idea? Unless there are external users outside of the SDK relying on this behavior, then I think this should be an opt-in behavior with no default value. The consequence of the default value is that we are performing a lot of isBlocked calls on each mutation when the page might not even contain any rr-block elements... :/

eoghanmurray:
Ya this is so we can tell people to 'add `rr-block` class to anything you want blocked' without having to also have to explain how to modify their config (or redeploy to a customer).  I don't think this can be changed now unless in a major version.  We could provide a recommended installation which configures `blockClass: false, // performance: omit default check class of 'rr-block'`  but this would need to be a recipe for backwards compatibility (IMO)

JonasBa:
> Ya this is so we can tell people to 'add `rr-block` class to anything you want blocked' without having to also have to explain how to modify their config (or redeploy to a customer). I don't think this can be changed now unless in a major version. We could provide a recommended installation which configures `blockClass: false, // performance: omit default check class of 'rr-block'` but this would need to be a recipe for backwards compatibility (IMO)

@eoghanmurray yes, that makes sense. Imo a recommendation is fine since this is definitely in the hot path

JonasBa:
@eoghanmurray friendly ping :)

JonasBa:
@eoghanmurray any chance we could get a pair of eyes on this PR?

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -262,6 +262,10 @@ export function _isBlockedElement(
   blockClass: string | RegExp,
   blockSelector: string | null,
 ): boolean {
+  if (!blockClass && !blockSelector) {
+    return false;
+  }
```
pauldambra, line 267: 🙌

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -850,33 +881,30 @@ function slimDOMExcluded(
...
-        lowerIfExists(sn.attributes.name).match(/^description|keywords$/)
+        (snAttributeName === 'description' || snAttributeName === 'keywords')
       ) {
         return true;
       } else if (
         slimDOMOptions.headMetaSocial &&
-        (lowerIfExists(sn.attributes.property).match(/^(og|twitter|fb):/) || // og = opengraph (facebook)
-          lowerIfExists(sn.attributes.name).match(/^(og|twitter):/) ||
-          lowerIfExists(sn.attributes.name) === 'pinterest')
+        (OG_TWITTER_OR_FB_REGEXP.test(snAttributeProperty) || // og = opengraph (facebook)
```
pauldambra, line 899: very tiny nitpick: that the comment makes more sense up at the declaration now 🙈

marandaneto:
> @eoghanmurray any chance we could get a pair of eyes on this PR?

Maybe @Juice10 can help here as I've seen their activity on other PRs recently.

pauldambra (approved):
fly-by approval fwiw since I'd love to see this land 🚀 ❤️

Yuyz0112:
Is there any chance @JonasBa  can split these cool optimizations into multiple PRs? So some small tweaks should be merged more quickly.

JonasBa:
> Is there any chance @JonasBa can split these cool optimizations into multiple PRs? So some small tweaks should be merged more quickly.
> 
> Since I found there are already some conflicts in this PR(sorry for the late review).

Late to this, but I'm starting to do that yes.

JonasBa:
Closing this in favor of the smaller PRs that I opened

PR #1272, opened 2023-08-01
Perf: Avoid creation of intermediary array when iterating over style rules

eoghanmurray:
Discovered there's a second `mapFn` argument of Array.from

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from


JonasBa (commented):
🥇

Juice10:
I'm happy to merge this once we have benchmarks running on CI

JonasBa:
@Juice10 If you plan on doing that in github actions just note that performance may vary quite a lot there. We (sentry) collect performance data and profiles from a lot of our CI runs and we see numbers vary about 30% for the same runners.

Juice10:
@JonasBa I'm hoping we can do relative benchmarking (more info https://labs.quansight.org/blog/2021/08/github-actions-benchmarks ) to help with this. But it's definitely a valid concern. 

I'm not sure if there are any companies like https://buildjet.com that offer dedicated resources for benchmarking.

JonasBa:
> @JonasBa I'm hoping we can do relative benchmarking (more info https://labs.quansight.org/blog/2021/08/github-actions-benchmarks ) to help with this. But it's definitely a valid concern.
> 
> I'm not sure if there are any companies like https://buildjet.com that offer dedicated resources for benchmarking.

@Juice10 yes, that is a good approach if you plan on using github's runners as your benchmarking infra. I would probably invest into seeing if it's possible to have an automated workflow that re-runs the benchmarks a few times so that they hopefully run on different runners - this could help improve the signal to noise ratio.

Juice10 (approved):

eoghanmurray:
To note: MooTools versions less than 1.6 add an `Array.from` to the Array prototype, which doesn't have the second argument; they symptoms are

```
<style>
[object CSSStyleRule][object CSSStyleRule][object CSSStyleRule][object CSSStyleRule][object CSSStyleRule][object CSSStyleRule][object CSSStyleRule][object CSSStyleRule][object CSSStyleRule][object CSSStyleRule][object CSSS
```

I don't know if we should guard against this similar to #1196 ...

PR #1273, opened 2023-08-01
Canvas recording: Preserve drawing buffer 

Juice10:
Sometimes canvas isn't being recorded because the drawing buffer was sent to the graphics card and then emptied right away, and any samples we do from the canvas ends up just being empty/transparent. This is especially true for canvas that are mostly static (things like maps, graphs and drawings).

This PR sets `preserveDrawingBuffer: true` for canvas being recorded via the FPS method (`sampling: { canvas: number }`).

Canvas replay was also sometimes flickering due to canvas commands not being run immediately after each other. This was due to async arguments taking a while to complete. And is fixed in this PR.

I was having some issues with CI's Code Style Check always returning a different yarn.lock, so I've tweaked CI to not do that anymore (`--frozen-lockfile`).

I also needed to upgrade jest + puppeteer to work with M1 Macs.

- [x] Upgrade jest to 29
- [x] Upgrade puppeteer to 20 in rrweb
- [x] set `preserveDrawingBuffer: true` on by default for recordings with `{ recordCanvas: true,   sampling: { canvas: 15 }}`


packages/rrweb/test/utils.ts
```diff
@@ -192,6 +194,27 @@ function stringifySnapshots(snapshots: eventWithTime[]): string {
...
+          ).data.payload;
+
+          if (pluginPayload?.trace.length) {
+            pluginPayload.trace = pluginPayload.trace.map((trace) => {
+              return trace.replace(/^pptr:evaluate;.*%2F/, 'pptr:evaluate;');
+            });
+          }
+          if (pluginPayload?.payload.length) {
+            pluginPayload.payload = pluginPayload.payload.map((payload) => {
+              return payload.replace(/pptr:evaluate;.*%2F/g, 'pptr:evaluate;');
```
YunFeng0817, line 215: Once we improve this regex, I think this pull request will be good to go.

YunFeng0817 (approved):

packages/rrweb/test/utils.ts
Juice10: Done!

PR #1276, opened 2023-08-07
ref(test): Filter out scroll from snapshots

billyvg:
In addition to mouse move and viewport resize events, filter out scroll events from snapshots. These seem to flake a lot locally.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5503,9 +5503,14 @@ csso@^4.0.2:
   dependencies:
     css-tree "^1.1.2"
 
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -5503,10 +5503,15 @@ csso@^4.0.2:
   dependencies:
     css-tree "^1.1.2"
 
-cssom@^0.4.4, cssom@^0.5.0, "cssom@https://registry.npmjs.org/rrweb-cssom/-/rrweb-cssom-0.6.0.tgz":
+cssom@^0.4.4, "cssom@https://registry.npmjs.org/rrweb-cssom/-/rrweb-cssom-0.6.0.tgz":
```
billyvg, line 5506: CI did this

PR #1278, opened 2023-08-08
Added requestly in who's using rrweb section

wrongsahil:


YunFeng0817 (approved):
Thank you for contributing and sponsoring rrweb!

Issue #1249, opened 2023-07-05
[Bug]: css capture that doesn't display in all browsers


pauldambra:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

2.0.0-alpha.4

### Expected Behavior

That captured CSS should display the same when applied to an element in a browser as it did when captured from it

### Actual Behavior

A customer of ours (PostHog) is using some CSS to make titles on their page pretty

You can see the setup in this jsfiddle https://jsfiddle.net/pauldambra/k4rqm593/41/

In safari and firefox the CSS as captured by rrweb displays correctly when applied to an element in the fiddle. But in Chrome the captured CSS causes the text to appear to be invisible.

![Screenshot 2023-07-05 at 15 44 37](https://github.com/rrweb-io/rrweb/assets/984817/2442395f-221e-45f3-8995-aec433b66ec1)

I'm not certain this is a bug in rrweb to be honest. But would be great to get advice on if there's a workaround we can apply or suggest to the customer if not.

In the fiddle you can see that the CSS that is applied is

```
  background: linear-gradient(127.44deg, #C63A3A 32.29%, #3A2828 100%);
  -webkit-background-clip: text;
  -webkit-text-fill-color: transparent;
  background-clip: text;
  text-fill-color: transparent;
```

rrweb captures that as

```
  background: linear-gradient(127.44deg, rgb(198, 58, 58) 32.29%, rgb(58, 40, 40) 100%) text;
  -webkit-text-fill-color: transparent;
```

That captured CSS works in Safari and Firefox but not in Chrome

### Steps to Reproduce

see https://jsfiddle.net/pauldambra/k4rqm593/41/

Juice10:
Do you know what browser was used to capture this?

eoghanmurray:
As OP mentioned, the jsfiddle works fine in browsers apart from Chrome, so it is a Chrome bug.

For the example of the inline <style>, the problem is in snapshot.ts where this happens:

```
if ((n.parentNode as HTMLStyleElement).sheet?.cssRules) {
        textContent = stringifyStyleSheet(
          (n.parentNode as HTMLStyleElement).sheet!,
        );
 }
 ```

The textContent is correct:
```
> textContent
[...]

PR #1279, opened 2023-08-08
Extend to run fixBrowserCompatibilityIssuesInCSS over inline stylesheets

eoghanmurray:
@Juice10 a small PR assuming it's correct to also run inline <style> elements over the `fixBrowserCompatibilityIssuesInCSS` function that was added in #1047 

This is a starting point for fixing #1249


Juice10 (approved):
Good catch

PR #1280, opened 2023-08-08
Perf: don't run the regex replace unless the selectorText contains a colon

eoghanmurray:
(rules generally contain colons)

This is an imagined performance improvement on #1253 so should be measured!


packages/rrweb-snapshot/src/utils.ts
```diff
@@ -75,25 +75,27 @@ export function getCssRuleString(rule: CSSRule): string {
     } catch {
       // ignore
     }
+  } else if (isCSSStyleRule(rule) && rule.selectorText.includes(':')) {
+    // Safari does not escape selectors with : properly
```
eoghanmurray, line 79:
```suggestion
    // Safari does not escape selectors with : properly
    // see https://bugs.webkit.org/show_bug.cgi?id=184604    
```

mydea:
Looks good to me - I also had a hard time testing this locally, but this should be good I think (don't have any setup right now to verify this, but the existing test _should_ cover this hopefully?)

eoghanmurray:
Checks are failing as I authored it with #1279 in place, so that should be considered a prerequisite of this

Juice10 (approved):

Juice10:
@eoghanmurray could you rebase this PR, I just merged in #1279

PR #1288, opened 2023-08-14
Make sure CI doesn't trigger updates to yarn.lock files

Juice10:
Currently we have a super annoying yarn bug + double GitHub actions run that cause conflicts and cause every code format run to fail. This should fix that issue


YunFeng0817 (approved):
Looks good

PR #1294, opened 2023-08-30
perf: only call createHTMLDocument when it is needed

mdellanoce:
I noticed on a very large mutation that createHTMLDocument was a significant portion of the processing time. It appears that `unnattachedDoc` is only used in one specific branch, so rather than create it every time `processMutation` is called, I moved it so that it is only created in the branch where it is used.

This improves a couple of the benchmarks, while the others appear to be unaffected either way:

Before:
image
image

After:
image
image

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -497,14 +497,6 @@ export default class MutationBuffer {
     if (isIgnored(m.target, this.mirror)) {
       return;
     }
-    let unattachedDoc;
-    try {
-      // avoid upsetting original document from a Content Security point of view
-      unattachedDoc = document.implementation.createHTMLDocument();
-    } catch (e) {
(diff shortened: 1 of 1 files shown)


kwalker3690:
@YunFeng0817 @Juice10 can this PR be approved/merged? We just ran into this performance improvement as well and would really like to incorporate it into our use case

eoghanmurray:
I'm gonna do a slight refactor so that the document is shared once created

eoghanmurray:
Here is my suggested change:
https://github.com/eoghanmurray/rrweb/commit/1ad42a7f331ab9409bdc21dc1b33ae77d46d10f6
(I don't know how to push that as a new commit on top of this PR)

mdellanoce:
@eoghanmurray just close this and open a new PR, i guess

mdellanoce:
closing due to duplication with #1321

PR #1295, opened 2023-08-30
Return early for child same origin frames

colingm:
If we have cross origin record turned on but we are in a child frame that has the same origin as its parent we end up in sort of an inefficient state. We will start the recording, record mutations, but then never actually emit them anywhere since inEmittingFrame and passEmitsToParent are both false. This is a waste of resources and we might as well just never start the recording.

Brought this up in slack: https://rrweb.slack.com/archives/C01BYDC5C93/p1692810269238919

### DIFF
--- a/packages/rrweb/src/record/index.ts
+++ b/packages/rrweb/src/record/index.ts
@@ -114,6 +114,9 @@ function record<T = eventWithTime>(
   if (inEmittingFrame && !emit) {
     throw new Error('emit function is required');
   }
+  if (!inEmittingFrame && !passEmitsToParent) {
+    return () => {  /* no-op since in this case we don't need to record anything from this frame in particular */ };
+  }
   // move departed options to new options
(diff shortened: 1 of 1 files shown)


colingm:
@Juice10 any chance you can take a look here as you had context on the original slack question?

YunFeng0817:
@colingm Can you repost or screenshot the original issue here? The message in the slack channel is no longer available to us.
This PR actually looks good to me. I'm willing to merge this in.

YunFeng0817 (approved):

colingm:
@YunFeng0817 yeah I don't have access to the exact text but my first comment/description for the PR explain the same thing I explained in slack. In slack I just brought up that idea and @Juice10 said yeah that it is kind of wasteful to keep same origin child frames running if you have cross origin recording turned on

Juice10 (approved):

PR #1277, opened 2023-08-07
perf(genAdds): refactor recursive procedure to iterative

JonasBa:
Not sure how representative the benchmark is, so I'm adding the screenshot of the profile (deep DOM tree benchmark). Looking at timings, it seems to take ~18ms before and 13ms after my change, or about ~30% gain. Feel free to do more extensive benchmarking and take those timings with a grain of salt as I just ran the benchmarks once

Before
CleanShot 2023-08-08 at 10 11 25@2x

After
CleanShot 2023-08-08 at 10 11 36@2x

We can see the stack is shallower as expected (and uses less memory, sadly no metrics on that).
A small benefit is that we also remove the overhead of calling into the polymorphic genAdds fn.

The queue has a magic default size of 1000 - feel free to tweak that to some better value, I'm not sure what a sensible default could be here

PR #1302, opened 2023-09-09
perf(rrweb): optimize random shuffled addList

wfk007:
ref: https://github.com/rrweb-io/rrweb/pull/1300

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -699,7 +699,13 @@ export default class MutationBuffer {
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
 
     // if n is added to set, there is no need to travel it and its' children again
-    if (this.addedSet.has(n) || this.movedSet.has(n)) return;
+    if (this.addedSet.has(n)) {
+      // update the order of added nodes in addedSet for the `emit` phase
+      this.addedSet.delete(n);
+      this.addedSet.add(n);
+      return;
+    }
+    if (this.movedSet.has(n)) return;
 
     if (this.mirror.hasNode(n)) {
       if (isIgnored(n, this.mirror)) {


packages/rrweb/src/record/mutation.ts
```diff
@@ -699,7 +699,13 @@ export default class MutationBuffer {
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
 
     // if n is added to set, there is no need to travel it and its' children again
-    if (this.addedSet.has(n) || this.movedSet.has(n)) return;
+    if (this.addedSet.has(n)) {
+      // update the order of added nodes in addedSet for the `emit` phase
+      this.addedSet.delete(n);
+      this.addedSet.add(n);
+      return;
```
eoghanmurray, line 706: So it looks like the order of the addedSet has a big effect on performance; we should acknowledge this and change the data structure to e.g. an array or List, as a browser could implement the Set data structure differently, or a future compiler could recognise that logically add + delete is a 'null' operation, and could optimise by removing these two lines ... so I'd recommend exploring whether changing `addedSet` to a list data structure would make sense

eoghanmurray: Actually I was mistaken and the Javascript Set object iterates over things in insertion order:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
So you can continue to use it

mdellanoce:
@wfk007 it unfortunately looks like there are some performance impacts for this change in certain situations. I recently encountered a slow mutation with an old version of the jstree library. Here's a [plunker](https://plnkr.co/edit/6fO3swxUA88Oj2Yu) that demonstrates the issue. There's 2 rrweb builds in there, one with the "shuffle fix" and one without.

Here's the profile for the "without" build:
image

Here's the profile for the "with" build:
image

Looks like this particular mutation spends a lot more time in that `while(addList)` loop with the fix applied, but I'm not sure why.

Sidenote: by applying a couple more fixes that are in PR limbo, I got the processing time for the mutation in the plunker down to just over 100ms. Some of the proprietary recording solutions out there can handle it in 50ms...

wfk007:
@mdellanoce I considered that this PR was not the optimal solution, and the implementation was not very elegant, so I changed it to a draft. When I am not busy later, I can also take a look at this case. Thanks for your feedback.

mdellanoce:
another possibility for a fix:

`diff --git a/packages/rrweb/src/record/mutation.ts b/packages/rrweb/src/record/mutation.ts
index 097d1a8..cdceaf5 100644
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -425,7 +425,13 @@ export default class MutationBuffer {
         }
         break;
       }
-      candidate = node.previous;
+      if (node.value.previousSibling && isNodeInLinkedList(node.value.previousSibling)) {
+        candidate = node.value.previousSibling.__ln;
+      } else if (node.value.parentNode && isNodeInLinkedList(node.value.parentNode)) {
+        candidate = node.value.parentNode.__ln;
+      } else {
+        candidate = node.previous;
+      }
       addList.removeNode(node.value);
       pushAdd(node.value);
     }
`

Tries a little harder to find nodes that might have a serialized parent/sibling. Fixes the shuffle issue and performs well for the jstree issue as well. Unfortunately, it does slow down the "_create 1000 DOM nodes and append into its previous looped node_" benchmark. Feels like the worst case just keeps changing depending on the fix. The original "defragment" change in #1300 is the most general purpose I've found so far, but obviously has the downside of slowing down the average case.

wfk007:
@mdellanoce I know why "without shuffle fix"  is faster than "with".

image

The above Dom structure can be created in many ways.

1. Add `div1` as a whole to body
```js
const div1 = document.createElement('div');
div1.setAttribute('id', 'div1')
const div2 = document.createElement('div');
div2.setAttribute('id', 'div2')
        
div1.append(div2)
document.body.append(div1)
```

1. Add `div1` to body and add `div2` to `div1` separately
```js
const div1 = document.createElement('div');
div1.setAttribute('id', 'div1')
document.body.append(div1)

const div2 = document.createElement('div');
div2.setAttribute('id', 'div2')
div1.append(div2)
```

These two approach will generate different MutationObserver by browser

1. means: div add to body
```js
[
  {
    target: body,
    addNodes: [div1],
    type: "childList"
  }
]
```

2. means: div1 add to body and div2 add to div1
```js
[
  {
    target: body,
    addNodes: [div1],
    type: "childList"
  },
  {
    target: div1,
    addNodes: [div2],
    type: "childList"
  },
]
```

In order to comprehensively consider the above two cases, when looping `addNodes`,  we will travel all its descendant children. It means that in case two, we will read div2 twice(one in index===0 and another in index===1), and the right order will be broken in the second loop.

The order will directly affect the performance in the emit phase.

mdellanoce:
> Ensure the performance of emit even if addedSet is not in the right order: It is possible to increase the average time consumption when optimizing the performance of boundary cases.

@wfk007  I've been (slowly) working at it from this angle. I'm trying to ensure linear time processing for added nodes on both the record and replay sides. I hope to have something to share soon.

mdellanoce:
i updated my PR #1300 with another potential solution

mdellanoce:
i updated #1300 with a benchmark for the 2nd style of mutation mentioned above

eoghanmurray:
So just wondering if anyone has compared this with #1277 and whether they are competing solutions or can work in tandem?  I'm just trying to chase up over there and tie everything together

Issue #1218, opened 2023-05-02
[Bug]: inlineImages not working


econte-sprig:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

2.0.0-alpha.4

### Expected Behavior

I expect the record function to emit events with the images as base64 rr_dataURL attributes. 

### Actual Behavior

Events are emitted with the normal img src attribute unchanged and no rr_dataURL attribute. No warnings or errors in the console.

### Steps to Reproduce

Using Chrome Version 112.0.5615.137 (Official Build) (arm64)
Begin recording with the setting inlineImages: true, and checkoutEveryNms of some value (say 10000).
Look at the events being emitted, you will notice that they do not have the rr_dataURL attribute

### Additional Information

You will notice in the Chrome network tab that every 10 seconds, the browser refetches the images on the page. See the following screenshot that traces the image fetch to rrweb-snapshot.
<img width="1096" alt="image" src="https://user-images.githubusercontent.com/104794951/235579120-62bbe127-2983-4291-a2a7-e14a89a0d248.png">

That line is the following:
<img width="646" alt="image" src="https://user-images.githubusercontent.com/104794951/235579187-3e09b122-ba35-4617-97d2-79b2a99c1f41.png">

Setting the crossOrigin attribute on the img element is causing it to re-load the image. Which then leads to a bug. The condition on line 676 `if (image_1.complete && image_1.naturalWidth !== 0)` is _always false_, because setting crossOrigin causes the image to be reloaded.

Furthermore, I don't understand the point of setting the onload event handler. The function serializeElementNode is _not_ asynchronous, by the time the handler is called, the function has already returned the attributes object.

econte-sprig:
Building with a locally modified version of rrweb I have tried deleting the following lines:
```image_1.crossOrigin = 'anonymous';```
and
```oldValue_1
                ? (attributes.crossOrigin = oldValue_1)
                : image_1.removeAttribute('crossorigin');```
and
```else
            image_1.onload = recordInlineImage;```

And then inline images works as expected, generating outputs with the `rr_dataURL` attribute as expected

wfk007:
@econte-sprig Thanks for your PR and detailed explanation, but your change is not working with the following case, and it explains why we need to set the `onload` event handler.

```html
<html>
    <body>
[...]

PR #1306, opened 2023-09-14
Fix for inline images

econte-sprig:
https://github.com/rrweb-io/rrweb/issues/1218


eoghanmurray:
@wfk007 would you be free to have a call next week with core team members on this one, and perhaps some other PRs that you have contributed to?  We can make arrangements through Slack.

eoghanmurray:
I would vote to defer this PR until we can see if Justin can get #1239 complete which should solve the same issue?

Issue #1308, opened 2023-09-23
[Feature Request]: Enable preserveSource option of rollup-plugin-web-worker-loader


KentoMoriwaki:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a feature request that matches the one I want to file, without success.

### What package is this feature request for?

rrweb

### Problem Description

I am developing a Chrome Extension using rrweb (2.0.0-alpha.4), but it has been rejected during the review process for the following reason:

[Chrome Web Store Policy - Obfuscation](https://developer.chrome.com/docs/webstore/troubleshooting/#obfuscation)

> The following are some examples of violations of this type of policy:
> - Base 64 encoding (e.g. 'SSdtIGluIHVyIGJhc2U=')

Upon investigation, I found that the base64 encoded worker code, generated as a result of using rollup-plugin-web-worker-loader within rrweb, was the cause of the rejection.

### Proposed Solution

By using the setting `preserveSource: true` [here](https://github.com/rrweb-io/rrweb/blob/58c9104eddc8b7994a067a97daae5684e42f892f/packages/rrweb/rollup.config.js#L127),  it’s possible to create a worker without having to rely on base64 encoding. Unless there is a specific reason not to, I would like to propose setting this option to true by default to avoid similar issues in the future.

### Alternatives Considered

If it is not possible to change `preserveSource` for some reasons, I suggest considering outputting the worker as a separate file.

PR #1309, opened 2023-09-23
Enable preserveSource of rollup-plugin-web-worker-loader

KentoMoriwaki:
Ref: https://github.com/rrweb-io/rrweb/issues/1308

This change modifies the build output `` as detailed below:

### Before 

```
var WorkerFactory = createBase64WorkerFactory('Lyogcm9sbHVwLXBsdWdpbi13Z...', null, false);
```

### After

```
var WorkerFactory = createInlineWorkerFactory(/* rollup-plugin-web-worker-loader */function () {
(function () {
    '__worker_loader_strict__';

   --- SOURCE CODE HERE ---

})();
}, null);
```

### DIFF
--- a/packages/rrweb/rollup.config.js
+++ b/packages/rrweb/rollup.config.js
@@ -125,6 +125,7 @@ function getPlugins(options = {}) {
     webWorkerLoader({
       targetPlatform: 'browser',
       inline: true,
+      preserveSource: true,
       sourceMap,
     }),
     esbuild({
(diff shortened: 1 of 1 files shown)


KentoMoriwaki:
I tried using this build locally to record canvas, but it did not work correctly (the initial drawing is saved, but the modifications are not).
Interestingly, when I disabled minification in Vite for my application, the canvas recording functioned correctly.

It seems this change has the potential to break existing applications, and it might be necessary to alert users to modify their settings accordingly.

eoghanmurray (approved):
As per Justin, it wasn't deliberate that it was base64 encoded

daibhin:
Can confirm that this is breaking in production since upgrading recently

PR #1310, opened 2023-09-26
Extended text masking function to include relevant HTMLElement

benjackwhite:
A while back [I added](https://github.com/rrweb-io/rrweb/pull/1188) the element-to-be-masked to the masking function for inputs. This offers ultimate flexibility as the implementor can re-implement any of the other masking checks or any other thing they can come up with.

What was missing was the corresponding thing for text. This PR adds it.

For example it can solve the request in [this issue](https://github.com/rrweb-io/rrweb/issues/1096) without needing to add more, conflicting options.
```
{
    maskAllInputs: true,
    maskInputFn: (text, element) => {
        if (element?.dataset['rr-unmask'] === 'true') {
            return text
        }
        return '*'.repeat(text.length)
    },
    maskTextSelector: "*",
    maskTextFn: (text, element) => {
        if (element?.dataset['rr-unmask'] === 'true') {
            return text
        }
        return '*'.repeat(text.length)
    },
}
```

With this example, everything will be masked **except** elements with the data attribute `rr-unmask`.

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -30,6 +30,7 @@ import {
   isSerializedStylesheet,
   inDom,
   getShadowHost,
+  getElementFromNode,
 } from '../utils';
 
 type DoubleLinkedListNode = {
@@ -508,6 +509,8 @@ export default class MutationBuffer {
     switch (m.type) {
       case 'characterData': {
         const value = m.target.textContent;
+        const el = getElementFromNode(m.target)
+
         if (
           !isBlocked(m.target, this.blockClass, this.blockSelector, false) &&
           value !== m.oldValue
@@ -520,7 +523,7 @@ export default class MutationBuffer {
                 this.maskTextSelector,
               ) && value
                 ? this.maskTextFn
-                  ? this.maskTextFn(value)
+                  ? this.maskTextFn(value, el)
                   : value.replace(/[\S]/g, '*')
                 : value,
             node: m.target,
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -523,7 +523,7 @@ export default class MutationBuffer {
                 this.maskTextSelector,
               ) && value
                 ? this.maskTextFn
-                  ? this.maskTextFn(value, el)
+                  ? this.maskTextFn(value, getElementFromNode(m.target))
                   : value.replace(/[\S]/g, '*')
                 : value,
             node: m.target,
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -522,7 +522,7 @@ export default class MutationBuffer {
                 this.maskTextSelector,
               ) && value
                 ? this.maskTextFn
-                  ? this.maskTextFn(value, getClosestElementForNode(m.target))
+                  ? this.maskTextFn(value, closestElementOfNode(m.target))
                   : value.replace(/[\S]/g, '*')
                 : value,
             node: m.target,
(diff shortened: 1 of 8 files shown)


ArkadyFukzon:
+1 for this, really useful change

packages/rrweb/src/record/mutation.ts
```diff
@@ -520,7 +523,7 @@ export default class MutationBuffer {
                 this.maskTextSelector,
               ) && value
                 ? this.maskTextFn
-                  ? this.maskTextFn(value)
+                  ? this.maskTextFn(value, el)
```
Juice10, line 526: Lets do this inline so it doesn't get called if this.maskTextFn is false.
These get called often and can create performance issues, so I'd prefer we don't call it if it isn't absolutely necessary
```suggestion
                  ? this.maskTextFn(value, getElementFromNode(m.target))
```

packages/rrweb/src/record/mutation.ts
```diff
@@ -508,6 +509,8 @@ export default class MutationBuffer {
     switch (m.type) {
       case 'characterData': {
         const value = m.target.textContent;
+        const el = getElementFromNode(m.target);
```
Juice10, line 512:
```suggestion
```

See other comment

packages/rrweb/src/record/mutation.ts
benjackwhite: Ah that's a great catch.

benjackwhite:
@Juice10 made the changes 👍

Juice10 (approved):
Great addition @benjackwhite thanks!

packages/rrweb/src/record/mutation.ts
```diff
@@ -30,6 +30,7 @@ import {
   isSerializedStylesheet,
   inDom,
   getShadowHost,
+  getElementFromNode,
```
eoghanmurray, line 33:
```suggestion
  closestElementOfNode,
```

packages/rrweb/src/record/mutation.ts
```diff
@@ -520,7 +522,7 @@ export default class MutationBuffer {
                 this.maskTextSelector,
               ) && value
                 ? this.maskTextFn
-                  ? this.maskTextFn(value)
+                  ? this.maskTextFn(value, getElementFromNode(m.target))
```
eoghanmurray, line 525:
```suggestion
                  ? this.maskTextFn(value, getClosestElementForNode(m.target))
```

packages/rrweb/src/utils.ts
```diff
@@ -215,6 +215,23 @@ export function getWindowWidth(): number {
   );
 }
 
+/**
+ * Returns the given node as an HTMLElement if it is one, otherwise the parent node as an HTMLElement
+ * @param node - node to check
+ * @returns HTMLElement or null
+ */
+
+export function getElementFromNode(node: Node | null): HTMLElement | null {
```
eoghanmurray, line 224:
```suggestion
export function closestElementOfNode(m.target))(node: Node | null): HTMLElement | null {
```

eoghanmurray (approved):
naming suggestion

`getElementFromNode` sounded like a built in `getElementById` and hid the 'find parent' aspect.

the "closest" idea has precedence in https://developer.mozilla.org/en-US/docs/Web/API/Element/closest but maybe I'm overthinking it?

packages/rrweb/src/utils.ts
benjackwhite:
```suggestion
export function closestElementOfNode(node: Node | null): HTMLElement | null {
```

benjackwhite:
@eoghanmurray committed your changes (an then corrected the typos that I didn't initially notice 😅 ) 
Naming definitely makes sense

PR #1311, opened 2023-09-26
Add config option to turn off all snapshotting and related observers

eoghanmurray:
- allows RRWEB to be used for click/movement tracking alone, e.g. for a heatmaps use case
 - could also be used if there was a separate process for recording the DOM (in which case a 3rd party library like https://github.com/antonmedv/finder could be added to record targets instead of the mirror)

### DIFF
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -1321,9 +1321,7 @@ export function initObservers(
 
   return callbackWrapper(() => {
     mutationBuffers.forEach((b) => b.reset());
-    if (mutationObserver) {
-      mutationObserver.disconnect();
-    }
+    mutationObserver?.disconnect();
     mousemoveHandler();
     mouseInteractionHandler();
     scrollHandler();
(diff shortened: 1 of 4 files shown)


packages/rrweb/src/record/observer.ts
```diff
@@ -1314,7 +1321,9 @@ export function initObservers(
 
   return callbackWrapper(() => {
     mutationBuffers.forEach((b) => b.reset());
-    mutationObserver.disconnect();
+    if (mutationObserver) {
+      mutationObserver.disconnect();
+    }
```
Juice10, line 1326:
```suggestion
      mutationObserver?.disconnect();
```

Juice10 (approved):

packages/rrweb/src/record/observer.ts
eoghanmurray: Ah TIL the optional chaining operator isn't just a thing to squash a 'could be null' type error, but actually short circuits if the lhs is null/undeined.

PR #1313, opened 2023-09-27
[rrweb] Fix hookSetter maximum call stack exceeded

### LINKED ISSUE #731: too much recursion error
系统报了这个错误，
error name = InternalError,
error message = too much recursion error
stack：

用户浏览器 firefox 93和92， 目前发现是这两个版本。 可以确定chrome不会触发。 因为业务系统绝大多数都是谷歌，但是报错收集的只有firefox。
对应错误位置是 original.set.call(this, value);    的 set这里
set的调用无限循环，直接导致页面崩溃。

触发的地方是页面的input输入框，我们自己的ui组件， 复现几率 非100%，偶发。

```
function hookSetter(target, key, d, isRevoked, win) {
    if (win === void 0) { win = window; }
    var original = win.Object.getOwnPropertyDescriptor(target, key);
    win.Object.defineProperty(target, key, isRevoked
        ? d
        : {
            set: function (value) {
                var _this = this;
                setTimeout(function () {
                    d.set.call(_this, value);
                }, 0);
                if (original && original.set) {
                    original.set.call(this, value);
                }
            },
        });
    return function () { return hookSetter(target, key, original || {}, true); };
}
```

VladimirMilenko:
Fixes https://github.com/rrweb-io/rrweb/issues/731 by introducing the hooked property on the object.
This is needed to avoid recursive calls, which are currently happening at random inside the hook setter.
Also, instead of returning the hookSetter function in return, it will reset the original without creating more call stack entries.

### DIFF
--- a/packages/rrweb/src/utils.ts
+++ b/packages/rrweb/src/utils.ts
@@ -104,7 +104,14 @@ export function hookSetter<T>(
   isRevoked?: boolean,
   win = window,
 ): hookResetter {
+  const isAlreadyHooked = Object.prototype.hasOwnProperty.call(target, '__hooked__' + String(key));
+
+  // If already hooked, avoid re-hooking.
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  if (isAlreadyHooked && !isRevoked) return () => {};
+
   const original = win.Object.getOwnPropertyDescriptor(target, key);
+
   win.Object.defineProperty(
     target,
     key,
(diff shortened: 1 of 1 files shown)


packages/rrweb/src/utils.ts
```diff
@@ -104,15 +104,24 @@ export function hookSetter<T>(
   isRevoked?: boolean,
   win = window,
 ): hookResetter {
+  const isAlreadyHooked = Object.prototype.hasOwnProperty.call(
+    target,
+    '__hooked__' + String(key),
```
eoghanmurray, line 109: we should have some indication on the DOM property to indicate it originates with rrweb

eoghanmurray: We have `window.__rrMutationObserver` elsewhere, so maybe just `__rrHooked` would be good?

PR #1320, opened 2023-10-09
Fix a replay issue where the 500ms timeout was being hit

eoghanmurray:
- 3.7K nodes being inserted, with ~200 `<tr>`s as the root nodes making up the `resolveTrees` array
 - each table row was dependent on the next, resulting in cascading failures; 19K instances of [repopulation of queue](https://github.com/rrweb-io/rrweb/blob/30be92705b76528582eec99558bad5b29a1a82c3/packages/rrweb/src/replay/index.ts#L1503) due to `nextNotInDOM`, and an additional cascading 350K [repopulation of queue](https://github.com/rrweb-io/rrweb/blob/30be92705b76528582eec99558bad5b29a1a82c3/packages/rrweb/src/replay/index.ts#L1483) due to missing parent nodes as algorithm attempted to continue over children


packages/rrweb/src/utils.ts
```diff
@@ -368,7 +369,16 @@ export function queueToResolveTrees(queue: addedNodeMutation[]): ResolveTree[] {
...
       continue;
     }
-    queueNodeTrees.push(putIntoMap(mutation, null));
+    if (nextId) {
+      rootNexts.add(nextId);
+    }
+    const rootTree = putIntoMap(mutation, null);
+    if (rootNexts.has(mutation.node.id)) {
+      // some existing tree can't be inserted until this one is (nextNotInDOM check)
+      queueNodeTrees.unshift(rootTree);
```
JonasBa, line 378: Word of caution here - shift is O(n) which makes this operation O(n^2) now. Since this seems to be a hot path, you might be better off using a different data structure here

eoghanmurray: I think the assumption here was that queueNodeTrees will be small in comparison to the tree data structure within each one.  At any rate, the `queueToResolveTrees` function is there to put some organisation on the nodes in order to forestall the pathological cases elsewhere (the `appendNode` and the fact that `appendNode` can fail and put things back on the `queue` which causes the whole `while (queue.length) {` bit to brute force again and again)
So I think this is a minor concern; it could be replaced by a doubly linked list but my assumption would be that it's not a hot path.

JonasBa: Seems reasonable, I just wanted to highlight it :)

eoghanmurray:
This has been in production at Statcounter since Oct 2023

PR #1322, opened 2023-10-10
Browser/Spec bug: Failing test for capturing css shorthand with confounding variable

eoghanmurray:
This PR contains a new test which shows the failing behaviour;
I can't currently see a way around this, so this is a bug report and maybe a place to link to any bugs in browser version trackers.

#1268 PR (as reported in #1246) is very similar however this version occurs while serializing the static stylesheet, rather than as part of a mutation (where we can work around it)

Problem:

```
#one {
	animation: var(--animation-fade-in);
	animation-delay: 0.2s;
}
```
in a stylesheet gets serialized as

```
#one {
  animation-duration: ; 
  animation-timing-function: ; 
  animation-iteration-count: ; 
  animation-direction: ; 
  animation-fill-mode: ; 
  animation-play-state: ; 
  animation-name: ; 
  animation-timeline: ; 
  animation-range-start: ; 
  animation-range-end: ; 
  animation-delay: 0.2s; 
}
```

I presume a similar thing might happen for e.g. `background: var(--mycolor); background-image: url('x.png');`

### DIFF
--- a/packages/rrweb/test/__snapshots__/integration.test.ts.snap
+++ b/packages/rrweb/test/__snapshots__/integration.test.ts.snap
@@ -3348,6 +3348,204 @@ exports[`record integration tests can record node mutations 1`] = `
 ]"
 `;
 
+exports[`record integration tests can record shorthand properties defined via variables 1`] = `
+"[
+  {
+    \\"type\\": 0,
+    \\"data\\": {}
+  },
+  {
+    \\"type\\": 1,
+    \\"data\\": {}
+  },
+  {
+    \\"type\\": 4,
+    \\"data\\": {
+      \\"href\\": \\"about:blank\\",
+      \\"width\\": 1920,
+      \\"height\\": 1080
+    }
+  },
+  {
+    \\"type\\": 2,
+    \\"data\\": {
+      \\"node\\": {
+        \\"type\\": 0,
+        \\"childNodes\\": [
+          {
+            \\"type\\": 1,
+            \\"name\\": \\"html\\",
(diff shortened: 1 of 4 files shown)


eoghanmurray:
Similarly the following would exhibit the problem:
```
.gradient {
  background: rgb(var(--color-background));
  background: var(--gradient-background);
  background-attachment: fixed;
}
```

eoghanmurray:
There is a possibility to fix this when we have access to the raw `.innerText` of the style; 
 - we could construct an artificial Stylesheet using https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet
 - then compare the rules with the derived rules from `.sheet`

If they are the same, that means we can trust the `.innerText` as being up-to-date (not mutated by .insertRule etc)
In that case we just use the `.innerText` as the serialization.

For external stylesheets (<link> elements) we can [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch) the raw contents and do the comparison that way (fetch should be from cache).

This will all be possible in an async manner after #1475 

Hat-tip: CSSStylesheet idea was from https://github.com/rrweb-io/rrweb/pull/1357/files#diff-c61c2c2a3a84e2ee30a603d8363a994acce4caef8827931eb5b7317f08b53cb1R588

Juice10:
This also happens to `padding` and `margin` shorthands.
More info: https://issues.chromium.org/issues/40771479#comment13
And for both comedy and even more information, you get to see me getting yelled at here: https://issues.chromium.org/issues/40771479#comment14

PR #1324, opened 2023-10-13
Pref: export eventWithTime

Belen-Luo:
When using emit, you do not know the type of event and can only write it as any[], which is not very elegant. rrweb should provide the type directly

### DIFF
--- a/.changeset/lemon-lamps-switch.md
+++ b/.changeset/lemon-lamps-switch.md
@@ -1,0 +1,5 @@
+---
+'rrweb': patch
+---
+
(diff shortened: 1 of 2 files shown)


Belen-Luo:
> 这是供另一个打字稿程序使用的吗？ 这是我们可以添加到https://github.com/rrweb-io/rrweb/blob/master/docs/recipes/的东西还是微不足道的？

Just provide an accessible type that can be recorded without going to https://github.com/rrweb-io/rrweb/blob/master/docs/recipes/

eoghanmurray (approved):

Issue #1301, opened 2023-09-09
[Bug]: Textarea shows default value twice in replay


josephmathew900:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

latest

### Expected Behavior

Textarea shows default value twice in replay

### Actual Behavior

Textarea default value should not be duplicated.

### Steps to Reproduce

https://vimeo.com/862640906/5ee922b408

### Additional Information

I found a similar issue https://github.com/rrweb-io/rrweb/issues/745, and in that issue, it is mentioned that the bug has been fixed, but the issue still exists.

YanJie-001:
I also encountered this bug and I didn't find that this bug was fixed.

YanJie-001:
This problem still exists. You can use rrweb-player and remember not to use the default version.To use 2.0.0-alpha.11.

eoghanmurray:
Thank you for putting together a great bug report.
In your video at 01:15 there is a screenshot of an event.
Is the textare node there part of a FullSnapshot or of a Mutation?

It seems to me that we should be addressing the problem on the _record_ side, in that we shouldn't be producing textarea nodes which contain both a value attribute and a child with the same value in it.  Would it be possible to look into that, and then see whether your fix in #1326 is still needed?

eoghanmurray:
I'm just checking in case the record side can produce all variants of textareas:
 - attribute value and child text node set (as you witnessed)
 - only attribute value set
 - only child text node set
If that is the case, then #1326 might break one of the latter two above

josephmathew900:
@eoghanmurray 
```
In your video at 01:15 there is a screenshot of an event.
Is the textare node there part of a FullSnapshot or of a Mutation?
```

It is part of a FullSnapshot.

eoghanmurray:
#1351 is ready for review; you can execute the code to check that it fixes your use case and let me know if not

josephmathew900:
I tested my use case with the code changes, and it is working perfectly fine.

PR #1326, opened 2023-10-13
bugfix: Removve redundant content so that the textarea content in the replay will no longer appear repeatedly.

YanJie-001:
Removve redundant content so that the textarea content in the replay will no longer appear repeatedly.
To fix [Bug]: Textarea shows default value twice in replay #1301

### DIFF
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -182,22 +182,11 @@ function buildNode(
           continue;
         }
 
-        const isTextarea = tagName === 'textarea' && name === 'value';
         const isRemoteOrDynamicCss = tagName === 'style' && name === '_cssText';
         if (isRemoteOrDynamicCss && hackCss && typeof value === 'string') {
           value = addHoverClass(value, cache);
         }
-        if ((isTextarea || isRemoteOrDynamicCss) && typeof value === 'string') {
-          const child = doc.createTextNode(value);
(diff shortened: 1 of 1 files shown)


eoghanmurray:
See my comments in #1301

eoghanmurray:
Ignoring the current failing tests which are related to `_cssText` which is presumably unrelated, this PR really needs a few exemplifying test cases to ensure there isn't a future regression on this (it looks like there was a few prior rounds of changes around this issue in the git history).  Would you be able to supply that or would you prefer to wait until I can figure it out?

I imagine we'd want to add a <textarea> one of the test cases in packages/rrweb/test/record.test.ts in order to show that the repeated value is a problem.

eoghanmurray:
As currently implemented, this PR modifies the tests as follows:

```
 rrweb-snapshot:test:     -         <textarea name="" id="" cols="30" rows="10">1234</textarea>
rrweb-snapshot:test:     +         <textarea name="" id="" cols="30" rows="10" value="1234"></textarea>
```

This is less correct in that if the latter is rendered in HTML, then the value attribute is ignored (the `.value` parameter is available in javascript)

eoghanmurray:
I'm gonna close this one as #1351 goes into much more detail than this one by fixing at recording time, but also fixing for existing recordings where the bad data is already in the event (like this fix does)

eoghanmurray:
Forgot to say: thank you for this PR, I wouldn't have made my one if I could not see your intended changes!

PR #1328, opened 2023-10-17
Refactor to preclude the need for a continuous raf loop

eoghanmurray:
It seems to me that the raf loop is unnecessary, so I've removed it to get your feedback.

Instead we queue up a single raf callback which will clear anything added in this frame.

I've no understanding as to why this exclusivity is per animation frame, and not (say) per emitted mutation event, any insight into that?


eoghanmurray:
Oh yeah; these changes haven't been tested!

eoghanmurray:
Although I have it in production, so it's not obviously breaking things, there could be some subtleties with the shadowbuffer that might need expert analysis or a test...

I wrote this based on the logic of how the weak map was cleared etc.

JonasBa:
@eoghanmurray very welcome change! This was creating a lot of noise when profiling the code 👏🏼

eoghanmurray:
@JonasBa this is a hard one for anyone to review and we don't have tests on it.
Are you running this in production by any chance?

Juice10 (approved):

PR #1331, opened 2023-10-20
fix(snapshot): dimensions for blocked element not being applied

billyvg:
Dimensions are not being properly applied when you seek to a position in the replay. Need to use `setProperty`, otherwise the style attributes get lost. This was happening when node was an RRElement.

### DIFF
--- a/.changeset/perfect-dolls-grab.md
+++ b/.changeset/perfect-dolls-grab.md
@@ -1,0 +1,5 @@
+---
+"rrweb-snapshot": patch
+---
+
+fix dimensions for blocked element not being applied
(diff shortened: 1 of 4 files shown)


eoghanmurray:
Thank you for the contribution and for contributing tests :)

To further understand: this was a problem with RRDom but the original code was fine if applied in a real DOM context?

billyvg:
@eoghanmurray correct, the test I added in `rebuild.test` works prior to this change.

YunFeng0817 (approved):
Thanks for the fix
Can you please add a change set log for this PR?

eoghanmurray:
@billyvg this is good to merge, but could you either rebase/merge from master, or else tick the box to 'Allow edits from maintainers' so we can do that on your behalf?

Issue #1337, opened 2023-10-24
[Feature Request]: Optimize full snapshot serialization performance 


ababik:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a feature request that matches the one I want to file, without success.

### What package is this feature request for?

rrweb

### Problem Description

The full snapshot serialization process takes a significant amount of time, blocking the UI thread. I spent some time analyzing the process and noticed that there is a relatively expensive `closest` function call for each text node, which can be avoided.

### Proposed Solution

Within the snapshot serialization, I refined the method by which rrweb determines if text should be masked. Presently, for every individual text node during recursion, the [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) method is invoked to ascertain if the node is a descendant of an element marked for masking. I modified this by passing the masking flag directly as a parameter to the recursion function, thereby eliminating superfluous lookups.

### Alternatives Considered

N/A

### Additional Information

I ran a series of performance measurements for the full snapshot serialization of a document generated by the following code:
```
for (let i1 = 0; i1 < 100; i1++) {
    let parent = document.createElement("div");
    document.getElementById("container").append(parent);
    for (let i2 = 0; i2 < 100; i2++) {
        let child = document.createElement("div");
        child.innerText = `DIV ${i1} - ${i2}`;
        parent.append(child);
        parent = child;
    }
}
```
The original function takes avg `34.55 ms` while optimized is avg `27.35 ms`.

eoghanmurray:
Would it be possible to re-run your benchmark measurements against #1349 ? 

I think my approach adds a little less complexity than your one, so if it performs just as good then I'd prefer to go with it, but you'll have to let me know if performance is okay!

ababik:
Thanks for getting back to me. I'll work on this during next couple of days and let you know. Thanks.

ababik:
I like the idea of having all masking logic in the same place; the code is more maintainable.
Although it's less performant compared to my approach, where I determine the list of elements to mask only once.
I ran the same test, and it takes ~31ms.
The original implementation (`closest`): 34.55 ms
Your implementation: ~31 ms
My implementation: 27.35 ms

It's definitely worth merging your optimization.

eoghanmurray:
[...]

PR #1338, opened 2023-10-24
[masking] Optimize full snapshot serialization performance

ababik:
Implementation for https://github.com/rrweb-io/rrweb/issues/1337

Within the snapshot serialization, I refined the method by which rrweb determines if text should be masked. Presently, for every individual text node during recursion, the [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) method is invoked to ascertain if the node is a descendant of an element marked for masking. I modified this by passing the masking flag directly as a parameter to the recursion function, thereby eliminating superfluous lookups.

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -440,6 +440,7 @@ function serializeNode(
      * `newlyAddedElement: true` skips scrollTop and scrollLeft check
      */
     newlyAddedElement?: boolean;
+    maskCurrentNode?: boolean;
   },
 ): serializedNode | false {
   const {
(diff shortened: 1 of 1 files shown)


eoghanmurray:
I'm concerned that this introduces a bit too much (incremental) complexity by adding a separate method of doing the check.

Would you consider the following strategy?

 - pass a new boolean `isMasked` down the recursive calls, starting off `false`
 - if this is still false, in `serializeElementNode` do a single masking check without using `closest`, so like the body of the function `needMaskingText`, but without having to move to `node.parentElement`, and only doing `classList.contains`/`classMatchesRegex` and `el.matches(maskTextSelector)`
 - pass the value from that down recursively, so descendents don't have to redo the check if it is true

That might be a bit slower than your method, but would be faster for the regex case and would mean we'd have the masking logic all in one place.

eoghanmurray:
I've embodied what I've described above in #1349 

Would you be able to compare it from a performance perspective in the same way you did your one?

My take turned out to be not as elegant as I hoped as it exposed that `needMaskingText` is also used during mutations, and I'm not sure whether it will correctly check the ancestors in the regexp case.

PR #1339, opened 2023-10-25
ref: Avoid cloning events to add `timestamp`

mydea:
We are always passing in fresh objects to this method, so instead of cloning this into a new object, we can just put the `timestamp` on the given object directly and return it, saving a bit of processing cost.

### DIFF
--- a/.changeset/fast-colts-itch.md
+++ b/.changeset/fast-colts-itch.md
@@ -1,0 +1,5 @@
+---
+'rrweb': patch
+---
+
+ref: Avoid cloning events to add `timestamp`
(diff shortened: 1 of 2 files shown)


eoghanmurray:
This is modifying the underlying object?  Are you sure there aren't any typing problems with this, or possible knock on problems? (I can't think of anything over and above it being 'less clean')

mydea:
Yes, but as far as I can tell we are always passing in a fresh object into this anyhow. We've been running this on our fork at Sentry without problems (as far as I can tell...) for some time now!

eoghanmurray:
I've gone a bit further with #1441 to avoid that extra function call at all

eoghanmurray:
And thank you for the contribution!

(I couldn't make commits directly to this PR)

eoghanmurray:
(actually #1441 needs some more work)

eoghanmurray:
#1441 is ready now as a larger change than this one

eoghanmurray:
#1441 has been merged and I've tagged you in the merge commit, thank you for the PR!

PR #1340, opened 2023-10-25
ref: Avoid unnecessary cloning of objects or arrays

mydea:
Avoid some places where we necessarily cloned objects/arrays.


Juice10 (approved):
Thanks for this @mydea!

PR #1341, opened 2023-10-25
WIP: Ensure text of checkboxes is always masked

mydea:
This is on top of https://github.com/rrweb-io/rrweb/pull/1340 - with this PR we ensure to always mask the text value of checkboxes that are disabled.


mydea:
Ah, I see, I've already opened a PR for this: https://github.com/rrweb-io/rrweb/pull/1164/files so closing this

PR #1346, opened 2023-11-03
Add full support for custom elements

YunFeng0817:


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -1069,7 +1069,15 @@ export function serializeNodeWithId(
...
-    for (const childN of Array.from(n.childNodes)) {
+    let childNodes = Array.from(n.childNodes) as Node[];
+    childNodes = childNodes.filter(
+      // filter out nodes that are assigned to a slot other places.
+      (childN) => !(childN as HTMLElement).assignedSlot,
+    );
+    if (n instanceof HTMLSlotElement)
+      // add nodes that are assigned to this slot and treat them as normal children.
+      childNodes = childNodes.concat(Array.from(n.assignedNodes()));
+    for (const childN of Array.from(childNodes)) {
```
Juice10, line 1080: `instanceof HTMLSlotElement` doesn't work cross iframe, this should:
```suggestion
    if ('assignedNodes' in n)
      // add nodes that are assigned to this slot and treat them as normal children.
      childNodes = childNodes.concat(Array.from(n.assignedNodes()));
    for (const childN of Array.from(childNodes)) {
```

PR #1351, opened 2023-11-09
Fix serialization and mutation of <textarea> elements

### LINKED ISSUE #1301: [Bug]: Textarea shows default value twice in replay

eoghanmurray:
Taking account the duality that the value can be set in either the child node, or in the value _parameter_ (not attribute)

This is an alternative to #1326 to fix #1301


eoghanmurray:
I've added a new aspect to rrweb/test/integration.test.ts in that it now tests the replayer as well as checking the snapshot; the replayer is needed to read out what the actual values of the textarea are after specific mutations are applied.

josephmathew900:
https://github.com/rrweb-io/rrweb/issues/1301#issuecomment-1817422957

@eoghanmurray Do we have any blockers to getting this PR merged? After merging this PR, in which version will the fix be available?

Issue #1344, opened 2023-10-30
[Bug]: CSSRules inserted w/ CSSStyleSheet.insertRule() are lost on skipping w/ virtualDom


jaj1014:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

2.0.0-alpha.11

### Expected Behavior

When using `virtualDom`, CSSRules inserted (`CSSStyleSheet.insertRule()`) on a style element should persist when skipping.

### Actual Behavior

When using `virtualDom`, CSSRules (`CSSStyleSheet.insertRule()`) inserted on a style element are lost when skipping.

### Steps to Reproduce

What I found to be the crux of the issue is actually in rrdom when `diff()`'ing current dom against the virtual dom. If a style element which had `IncrementalSource.StyleSheetRule` events applied gets moved due to the diff, it loses those inserted style rules. This appears to be a side effect of using `Node.insertBefore()` on style elements.

jsfiddle demonstrating what insertBefore does: https://jsfiddle.net/8ps3zmf7/

While this is where I noticed things breaking down, there is more contributing to the issue. I've included more detail on my overall learnings in 'Additional Information' below.

### Additional Information

I did a fair amount of debugging to try to figure out what was happening to cause the issue and these are the things I found:

1. IncrementalSnapshot type events with a source of StyleSheetRule that add rules use `CSSStyleSheet.insertRule(rule, index)` to add them to the style element.
2. `diff` in rrdom uses `Node.insertBefore()` to 'move' elements to their updated location based on the diff algo. When used on a style element (or parent elements which contain style elements as children), only rules derived from innerText/innerHTML persist. Any inserted styles (`CSSStyleSheet.insertRule()`) do not persist with the style element to it's new location. 
	1. This can be repro'd in the console pretty easily.  adding a style element in the dom, inserting styles on it using `CSSStyleSheet.insertRule(rule, index)`, then moving the element using `Node.insertBefore()`. When you inspect the element, all of the styles that were inserted will be missing. Fiddle example: https://jsfiddle.net/8ps3zmf7/
[...]

PR #1357, opened 2023-11-20
fix: inserted styles lost when moving elements

jaj1014:
hello 👋 - I created an [issue](https://github.com/rrweb-io/rrweb/issues/1344) a couple weeks back and had some time to dig in a bit more to see if I could figure out a potential solution.

The issue linked above has more detail, but the tl;dr; is that styles which are inserted using `CSSStyleSheet.insertRule()` are lost during diff operations that result in that style element getting moved. 

In order to better demonstrate the problem, I was finally able to create a recording (linked below): If you skip from 0:04 seconds to 0:15, you'll see the teal div at the top loses it's teal color. If you let it play through from 0:04 seconds to 0:15, you'll notice the teal div is still teal. This is because skipping triggers the style element which had the teal styles inserted gets moved as a result of the diff, so the inserted styles get lost. Letting it play through, they aren't lost because diff isn't run and things get mutated normally (w/o virtual dom).

https://rrwebdebug.com/play/index.html?url=https%3A%2F%2Fgist.github.com%2Fjaj1014%2F6e6658d0f24ea297d31edbcfdd44379e&version=2.0.0-alpha.8&virtual-dom=on&play=on

Because the issue has several things contributing to the issue, I wasn't sure where it made the most sense to apply a fix. I took a stab at applying a fix during the diff process inside `buildFromNode` to copy CSSRules from the dom `node` into the `rrNode` for style elements. Later in the diff flow, [there is code](https://github.com/pendo-io/rrweb/blob/master/packages/rrdom/src/diff.ts#L271) that will take those style rules copied from the original and apply them.

### DIFF
--- a/packages/rrweb/test/replayer.test.ts
+++ b/packages/rrweb/test/replayer.test.ts
@@ -11,6 +11,7 @@ import {
   waitForRAF,
 } from './utils';
 import styleSheetRuleEvents from './events/style-sheet-rule-events';
+import movingStyleSheetOnDiff from './events/moving-style-sheet-on-diff';
 import orderingEvents from './events/ordering';
 import scrollEvents from './events/scroll';
 import scrollWithParentStylesEvents from './events/scroll-with-parent-styles';
@@ -177,6 +178,22 @@ describe('replayer', function () {
     await assertDomSnapshot(page);
   });
 
+  it('should persist StyleSheetRule changes when skipping triggers parent style element to move in diff', async () => {
+    await page.evaluate(`events = ${JSON.stringify(movingStyleSheetOnDiff)}`);
+
+    const result = await page.evaluate(`
+      const { Replayer } = rrweb;
+      const replayer = new Replayer(events);
+      replayer.pause(3000);
+      const rules = [...replayer.iframe.contentDocument.styleSheets].map(
+        (sheet) => [...sheet.rules],
+      ).flat();
+      rules.some((x) => x.cssText === '.target-element { background-color: teal; }');
+    `);
+
+    expect(result).toEqual(true);
+  });
+
   it('should apply fast forwarded StyleSheetRules that where added', async () => {
     await page.evaluate(`events = ${JSON.stringify(styleSheetRuleEvents)}`);
     const result = await page.evaluate(`
@@ -228,7 +245,7 @@ describe('replayer', function () {
     await waitForRAF(page);
 
     /** check the second selection event */
-    [startOffset, endOffset] = (await page.evaluate(`      
+    [startOffset, endOffset] = (await page.evaluate(`
       replayer.pause(410);
       var range = replayer.iframe.contentDocument.getSelection().getRangeAt(0);
       [range.startOffset, range.endOffset];
@@ -705,7 +722,7 @@ describe('replayer', function () {
       events = ${JSON.stringify(canvasInIframe)};
       const { Replayer } = rrweb;
       var replayer = new Replayer(events,{showDebug:true});
-      replayer.pause(550);            
+      replayer.pause(550);
     `);
     const replayerIframe = await page.$('iframe');
     const contentDocument = await replayerIframe!.contentFrame()!;
@@ -767,7 +784,7 @@ describe('replayer', function () {
       const replayer = new Replayer(events);
       replayer.play();
     `);
-    await page.waitForTimeout(50);
+    await waitForRAF(page);
 
     await assertDomSnapshot(page);
   });
@@ -791,7 +808,7 @@ describe('replayer', function () {
     await page.evaluate(`
       const { Replayer } = rrweb;
       let replayer = new Replayer(events);
-      replayer.play();      
+      replayer.play();
     `);
 
     const replayerWrapperClassName = 'replayer-wrapper';
--- a/packages/rrdom/src/diff.ts
+++ b/packages/rrdom/src/diff.ts
@@ -413,7 +413,7 @@ function diffChildren(
       nodeMatching(oldStartNode, newEndNode, replayer.mirror, rrnodeMirror)
     ) {
       try {
-        oldTree.insertBefore(oldStartNode, oldEndNode.nextSibling);
+        handleInsertBefore(oldTree, oldStartNode, oldEndNode.nextSibling);
       } catch (e) {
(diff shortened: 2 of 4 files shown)


jaj1014:
The original approach I took w/ copying cssRules into the virtual nodes created an issue w/ duplicate styles being applied after the diff. This was because that approach wasn't surgical enough and applied to _all_ style sheets - not just the ones that were being moved. To address that, I moved the fix into the diff work to only copy inserted styles over when insertBefore() is being used.

I also added a test case in replayer.test that fails if you run w/o my changes (demonstrating the issue) and passes w/ my changes.

I'm not sure if the written code satisfies rrweb style/standards, so I haven't written tests for the 'util' functions that were added to diff.ts. Happy to adjust where that code lives based on comments. :)

packages/rrweb/test/replayer.test.ts
```diff
@@ -718,7 +735,7 @@ describe('replayer', function () {
       const replayer = new Replayer(events);
       replayer.play();
     `);
-    await page.waitForTimeout(50);
+    await page.waitForTimeout(150);
```
Juice10, line 738: This is a more reliable way of forcing a wait
```suggestion
    await waitForRAF(page);
```

jaj1014: updated and thank you!

jaj1014:
Finally go around to adding unit test coverage for the added functions.

packages/rrweb/src/record/iframe-manager.ts
```diff
@@ -98,6 +111,15 @@ export class IframeManager {
     const iframeSourceWindow = message.source;
     if (!iframeSourceWindow) return;
 
+    if (
+      iframeSourceWindow == window.parent &&
+      window != window.parent &&
+      message.data.snapshot
```
eoghanmurray, line 117: Annoyingly, typescript (eslint) is failing here; you could define your own message type (see CrossOriginIframeMessageEventContent) and cast to that message type with the `as` keyword, or maybe the following will work

```suggestion
            'snapshot' in message.data
```

packages/rrweb/src/record/observer.ts
```diff
@@ -93,7 +91,7 @@ export function initMutationObserver(
   // see mutation.ts for details
   mutationBuffer.init(options);
   let mutationObserverCtor =
```
eoghanmurray, line 93: No idea what all the angularZoneSymbol stuff is, but now that this is no longer reassigned by that code, eslint requires us to change to a `const`

```suggestion
  const mutationObserverCtor =
```

eoghanmurray (changes requested):

jaj1014:
sorry @eoghanmurray my branch got screwed up w/ some other commits. I'll try to get it clean up and back to only the changes pertinent to this PR. or worst case, close and open a clean one. sorry about this.

eoghanmurray:
> diff with actual and get only inserted styles

The new `getInsertedStylesFromElement` looks pretty cool; basically you are comparing raw text node rules to derived css rules and figuring out the diffs?  This approach could potentially be useful during snapshotting as we've opted to go directly to `.sheet.rules` in case there were previous inserted rules.

For this PR though, I feel like it might be overkill and that it adds too much complexity; I feel like the underlying problem should be fixed instead (that elements don't get moved around after diffChildren).  But maybe I'm missing something and this PR is the only way of doing it?

jaj1014:
[cae6981](https://github.com/rrweb-io/rrweb/pull/1357/commits/cae6981afab28400183db196eb9cd72393aff5c7) has the changes pertinent to the originally described problem. all the other commits got mixed in unintentionally. unsure if the complexity you're referring to is contained in this one commit, or if you're referring to the others.

you're correct on the gist of the code: it rectifies the issue caused by `insertBefore` dropping styles that were added with `CSSStyleSheet.insertRule()`. It does this by creating a copy of the stylesheet without those inserted rules and diffing the copy against the actual (before being moved) to get a map of the inserted styles it can apply after the element is moved.

It's been a while since I've been in this code, and what I thought I know may not even be right. I was thinking that moving elements was just applying the result of diffing with the virtualDom. When not using the virtualDom optimization, this issue doesn't happen. I'm don't know if it is possible to not move elements after diffChildren

eoghanmurray:
Ah okay; if you wish you could cherry-pick that commit onto an up to date trunk and push here.

Until then I'll leave it to @YunFeng0817 to look at the code and see if this is the best approach or if there is a simpler way of achieving the same thing with rrdom.

billyvg:
I've been trying this out and I think this causes a perf regression. I'm seeing the differ go from ~300-400 to ~600-700ms. The profiler indicates that `getInsertedStylesFromElement` is forcing reflow quite often.

jaj1014:
> I've been trying this out and I think this causes a perf regression. I'm seeing the differ go from ~300-400 to ~600-700ms. The profiler indicates that `getInsertedStylesFromElement` is forcing reflow quite often.

Interesting. It makes sense that it's going to cause a performance hit to diff since it's running code that wasn't there before, but it should only run for style elements since those are the ones it's concerned with - and I'm not sure what in that function would cause a reflow...

Was there something specific you saw w/ `getInsertedStylesFromElement` that was causing the reflow? 

The forEach that [goes through the inserted styles](https://github.com/rrweb-io/rrweb/pull/1357/files#diff-c61c2c2a3a84e2ee30a603d8363a994acce4caef8827931eb5b7317f08b53cb1R671) found by `getInsertedStylesFromElement` that inserts onto the sheet... I could see each of those possibly triggering a reflow 🤔. But I'm also not super knowledgeable about profiler and didn't do testing for performance when I originally created this PR over a year ago.

jaj1014:
couldn't resist and was curious...

maybe because it uses `innerText` (sources: [what forces layout/reflow from Paul Irish](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#also) and [an article that outlines differences](https://dev.to/4myc/javascript-innerhtml-innertext-and-textcontent-ih#:~:text=One%20drawback%20to%20using%20innerText,for%20re%2Drendering%20the%20document))

Based on the article, it looks like `.textContent` instead would avoid the reflow. So line 627 would be: `tempStyleSheet.replaceSync(styleElement.textConent);`

PR #1359, opened 2023-11-27
fix: inlineImages and image is cors,the image tag is broken

xujiujiu:
#1243


Juice10:
Thanks for creating this fix @xujiujiu, we're going to be replacing all of the existing code for `inlineImages` with https://github.com/rrweb-io/rrweb/pull/1239 we might use your code as a fallback in the case where `fetch` isn't allowed due to security settings.

PR #1360, opened 2023-11-28
🎈 perf(web-extension): conditional check

huangkairan:
update condition check in Player component


huangkairan:
`sessionId`  has been checked at the top of `useEffect`  https://github.com/rrweb-io/rrweb/blob/master/packages/web-extension/src/pages/Player.tsx#L20

YunFeng0817 (approved):
Thanks!

PR #1373, opened 2023-12-08
allow passing null to maskTextClass/blockClass

mdellanoce:
Allows you to disable maskTextClass and blockClass by passing `null`, which can give a little performance boost for large snapshots/mutations if you only do your masking/blocking with selectors. Regex class matching in particular seems to be a pretty big drag on performance.

### DIFF
--- a/.changeset/quick-cows-chew.md
+++ b/.changeset/quick-cows-chew.md
@@ -1,0 +1,5 @@
+---
+'rrweb-snapshot': patch
+---
+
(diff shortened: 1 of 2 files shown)


eoghanmurray:
This seems great, sorry for the delay in getting it reviewed.
Could you merge or rebase on current master and push here in order to trigger the tests again?

colingm:
@eoghanmurray this is rebased, sorry for the delay

PR #1366, opened 2023-12-04
[chore] [tests] fix tests on master

Juice10:
currently tests on master are failing, upgrade jest and rollup to fix this

PR #1374, opened 2023-12-12
Capture stylesheets designated as `rel="preload"`

andrewpomeroy:
Encountered an issue where stylesheets failed to capture when they were referenced by link tag with `rel="preload"` instead of `rel="stylesheet"`. This change will broaden the rules so that these preloaded stylesheets can also be recognized and captured at the snapshot phase.

For background: This attribute assignment is commonly used as a performance hack for the initial loading phase of a page—meant to allow stylesheets to be preloaded—and is usually combined with `onload="this.rel = 'stylesheet'"` so that it is still, upon load, recognized as a stylesheet at runtime. The snapshot phase, naturally, will miss these unless we tweak the link-tag pattern matching to look for it specifically. More info: https://stackoverflow.com/a/61889123

In addition, I added reinforced pattern matching for both `.css` file paths—for the aforementioned scenario and the existing `.js` / `rel="prefetch"` scenario—so that paths will still be matched if they include search parameters or fragments.


packages/rrweb-snapshot/src/utils.ts
```diff
@@ -331,3 +331,18 @@ export function getInputType(element: HTMLElement): Lowercase<string> | null {
...
+/**
+ * Extracts the file extension from an a path, considering search parameters and fragments.
+ * @param path - Path to file
+ * @param baseURL - [optional] Base URL of the page, used to resolve relative paths. Defaults to current page URL.
+ */
+export function extractFileExtension(
+  path: string,
+  baseURL?: string,
+): string | null {
+  const url = new URL(path, baseURL ?? window.location.href);
```
Juice10, line 344: We should wrap this in a try catch in case someone adds a gibberish pathname that doesn't get recognised by URL and will otherwise throw

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -331,3 +331,18 @@ export function getInputType(element: HTMLElement): Lowercase<string> | null {
...
+  const regex = /\.([0-9a-z]+)(?:[?#]|$)/i;
+  const match = url.pathname.match(regex);
```
Juice10, line 346: I'm not sure you need to match against `?` or `#`, according to the MDN documentation .pathname already cuts this off

Juice10 (changes requested):
Thanks for creating this PR @andrewpomeroy! This is a nice addition to rrweb. I found two minor things, check them out and let me know what you think and then we can merge this.

packages/rrweb-snapshot/src/utils.ts
andrewpomeroy: Dropped it 👍

packages/rrweb-snapshot/src/utils.ts
andrewpomeroy: Done! 👌

andrewpomeroy:
Thanks @Juice10! Your suggestions make perfect sense. I went ahead and made those changes.

andrewpomeroy:
Hi @Juice10 ! It looks like this is now failing tests, which look to be the same tests currently failing on master. I tried checking out https://github.com/rrweb-io/rrweb/pull/1366 and fixing them myself, but didn't get much traction.

My team is looking to close out this issue on our end; no pressure at all intended, but what are the chances of this getting merged/released in the next few weeks? We just need to be able to make the call whether to fork or not. Thanks again for your contributions 🙂

andrewpomeroy:
@Juice10 Whoops, just saw your comment here https://rrweb.slack.com/archives/C01BYDC5C93/p1702300072935979?thread_ts=1701887171.474259&cid=C01BYDC5C93 ! I'm assuming it won't be much longer then 😄 Looking forward to it!

Juice10:
@andrewpomeroy hang tight Andrew, unfortunately I wasn’t able to fix master in https://github.com/rrweb-io/rrweb/pull/1366 (I should close that one), but indeed as mentioned in slack I’m hoping to merge https://github.com/rrweb-io/rrweb/pull/1239 soon which does fix master, and then we can merge this PR and do a release.

andrewpomeroy:
Great to hear! Thanks for the update.

Juice10:
@andrewpomeroy hey Andrew, could you pull from master, ci is passing now

Juice10 (approved):

eoghanmurray:
> and is usually combined with onload="this.rel = 'stylesheet'" so that it is still, upon load, recognized as a stylesheet at runtime

Shouldn't this get picked up as a mutation?  And I think the mutation should then serialize the stylesheet correctly?
I think we need a test for this scenario to either demonstrate the `preload` bit or remove it if it isn't necessary.

andrewpomeroy:
> Shouldn't this get picked up as a mutation?
It's been a while since I've looked at this, but to the best of my memory— these mutations don't trigger a full snapshot, which is when the stylesheet-inlining process would take place.

eoghanmurray:
Yes, thanks — I actually have since written some tests and done some more work on this; the problem was that the mutation weren't doing anything when an attribute change caused a <link> to 'become a stylesheet'
More details here: 
https://github.com/rrweb-io/rrweb/pull/1483
which is still draft and WIP

andrewpomeroy:
Gotcha. Good thinking! I'm all for it.

PR #1375, opened 2023-12-14
Make properties x and y optional in mouseInteractionParam

avillegasn:
The event representing a mouse interaction of focus an element does not include the properties `x` and `y`. Thus, they must be defined as optional.

Example of "focus" event without the `x` and `y` properties:

```
{
   "type": 3,
   "data": {
       "source": 2,
       "type": 5,
       "id": 408
    },
   "timestamp": 1702560401489
}
```

### DIFF
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -1161,8 +1161,8 @@ export class Replayer {
                 this.lastMouseDownEvent = null;
               }
               this.mousePos = {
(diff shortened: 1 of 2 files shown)


Juice10:
Good catch! The fix does break some typing cases though, could you fix those then I can merge this PR?

avillegasn:
Fixed!

Yuyz0112 (approved):

Issue #1379, opened 2024-01-02
[Bug]: addHoverClass creating invalid css


dengelke:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb-snapshot

### Version

2.0.0-alpha.11

### Expected Behavior

Styling should work

### Actual Behavior

Replay is crashing due to stylesheet becoming invalid on replay from the addHoverClass function

### Steps to Reproduce

Is you have a selector like 
```
[_nghost-ng-c4172599085]:not(.fit-content).aim-select:hover:not(:disabled, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--disabled, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--invalid, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--active) { border-color: rgb(84, 84, 84); }
``` 

then it will be replaced in the `addHoverClass` by

``` 
[_nghost-ng-c4172599085]:not(.fit-content).aim-select:hover:not(:disabled, [_nghost-ng-c4172599085]:not(.fit-content).aim-select.\:hover:not(:disabled, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--disabled, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--invalid, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--active) {
        border-color: rgb(84, 84, 84);
    }
```

Which is invalid css.

It looks to be something to do with the logic in the parsing, where the selectors looks like

```[_nghost-ng-c4172599085]:not(.fit-content).aim-select:hover:not(:disabled```

Which is then getting fed into the addHoverClass function

jaj1014:
I'm having a similar issue to the one described above. Here is some more detail that I was able to gather.

[...]

PR #1380, opened 2024-01-02
Fix for #1379

dengelke:
Checks if parentheses are balanced for the selectors


Juice10:
Hi @dengelke, thanks for identifying this issue and creating a fix for it. I was looking through the code and I think this isn't error proof yet unfortunately.
It is possible to add all kinds of parentheses inside strings in attribute matchers https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors so that would fail for this test case:

```
    it('check that parentheses are balanced and ignore parentheses inside strings', () => {
       const cssText = 'a:not(a[data-selector="("]):not(:disabled, div:not(.fit-content)) { border-color: rgb(84, 84, 84); }';
       expect(addHoverClass(cssText, cache)).toEqual(cssText);
     });
 ```

dengelke:
@Juice10 fair, I think realistically the issue is with the css parsing which doesn't parse `[_nghost-ng-c4172599085]:not(.fit-content).aim-select:hover:not(:disabled, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--disabled, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--invalid, [_nghost-ng-c4172599085]:not(.fit-content).aim-select--active) { border-color: rgb(84, 84, 84); }
` correctly 

Will see if I can come up with a better solution

jaj1014:
It looks like this PR is to address https://github.com/rrweb-io/rrweb/issues/1379. 

I added a new comment to that issue re: a similar issue w/ attribute selectors and css parse() failing to properly handle selectors that have curly brackets in them. If you're looking to update parse() in css, it might be helpful info for other cases to consider.

Issue #1385, opened 2024-01-04
[Bug]: maskInputFn is ignored during the creation of the full snapshot


ababik:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

2.0.0-alpha.11

### Expected Behavior

The `maskInputFn` setting must be applied for masking during the creation of the full snapshot.

### Actual Behavior

The `maskInputFn` setting is ignored.

### Steps to Reproduce

1. Create a simple page that contains an `input` with some value. E.g. 
```
<body>
    <input type="text" value="some text value">
</body>
```
2. Configure recorder to mask all elements except specified in the  `maskInputFn`
```
maskAllInputs: true,
maskInputFn(value, element) {
  // ... the logic to skip masking for our text input
  return value;
}
``` 
3. Record the page loading and replay events.
The initial input's text is masked in the recorded video. The element only gets unmasked if you start editing its value.

### Additional Information

I checked the code and it looks like `maskInputFn` setting is ignored for the full snapshot creation.

PR #1386, opened 2024-01-04
Fix maskInputFn is ignored during the creation of the full snapshot

ababik:
Fix for https://github.com/rrweb-io/rrweb/issues/1385

### DIFF
--- a/.changeset/beige-olives-roll.md
+++ b/.changeset/beige-olives-roll.md
@@ -3,5 +3,4 @@
 "rrweb": patch
 ---
 
-Fix maskInputFn is ignored during the creation of the full snapshot
-Fix maskInputFn type in `snapshot`
+Fix that the optional `maskInputFn` was being accidentally ignored during the creation of the full snapshot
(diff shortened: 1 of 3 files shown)


Juice10 (approved):

lewgordon-amplitude:
@ababik I saw the tests are failing. Any plans to fix them? I'm happy to help.

eoghanmurray:
@Juice10 was this just a copy/paste error on your part do you recall?

eoghanmurray:
For some context:

 - maskInputFn was added first in #409 
 - maskTextFn was added subsequently in #540 (motivated by #502)

At that point rrweb-snapshot was a separate repository.
(I was trying to find out if there was ever an intention for one to fallback to the other if one is defined and the other is not, but I can't see that ever being the case)

The default versions of these two functions are slightly different in that the `maskTextFn` doesn't replace whitespace in the content with '*' (stars), so preserves information about the original text, whereas `maskInputFn` replaces the entire value with stars (like a password field).

eoghanmurray (approved):

PR #1387, opened 2024-01-04
perf(snapshot): avoid recreate element `a` every time

H4ad:
Avoid recreating the same element every time, instead, we cache and we just update the element.

Before: 779k ops/s
After: 860k ops/s

Benchmark: https://jsbench.me/ktlqztuf95/1

On Chrome:
image

On Firefox:
image

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -218,7 +218,7 @@ function getHref(doc: Document, customHref?: string) {
   if (!customHref) {
     customHref = '';
   }
-  // note: using `new URL` is slower https://jsbench.me/uqlud17rxo/1
+  // note: using `new URL` is slower. See #1434 or https://jsbench.me/uqlud17rxo/1
   a.setAttribute('href', customHref);
   return a.href;
 }
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -195,23 +195,34 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
   return output.join(', ');
 }
 
+const cachedDocument = new WeakMap<Document, HTMLAnchorElement>();
+
 export function absoluteToDoc(doc: Document, attributeValue: string): string {
   if (!attributeValue || attributeValue.trim() === '') {
     return attributeValue;
   }
-  const a: HTMLAnchorElement = doc.createElement('a');
-  a.href = attributeValue;
-  return a.href;
+
+  return getHref(doc, attributeValue);
 }
 
 function isSVGElement(el: Element): boolean {
   return Boolean(el.tagName === 'svg' || (el as SVGElement).ownerSVGElement);
 }
 
-function getHref() {
-  // return a href without hash
-  const a = document.createElement('a');
-  a.href = '';
+function getHref(doc: Document, customHref?: string) {
+  let a = cachedDocument.get(doc);
+
+  if (!a) {
+    a = doc.createElement('a');
+    cachedDocument.set(doc, a);
+  }
+
+  if (!customHref) {
+    customHref = '';
+  }
+
+  a.setAttribute('href', customHref);
+
   return a.href;
 }
 
@@ -243,7 +254,7 @@ export function transformAttribute(
   } else if (name === 'srcset') {
     return getAbsoluteSrcsetString(doc, value);
   } else if (name === 'style') {
-    return absoluteToStylesheet(value, getHref());
+    return absoluteToStylesheet(value, getHref(doc));
   } else if (tagName === 'object' && name === 'data') {
     return absoluteToDoc(doc, value);
   }
@@ -505,6 +516,7 @@ function serializeNode(
       });
     case n.TEXT_NODE:
       return serializeTextNode(n as Text, {
+        doc,
         needsMask,
         maskTextFn,
         rootId,
@@ -535,6 +547,7 @@ function getRootId(doc: Document, mirror: Mirror): number | undefined {
 function serializeTextNode(
   n: Text,
   options: {
+    doc: Document;
     needsMask: boolean | undefined;
     maskTextFn: MaskTextFn | undefined;
     rootId: number | undefined;
@@ -566,7 +579,7 @@ function serializeTextNode(
         n,
       );
     }
-    textContent = absoluteToStylesheet(textContent, getHref());
+    textContent = absoluteToStylesheet(textContent, getHref(options.doc));
   }
   if (isScript) {
     textContent = 'SCRIPT_PLACEHOLDER';
@@ -660,7 +673,7 @@ function serializeElementNode(
       (n as HTMLStyleElement).sheet as CSSStyleSheet,
     );
     if (cssText) {
-      attributes._cssText = absoluteToStylesheet(cssText, getHref());
+      attributes._cssText = absoluteToStylesheet(cssText, getHref(doc));
     }
   }
   // form fields
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -218,6 +218,7 @@ function getHref(doc: Document, customHref?: string) {
   if (!customHref) {
     customHref = '';
   }
+  // note: using `new URL` is slower https://jsbench.me/uqlud17rxo/1
   a.setAttribute('href', customHref);
   return a.href;
 }
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -211,18 +211,14 @@ function isSVGElement(el: Element): boolean {
 
 function getHref(doc: Document, customHref?: string) {
   let a = cachedDocument.get(doc);
-
   if (!a) {
     a = doc.createElement('a');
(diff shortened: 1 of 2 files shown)


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -194,23 +194,34 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
...
-  a.href = '';
+function getHref(doc: Document, customHref?: string) {
+  let a = cachedDocument.get(doc);
+
+  if (!a) {
+    a = doc.createElement('a');
+    cachedDocument.set(doc, a);
+  }
+
+  // this is needed to work with SPA
```
JonasBa, line 219: Can you explain why this is needed for SPA? Folks reading this later might not necessarily have context or understanding why this is necessary :)

H4ad: I will add a comment to link to this answer. 

Basically, the cached `a` link stores the URL when the element is created.
SPAs will not update the page but it can change the URL, this makes the cached `a` link unusable.

In order to prevent that, I need to reset the cache by defining the URL to any value, then I need to reset again to set the URL to `''` to support relative paths like `./image.png` to be correctly displayed.

URLs are not updated if you set the same value, that's why I need to reset the value twice.

JonasBa: Makes perfect sense, I would add something similar as an inline comment

Juice10 (approved):

eoghanmurray:
@JonasBa Could you also test the following possibility:

```
function getHref(doc: Document) {
  // return a href without hash
  const url = new URL(doc.location.href);
  url.hash = ''
  return url.href;
}
```

Some more work would need to be done to support the new `customHref` case so maybe this PR is fine

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -194,23 +194,45 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
...
+
 export function absoluteToDoc(doc: Document, attributeValue: string): string {
   if (!attributeValue || attributeValue.trim() === '') {
     return attributeValue;
   }
-  const a: HTMLAnchorElement = doc.createElement('a');
-  a.href = attributeValue;
-  return a.href;
+
+  return getHref(doc, attributeValue);
```
YunFeng0817, line 204: > @H4ad Could you also test the following possibility:
> 
> ```
> function getHref(doc: Document) {
>   // return a href without hash
>   const url = new URL(doc.location.href);
>   url.hash = ''
>   return url.href;
> }
> ```
> 
> Some more work would need to be done to support the new `customHref` case so maybe this PR is fine

A slight change based on @eoghanmurray 's suggestion
```suggestion

  const url = new Url(attributeValue, doc.location);
  url.hash = '';
  return url.href;
```

eoghanmurray:
One way in which the `new URL` method throws an error, where the existing way doesn't is with `attributeValue = 'http://'`

We might want to use a fallback to the old method in that case.  Also not all browsers/platforms support the `URL` object.

eoghanmurray:
I've a version which should perform just as good without the need for caching in #1434 

@H4ad would you like to run benchmarks on that?

H4ad:
@eoghanmurray Sorry for taking so long to answer, I will run the benchmarks as soon I can and I will bring back the results

eoghanmurray:
(in case you miss the thread further up)

Could you cherry-pick https://github.com/eoghanmurray/rrweb/commit/615e164fe6bf3f48f5766f283e9e94d723cfc7fa

And also rebase or merge master

H4ad:
@eoghanmurray  Done!

YunFeng0817 (approved):

eoghanmurray:
I've added another review comment as I'd like to be able to recreate the SPA scenario via a test which necessitates the cache bust.  I created a test case but was not able to recreate the scenario where the cache bust is needed.

H4ad:
@eoghanmurray did you submit the review?

eoghanmurray:
> I've added another review comment
This is what I was referring to:
https://github.com/rrweb-io/rrweb/pull/1387/files#diff-6434d77e679a3d1e4217dcad03ee6f02551e41dc0b077a06a45020e68e1b4ee3R222

Here is the test I was talking about, which could be cherry-picked in here, but basically it doesn't demonstrate any need for the `./clear-current` stuff, so maybe you can enhance the test case?
https://github.com/eoghanmurray/rrweb/commit/4d521651eb4c27e96a8c786860d44a386411da34

And here's some possible refactoring depending on the outcome of the above:
https://github.com/eoghanmurray/rrweb/commit/38e677a29ff23e8e91fc697aa9e01772140c3171

eoghanmurray:
I propose merging a version of my changes:
https://github.com/rrweb-io/rrweb/commit/38e677a29ff23e8e91fc697aa9e01772140c3171

but without the `This is needed for SPAs.` bit as we don't yet have any evidence that that is necessary.

H4ad:
You are right, I don't know if it was a bug on Chrome when I tested but the behavior that I described in the comment is not true anymore.

I have updated the code to reflect your changes and I also removed those comments/code about SPA.

Thanks for the patience, I was very lazy on this PR.

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -195,23 +195,34 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
...
   return Boolean(el.tagName === 'svg' || (el as SVGElement).ownerSVGElement);
 }
 
-function getHref() {
-  // return a href without hash
-  const a = document.createElement('a');
-  a.href = '';
+function getHref(doc: Document, customHref?: string) {
+  let a = cachedDocument.get(doc);
+
```
eoghanmurray, line 214:
```suggestion
```

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -195,23 +195,34 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
...
+  if (!a) {
+    a = doc.createElement('a');
+    cachedDocument.set(doc, a);
+  }
+
```
eoghanmurray, line 219:
```suggestion
```

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -195,23 +195,34 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
...
+  if (!customHref) {
+    customHref = '';
+  }
+
```
eoghanmurray, line 223:
```suggestion
```

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -195,23 +195,34 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
...
+  a.setAttribute('href', customHref);
+
```
eoghanmurray, line 225:
```suggestion
```

eoghanmurray (approved):

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -195,23 +195,30 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
...
+function getHref(doc: Document, customHref?: string) {
+  let a = cachedDocument.get(doc);
+  if (!a) {
+    a = doc.createElement('a');
+    cachedDocument.set(doc, a);
+  }
+  if (!customHref) {
+    customHref = '';
+  }
+  a.setAttribute('href', customHref);
```
eoghanmurray, line 221:
```suggestion
  // note: using `new URL` is slower https://jsbench.me/uqlud17rxo/1
  a.setAttribute('href', customHref);
```

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -195,23 +195,31 @@ function getAbsoluteSrcsetString(doc: Document, attributeValue: string) {
...
+function getHref(doc: Document, customHref?: string) {
+  let a = cachedDocument.get(doc);
+  if (!a) {
+    a = doc.createElement('a');
+    cachedDocument.set(doc, a);
+  }
+  if (!customHref) {
+    customHref = '';
+  }
+  // note: using `new URL` is slower https://jsbench.me/uqlud17rxo/1
```
eoghanmurray, line 221:
```suggestion
  // note: using `new URL` is slower. See #1434 or https://jsbench.me/uqlud17rxo/1
```

eoghanmurray:
Actually I'll just create a new PR for that test case

eoghanmurray:
@H4ad I found a bug where a `blob:` style url would be persisted on the cached anchor element and would incorrect return values on subsequent calls to getHref.  Fix in #1453 

But could you re-review this PR in view of those single page app bugs you previously encountered?

H4ad:
Sure, I will try leave a review as soon as possible.

eoghanmurray:
Just to note, using `URL.canParse` can make #1434 more competitve; only ~20% slower in jsbench:

```
function absoluteToDocCanParse(doc, attributeValue) {
// todo: do we need check if attributeValue is ''? 
    if (URL && URL.canParse(attributeValue)) {
	return new URL(attributeValue, doc.location.href).href;
    } else {
	const a = doc.createElement('a');
	a.setAttribute('href', attributeValue);
	return a.href;
    }
}
```

Issue #1395, opened 2024-01-09
[Bug]: grid template areas are not inlined correctly


pauldambra:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

latest v2

### Expected Behavior

that the grid-template-area is captured correctly

### Actual Behavior

A customer site uses [grid template areas](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-areas) in their CSS. When playing back we see the content is not aligned as expected. When viewing the working site things appear in the correct place.

On checking the captured CSS it does not match the source

The source has

```
#wrapper {
  display: grid;
    width: 100%;
    height: 100%;
    grid-template-columns: repeat(2, 1fr);
    grid-template-rows: max-content;
    grid-template-areas:
        "header header"
        "main main"
        "footer footer";
    margin: 0px auto;
}
```

but the captured CSS has:

```
#wrapper { 
display: grid; 
width: 100%; 
height: 100%; 
grid-template: \"header header\" max-content / repeat(2, 1fr); 
margin: 0px auto; 
}
```

you can see the CSS has been collapsed to a short-hand `grid-template` and lost detail from the original

### Steps to Reproduce

You can see a working example of this here: https://jsfiddle.net/pauldambra/oLtcrjm4/10/ showing that the captured CSS has been mangled

pauldambra:
ah, looking in safari and firefox. this is another Chrome is the IE8 of the new millenium since it seems correct in those other browsers 🤣

debbyglance:
We are also impacted by this bug

jaj1014:
Resurfacing that this is still an issue. We were using the fix from @pauldambra [PR](https://github.com/rrweb-io/rrweb/pull/1396) that was closed, but ran into an issue with the regex that solution used. I have opened [another PR](https://github.com/rrweb-io/rrweb/pull/1586) that tweaks the original to address the issue.

PR #1396, opened 2024-01-10
fix: grid template is parsed incorrectly by chrome

pauldambra:
fixes: https://github.com/rrweb-io/rrweb/issues/1395

Chrome is not correctly providing `cssText` when the style includes a `grid-template-area` rule. This PR attempts to detect and correct this...

(it's my first PR against the repo, so have opened early to get feedback and guidance :))

### DIFF
--- a/.changeset/stale-llamas-divide.md
+++ b/.changeset/stale-llamas-divide.md
@@ -1,0 +1,5 @@
+---
+'rrweb-snapshot': patch
+---
+
+adds a correction for when chrome passes an incorrect grid-template-area cssText during snapshot
--- a/packages/rrweb-snapshot/src/utils.ts
+++ b/packages/rrweb-snapshot/src/utils.ts
@@ -1,16 +1,16 @@
 import {
+  documentNode,
+  documentTypeNode,
+  elementNode,
   idNodeMap,
+  IMirror,
   MaskInputFn,
   MaskInputOptions,
   nodeMetaMap,
-  IMirror,
-  serializedNodeWithId,
-  serializedNode,
   NodeType,
-  documentNode,
-  documentTypeNode,
+  serializedNode,
+  serializedNodeWithId,
   textNode,
-  elementNode,
 } from './types';
 
 export function isElement(n: Node): n is Element {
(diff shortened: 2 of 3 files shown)


jaj1014:
Came across this issue and it's one I've seen, too and wanted to add some additional detail to help illustrate the issue.

I put together a [gist](https://gist.github.com/jaj1014/d09b73ab195c9b7152806317de3692e6) and [debug video](https://rrwebdebug.com/play/index.html?url=https%3A%2F%2Fgist.github.com%2Fjaj1014%2Fd09b73ab195c9b7152806317de3692e6&version=2.0.0-alpha.11&virtual-dom=on&play=on) w/ a sample recording that demonstrates the issue.

For reference, this is what the layout _should_ look like.

Screenshot 2024-01-25 at 5 24 14 AM

eoghanmurray:
If you add a changeset file (see changeset-bot comment on this page) it should re-trigger the tests.
(this PR was created without permissions for me to do this for you myself)

pauldambra:
> (this PR was created without permissions for me to do this for you myself)

Ah, sorry. How do I create with permissions?

eoghanmurray:
In addition to the note above about the `repeat()` function in the tests, I'm wondering whether this is currently a bug in Chrome/Chromium or whether this PR is a fix for older versions of browsers?

eoghanmurray:
> Ah, sorry. How do I create with permissions?

There's a ' Allow edits and access to secrets by maintainers' checkbox in the rhs column

pauldambra:
> You might have missed my review query;
> "This doesn't appear to be valid grid-template syntax"

I did... in fact, I can't see it anywhere 🙈 

This was quite a while ago so my brain has jettisoned all context 🤣 

We've had a single user mention it, and they're not chasing us. So, I'm just as happy to get validation that the approach/files changed are about right (even if the actual implementation isn't desirable) and close the PR

eoghanmurray (commented):
In case you missed the questions

eoghanmurray:
Also, do you recall whether the problem was manifesting in a `<link rel="stylesheet">` context, or in a `<style>` element?
If the latter, we may be able to resolve it another way.

pauldambra:
I don't remember at all and we've not had further reports... since I captured a fiddle to demo the bug at reporting time I'll defer to https://github.com/rrweb-io/rrweb/pull/1458 and check if that fixes this

We can re-open this if need be then but let's not add code for an very rare case if we can avoid it :)

eoghanmurray:
OK there may be a new approach possible; see 
https://github.com/rrweb-io/rrweb/pull/1357#issuecomment-2102318235

Basically we stringify `.sheet.rules` during a snapshot as if any programmatic modifications have taken place (rule inserts/deletions), then that will have them, whereas the raw text does not.  So if we could verify that a 'normalized' `.sheet.rules` matches a normalized rawtext (not a simple task), then we could just go ahead and use the raw text.  

A job for another day, but if you think you can salvage the tests out of this one (I mentioned some seemingly incorrect syntax) then we could commit that albeit commented out with a TODO until we can fix this.

jaj1014:
I've opened a new PR that builds on the work from this one. We had been using the code from the closed PR, but hit an issue w/ the regex that was in the original work so I've updated to address that in the [new one](https://github.com/rrweb-io/rrweb/pull/1586)

PR #1398, opened 2024-01-16
perf(genadds) traverse children in reverse order

mdellanoce:
I noticed when you add a deep/complex tree as part of a mutation, a large portion of the tree ends up being added to the addList. This is because genAdds traverses the children of each node in order, but serialization in pushAdd wants each node to have a serialized parent _and_ nextSibling. Reversing the order of traversal in genAdds makes it more likely to avoid adding nodes to the addList.

I ran the benchmarks before/after and this was a pretty small improvement across the board.

### DIFF
--- a/.changeset/unlucky-flowers-remember.md
+++ b/.changeset/unlucky-flowers-remember.md
@@ -1,0 +1,5 @@
+---
+'rrweb': patch
+---
+
+traverse childNodes in reverse order to match desired serialization order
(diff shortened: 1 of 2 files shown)


mdellanoce:
this _might_ have an impact on performance when things do make it to the addList though, since the node order is reversed, and addList is traversed from the tail node to the head. I haven't found a case where that happens yet...

eoghanmurray:
I think this was previously discussed with the core team and we were hoping to have benchmarks in place before merging, but that hasn't happened.
The only possible objection I can think of is whether this could makes things worse for different types of data sets, but really your write-up makes a great case that the existing approach is worst-case.

colingm:
For now thinking we will abandon this PR, it seems to most likely be addressed anyways by #1277 and so we can always revisit it after that one goes through

PR #1401, opened 2024-01-24
Fix for test cases mentioned in #1379

### LINKED ISSUE #1379: [Bug]: addHoverClass creating invalid css

dengelke:
Update of parsing function using selector from [here](https://github.com/NxtChg/pieces/blob/master/js/css_parser/css_parse.js) and update to logic for `,` inside `()` and associated tests to fix #1379

### DIFF
--- a/.changeset/rich-dots-lay.md
+++ b/.changeset/rich-dots-lay.md
@@ -1,0 +1,5 @@
+---
+'rrweb-snapshot': patch
+---
+
+Fix css parsing errors
--- a/packages/rrweb-snapshot/src/css.ts
+++ b/packages/rrweb-snapshot/src/css.ts
@@ -431,22 +431,64 @@ export function parse(css: string, options: ParserOptions = {}) {
    */
 
   function selector() {
-    const m = match(/^([^{]+)/);
+
+    whitespace(); while(css[0] == '}'){ error('extra closing bracket'); css = css.slice(1); whitespace(); }
+
+		const m = match(/^(("(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[^{])+)/);
     if (!m) {
       return;
     }
     /* @fix Remove all comments from selectors
      * http://ostermiller.org/findcomment.html */
-    return trim(m[0])
+    const cleanedInput = m[0].trim()
       .replace(/\/\*([^*]|[\r\n]|(\*+([^*/]|[\r\n])))*\*\/+/g, '')
+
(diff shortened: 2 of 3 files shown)


jaj1014:
got around to testing this against the test recording I created that has the `{}` in the attribute value of the selectors and it seems to be working! It plays back w/ all the correct hovering and such.

thanks @dengelke ! - hopefully this can get into the queue for reviewing the solution 🤞

Juice10:
@dengelke can you add a changeset (see other comment above) and add the link of where you found the css parser to the code so we can keep track of where this came from if it ever needs updating?

dengelke:
@Juice10 updated the changeset and added a comment in the code

Juice10 (approved):

eoghanmurray:
@dengelke would you be able to take a quick look at #1440 which covered the same ground.  I've merged the remainder of it last night into trunk; the remaining case was to do with handling single/double quotes (attr selectors).

PR #1402, opened 2024-01-24
fix: The needsMask check should also be ignored when the childNode array length is equal to 0

Paulhejia:
When checking needsMask, it is not rigorous to determine whether _n.childNode_ exists, but also to determine whether the array length is greater than 0

### DIFF
--- a/.changeset/lazy-bananas-peel.md
+++ b/.changeset/lazy-bananas-peel.md
@@ -1,0 +1,5 @@
+---
+"rrweb-snapshot": patch
+---
+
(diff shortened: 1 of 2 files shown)


Juice10:
@Paulhejia looks like this breaks in certain cases (see failing tests), could you fix them?

eoghanmurray:
Thanks @Paulhejia 
These failing tests exposed the fact that this optimisation that I had (incorrectly) written actually introduced a bug.
#1512 reintroduces the optimisation in a better way so I'll close this version, with thanks.

PR #1408, opened 2024-02-12
Expose constant SKIP_TIME_THRESHOLD as skipTimeThreshold in replayer

avillegasn:
There is this hardcoded constant `SKIP_TIME_THRESHOLD` to define the amount of time in milliseconds of an inactive period. By default it is set to 10 seconds, but I believe this should be exposed as a configuration option of the replayer so that users may configure it to fit their needs.

This pull request exposes `SKIP_TIME_THRESHOLD` as `skipTimeThreshold`.


guide.md
```diff
@@ -301,6 +301,7 @@ The replayer accepts options as its constructor's second parameter, and it has t
 | root                | document.body | the root element of replayer                                                                                                                                                                                                   |
 | loadTimeout         | 0             | timeout of loading remote style sheet                                                                                                                                                                                          |
 | skipInactive        | false         | whether to skip inactive time                                                                                                                                                                                                  |
+| skipTimeThreshold   | 10000         | a threshold of inactive time in milliseconds                                                                                                                                                                                                  |
```
eoghanmurray, line 304:
```suggestion
| skipTimeThreshold   | 10000         | the threshold in milliseconds for what should be considered an inactive period                                                                                                                                                                                                  |
```

eoghanmurray (approved):

eoghanmurray:
Thanks for your contribution!  I don't use the svelte player but this looks good.

May I suggest a better name for the config setting?

Maybe `minInactivePeriod` or `inactivePeriodThreshold` rather than `skipTimeThreshold`?

avillegasn:
> Thanks for your contribution! I don't use the svelte player but this looks good.
> 
> May I suggest a better name for the config setting?
> 
> Maybe `minInactivePeriod` or `inactivePeriodThreshold` rather than `skipTimeThreshold`?

I did the change to `inactivePeriodThreshold`.

eoghanmurray (approved):

PR #1416, opened 2024-02-23
Feature/player style

warjiang:
1.  declare style for player, avoid style reset
2. add playerClassName props for style override


YunFeng0817:
@warjiang 
Thanks for your contribution. Can you please describe more about the new config about scenarios and usage?
Some additional actions are adding the new option and usage to the corresponding position in the markdown doc and adding a changeset log.

Issue #1421, opened 2024-03-06
[Bug]: Calling 'createImageBitmap' throws an exception


daibhin:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

2.0.0-alpha.11

### Expected Behavior

Canvas recording can be enabled without exception firing

### Actual Behavior

An exception is thrown within our SDK
```
Uncaught (in promise) DOMException: Failed to execute 'createImageBitmap' on 'Window': The source image width is 0.
    at e.<anonymous> (https://us-assets.i.posthog.com/static/recorder-v2.js?v=1.110.0:1:74170)
    at l (https://us-assets.i.posthog.com/static/recorder-v2.js?v=1.110.0:1:1844)
    at Generator._invoke (https://us-assets.i.posthog.com/static/recorder-v2.js?v=1.110.0:1:1597)
    at Generator.next (https://us-assets.i.posthog.com/static/recorder-v2.js?v=1.110.0:1:2207)
    at https://us-assets.i.posthog.com/static/recorder-v2.js?v=1.110.0:1:61073
    at new Promise (<anonymous>)
    at Qe (https://us-assets.i.posthog.com/static/recorder-v2.js?v=1.110.0:1:60818)
    at https://us-assets.i.posthog.com/static/recorder-v2.js?v=1.110.0:1:73736
    at Array.forEach (<anonymous>)
    at e (https://us-assets.i.posthog.com/static/recorder-v2.js?v=1.110.0:1:73708)
```

### Steps to Reproduce

Exact steps are not clear to me but the exception seems to suggest that it is caused by a canvas with a width of 0 occurring on the page

PR #1422, opened 2024-03-07
fix: createImageBitmap throws DOMException if source is 0

marandaneto:
Tries to fix https://github.com/rrweb-io/rrweb/issues/1421


pauldambra (approved):
looks good to me...

packages/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -207,6 +207,12 @@ export class CanvasManager {
               context.clear(context.COLOR_BUFFER_BIT);
             }
           }
+
+          // The browser throws if the canvas is 0 in size
+          // Uncaught (in promise) DOMException: Failed to execute 'createImageBitmap' on 'Window': The source image width is 0.
+          // Assuming the same happens with height
+          if (canvas.width === 0 || canvas.height === 0) return;
+
```
p-mazhnik, line 215: I believe this should be at the beginning of the function, surely above the `snapshotInProgressMap.set(id, true);`. Otherwise, canvas will never be recorded after its size is changed

p-mazhnik (approved):

marandaneto:
@Juice10 can we merge this? just fixed the formatting issue, thanks for the patch though.

Juice10 (approved):

Juice10:
@marandaneto if it passes CI (running now) I'll happily merge it

Juice10:
@marandaneto done!

Issue #1425, opened 2024-03-14
[Feature Request]: Setting name as data.payload.name instead of data.tag


Archish27:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a feature request that matches the one I want to file, without success.

### What package is this feature request for?

rrweb

### Problem Description

Idea is to have two multiple custom names under same tag.

Example: 
tag name as `error` & event name as `Some error occurred in application` 

Why what is the idea behind this?
Currently there is no way to group event names under same tag. By doing so we can define same color to specific tag.

### Proposed Solution

instead of this
```javascript
const customEvent = {
          name: event.data.tag,
          background: tags[event.data.tag] || 'rgb(73, 80, 246)',
          position: `${position(start, end, event.timestamp)}%`,
        };
customEvents.push(customEvent);
```

to 
```javascript
const customEvent = {
          name: event.data.name,
          background: tags[event.data.tag] || 'rgb(73, 80, 246)',
          position: `${position(start, end, event.timestamp)}%`,
 };
customEvents.push(customEvent);
```

### Alternatives Considered

N/A

eoghanmurray:
see #1426

PR #1426, opened 2024-03-14
chore: Grouping event names under tag

Archish27:
#1425

### DIFF
--- a/packages/rrweb-player/src/Controller.svelte
+++ b/packages/rrweb-player/src/Controller.svelte
@@ -91,7 +91,7 @@
        */
       if (event.type === EventType.Custom) {
         const customEvent = {
-          name: event.data.tag,
+          name: event.data.payload.name,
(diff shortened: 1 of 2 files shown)


packages/rrweb-player/src/Controller.svelte
```diff
@@ -91,7 +91,7 @@
        */
       if (event.type === EventType.Custom) {
         const customEvent = {
-          name: event.data.tag,
+          name: event.data.payload.name,
```
Juice10, line 94: Since the payload can be anything, `name` is not a guaranteed attribute on `payload` so this is not safe to do

Juice10 (changes requested):

eoghanmurray:
This is something that would be fine to do in your own codebase/implementation as you can enforce that all custom events have names but not really a good candidate for a general PR as-is; every implementation might have different goals on what to display.

PR #1431, opened 2024-03-21
Fix some css issues with :hover and rewrite max-device-width

eoghanmurray:
- Fix: We weren't recursing into media queries (or @supports etc.) to rewrite `:hover` pseudoclasses
 - Extension for replay: replace `min-device-width` and similar with `min-width` as the former looks out at the browser viewport whereas we need it to look at the replayer iframe viewport
 - Fix a gap in the css replayer caching if neither of the above were present in the css file

### DIFF
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -1,4 +1,4 @@
-import { parse } from './css';
+import { StyleRules, Rule, Media, parse } from './css';
 import {
   serializedNodeWithId,
   NodeType,
@@ -80,7 +80,7 @@ export function adaptCssForReplay(cssText: string, cache: BuildCache): string {
 
   const selectors: string[] = [];
   const medias: string[] = [];
-  function getSelectors(rule: any) {
+  function getSelectors(rule: StyleRules | Rule | Media) {
     if ('selectors' in rule) {
       (rule.selectors || []).forEach((selector: string) => {
         if (HOVER_SELECTOR.test(selector)) {
@@ -88,7 +88,7 @@ export function adaptCssForReplay(cssText: string, cache: BuildCache): string {
         }
       });
     }
-    if ('media' in rule && MEDIA_SELECTOR.test(rule.media)) {
+    if ('media' in rule && rule.media && MEDIA_SELECTOR.test(rule.media)) {
       medias.push(rule.media);
     }
     if ('rules' in rule) {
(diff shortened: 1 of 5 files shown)


packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -77,35 +79,61 @@ export function addHoverClass(cssText: string, cache: BuildCache): string {
   }
 
   const selectors: string[] = [];
-  ast.stylesheet.rules.forEach((rule) => {
+  const medias: string[] = [];
+  function getSelectors(rule: any) {
```
Juice10, line 83: There is some types you could use from `./css.ts` to type this.
```suggestion
  function getSelectors(rule: StyleRules) {
```

Juice10: You might need to change the name of `rule` here though --^

packages/rrweb-snapshot/test/rebuild.test.ts
```diff
@@ -81,47 +85,90 @@ describe('rebuild', function () {
...
+      expect(adaptCssForReplay(cssText, cache)).toEqual(
+        `ul li.specified a:hover img, ul li.specified a.\\:hover img,
+ul li.multiline
+b:hover
+img, ul li.multiline
+b.\\:hover
+img,
+ul li.specified c:hover img, ul li.specified c.\\:hover img {
+  color: white
+}`,
```
Juice10, line 122: I think you can add whitespace here to make the tests a little less messy, unless I'm missing something it should be fine.
```suggestion
         ul li.multiline
         b:hover
         img,
         ul li.specified c:hover img {
           color: white
         }`;
      expect(adaptCssForReplay(cssText, cache)).toEqual(
        `ul li.specified a:hover img, ul li.specified a.\\:hover img,
         ul li.multiline
         b:hover
         img, ul li.multiline
         b.\\:hover
         img,
         ul li.specified c:hover img, ul li.specified c.\\:hover img {
           color: white
         }`,
```

Juice10 (changes requested):
Good catch @eoghanmurray! Found a little typing improvement for you, but apart from that this is looking ready for release!

packages/rrweb-snapshot/src/rebuild.ts
eoghanmurray: This is actually using css.ts which has it's own interface.
I wasted a half hour trying to figure out the typings.  I can't see the benefit of typings, it has to iterate over note only StyleRules types but implementations of Node interface (see css.ts)

I get the following error if I add `Node` there:

```
src/rebuild.ts:85:30 - error TS2339: Property 'forEach' does not exist on type '{}'.

85       (rule.selectors || []).forEach((selector: string) => {
```

Juice10: Maybe I can hop on a call with you, we can try and figure it out together?

Juice10 (approved):

PR #1432, opened 2024-03-22
Full overhawl of video & audio playback to make it more complete

Juice10:
This pull request adds support for capturing media attributes such as playbackRate, muted, and loop in the rrweb-snapshot library. It also includes updates to the video playback and fixes many bugs related to pausing, playing, skipping, muting, and playbackRate on media elements in rrweb itself.


eoghanmurray (approved):
This is great work; I haven't had a chance to fully test this in production, but I believe it is better in trunk than out of it as it provides missing functionality.

PR #1437, opened 2024-04-05
Single style capture

eoghanmurray:
Prep PR for async <style> serialization via assets: refactor stringifyStylesheet to happen in a single place during initial snapshot.

- initial motivation was to change the approach on stringifying <style> elements to do so in a single place
 - I then learned that we need the style content present on child text elements, in order to support mutation
 - this is still supported, however the css content is now preferably stored in parent <style> element in that _cssText attribute, rather than in the (natural) position in the child text elements of that element.  This is to support the stylesheet-assets PR which will delay population of the _cssText attribute, as in both cases the snapshot will be recorded with empty child text elements
 - on rebuild, this content can now be 'spread' out to child text nodes (usually there's only one!) in order to give a correct target for any further text mutations on these children.
 - there is a new complex but comprehensive test 'can record and replay style text mutations' which covers all of this; it has been designed to ensure that it fails if css text is populated in the wrong way (e.g. putting the entire cssText on the first of many text children and leaving the others blank)

### DIFF
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -90,8 +90,8 @@ export function applyCssSplits(
   hackCss: boolean,
   cache: BuildCache,
 ): void {
-  let childTextNodes: serializedTextNodeWithId[] = [];
-  for (let scn of n.childNodes) {
+  const childTextNodes: serializedTextNodeWithId[] = [];
+  for (const scn of n.childNodes) {
     if (scn.type === NodeType.Text) {
       childTextNodes.push(scn);
     }
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -2,6 +2,7 @@ import { mediaSelectorPlugin, pseudoClassPlugin } from './css';
 import {
   type serializedNodeWithId,
   type serializedElementNodeWithId,
+  type serializedTextNodeWithId,
   NodeType,
   type tagMap,
   type elementNode,
@@ -80,41 +81,41 @@ export function createCache(): BuildCache {
 }
 
 /**
- * undo findCssTextSplits
+ * undo splitCssText/markCssSplits
  * (would move to utils.ts but uses `adaptCssForReplay`)
  */
 export function applyCssSplits(
   n: serializedElementNodeWithId,
   cssText: string,
-  cssTextSplits: number[],
   hackCss: boolean,
   cache: BuildCache,
 ): void {
-  const lenCheckOk =
-    cssTextSplits.length &&
-    cssTextSplits[cssTextSplits.length - 1] === cssText.length;
-  for (let j = n.childNodes.length - 1; j >= 0; j--) {
-    const scn = n.childNodes[j];
-    let ix = 0;
-    if (cssTextSplits.length > j && j > 0) {
-      ix = cssTextSplits[j - 1];
-    }
+  let childTextNodes: serializedTextNodeWithId[] = [];
+  for (let scn of n.childNodes) {
     if (scn.type === NodeType.Text) {
-      let remainder = '';
-      if (ix !== 0 && lenCheckOk) {
-        remainder = cssText.substring(0, ix);
-        cssText = cssText.substring(ix);
-      } else if (j > 0) {
-        continue;
-      }
-      if (hackCss) {
-        cssText = adaptCssForReplay(cssText, cache);
-      }
-      // id will be assigned when these child nodes are
-      // iterated over in buildNodeWithSN
-      scn.textContent = cssText;
-      cssText = remainder;
+      childTextNodes.push(scn);
     }
+  }
+  for (let i = 0; i < childTextNodes.length; i++) {
+    let remainder = '';
+    const scn = childTextNodes[i];
+    if (i !== childTextNodes.length - 1) {
+      const ix = cssText.indexOf('/* rr_split */');
+      if (ix !== -1) {
+        remainder = cssText.substring(ix + '/* rr_split */'.length);
+        cssText = cssText.substring(0, ix);
+      }
+    } else {
+      // todo: replaceAll after lib.es2021
+      cssText = cssText.replace(/\/\* rr_split \*\//g, '');
+    }
+    if (hackCss) {
+      cssText = adaptCssForReplay(cssText, cache);
+    }
+    // id will be assigned when these child nodes are
+    // iterated over in buildNodeWithSN
+    scn.textContent = cssText;
+    cssText = remainder;
   }
   if (cssText.length) {
     // something has gone wrong
@@ -128,8 +129,8 @@ export function applyCssSplits(
  * browser sees and serialize this to a special _cssText attribute, blanking
  * out any text nodes. This function reverses that and also handles cases where
  * there were no textNode children present (dynamic css/or a <link> element) as
- * well as multiple textNodes (`cssTextSplits`), which need to be repopulated
- * in case they are modified by subsequent mutations.
+ * well as multiple textNodes, which need to be repopulated (based on presence of
+ * a special `rr_split` marker in case they are modified by subsequent mutations.
  */
 export function buildStyleNode(
   n: serializedElementNodeWithId,
@@ -143,13 +144,7 @@ export function buildStyleNode(
 ) {
   const { doc, hackCss, cache } = options;
   if (n.childNodes.length) {
-    let cssTextSplits: number[] = [];
-    if (n.attributes._cssTextSplits) {
-      cssTextSplits = n.attributes._cssTextSplits
-        .split(' ')
-        .map((s) => parseInt(s));
-    }
-    applyCssSplits(n, cssText, cssTextSplits, hackCss, cache);
+    applyCssSplits(n, cssText, hackCss, cache);
   } else {
     if (hackCss) {
       cssText = adaptCssForReplay(cssText, cache);
@@ -235,8 +230,6 @@ function buildNode(
 
         if (name.startsWith('rr_')) {
           specialAttributes[name] = value;
-          continue;
-        } else if (name === '_cssTextSplits') {
           continue;
         }
 
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -96,26 +96,24 @@ export function applyCssSplits(
       childTextNodes.push(scn);
     }
   }
+  const cssTextSplits = cssText.split('/* rr_split */');
+  while (
+    cssTextSplits.length > 1 &&
+    cssTextSplits.length > childTextNodes.length
+  ) {
+    // unexpected: remerge the last two so that we don't discard any css
+    cssTextSplits.splice(-2, 2, cssTextSplits.slice(-2).join(''));
+  }
   for (let i = 0; i < childTextNodes.length; i++) {
-    let remainder = '';
-    const scn = childTextNodes[i];
-    if (i !== childTextNodes.length - 1) {
-      const ix = cssText.indexOf('/* rr_split */');
-      if (ix !== -1) {
-        remainder = cssText.substring(ix + '/* rr_split */'.length);
-        cssText = cssText.substring(0, ix);
-      }
-    } else {
-      // todo: replaceAll after lib.es2021
-      cssText = cssText.replace(/\/\* rr_split \*\//g, '');
+    const childTextNode = childTextNodes[i];
+    const cssTextSection = cssTextSplits[i];
+    if (childTextNode && cssTextSection) {
+      // id will be assigned when these child nodes are
+      // iterated over in buildNodeWithSN
+      childTextNode.textContent = hackCss
+        ? adaptCssForReplay(cssTextSection, cache)
+        : cssTextSection;
     }
-    if (hackCss) {
-      cssText = adaptCssForReplay(cssText, cache);
-    }
-    // id will be assigned when these child nodes are
-    // iterated over in buildNodeWithSN
-    scn.textContent = cssText;
-    cssText = remainder;
   }
 }
 
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -1,6 +1,7 @@
 import { mediaSelectorPlugin, pseudoClassPlugin } from './css';
 import {
   type serializedNodeWithId,
+  type serializedElementNodeWithId,
   NodeType,
   type tagMap,
   type elementNode,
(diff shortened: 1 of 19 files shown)


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -463,6 +465,7 @@ function serializeNode(
     recordCanvas,
     keepIframeSrcFn,
     newlyAddedElement = false,
+    blankTextNodes = false,
```
eoghanmurray, line 468: This addition is to prevent extra 'reading' of the textContent, while still generating an `id` for the text node.  
However, if there's multiple text node children on the <style> we still will have to 'read' the textContent in `findCssTextSplits`

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -219,14 +219,47 @@ function buildNode(
         if (name.startsWith('rr_')) {
           specialAttributes[name] = value;
           continue;
+        } else if (name === '_cssTextSplits') {
+          continue;
```
eoghanmurray, line 223: I could have called this `rr_cssTextSplits`  ?

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -367,11 +435,11 @@ function buildNode(
...
-      return doc.createTextNode(
-        n.isStyle && hackCss
-          ? adaptCssForReplay(n.textContent, cache)
-          : n.textContent,
-      );
+      if (n.isStyle && hackCss) {
+        // support legacy style
+        return doc.createTextNode(adaptCssForReplay(n.textContent, cache));
+      }
+      return doc.createTextNode(n.textContent);
```
eoghanmurray, line 442: This diff is intended to be functionally the same, but rewritten to show that the `isStyle` marker is no longer expected

packages/rrweb/test/__snapshots__/record.test.ts.snap
```diff
@@ -1386,18 +1386,19 @@ exports[`record captures inserted style text nodes correctly 1`] = `
                   {
                     \\"type\\": 2,
                     \\"tagName\\": \\"style\\",
-                    \\"attributes\\": {},
+                    \\"attributes\\": {
+                      \\"_cssText\\": \\"div { color: red; }section { color: blue; }\\",
+                      \\"_cssTextSplits\\": \\"19 43\\"
+                    },
                     \\"childNodes\\": [
```
eoghanmurray, line 1393: So @Juice10 I think you had a question here about why I am moving things up to `_cssText` and adding `_cssTextSplits` rather than just storing everything in the childNodes ...
I suppose the "real reason" is that in the asset version of this code, we don't have the `_cssText` or `_cssTextSplits` attributes at all, but instead get these later as part of the asset event.  So with the asset version, we are avoiding the inclusion of the css text content twice (in terms of size of total recording).

packages/rrweb-snapshot/src/types.ts
```diff
@@ -20,9 +20,19 @@ export type documentTypeNode = {
   systemId: string;
 };
 
-export type attributes = {
-  [key: string]: string | number | true | null;
+type cssTextKeyAttr = {
+  _cssText?: string;
+  _cssTextSplits?: string;
```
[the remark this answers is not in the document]
Juice10: There _might_ be slightly better way of doing this in typescript to allow the special properties to include other types of values.
Would be great if _cssTextSplits could be an array, or even better if _cssText could be an array. But I'd have to spend some time digging around in typescript to see if it's possible. (It might not be, I'm not 100% sure here)

Juice10 (commented):
Thanks for tackling this Eoghan!

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -311,11 +398,11 @@ function buildNode(
...
```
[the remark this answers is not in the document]
eoghanmurray: No, after this PR, the css text content is never recorded as a text node in the full snapshot

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -367,11 +375,11 @@ function buildNode(
...
     }
     case NodeType.Text:
-      return doc.createTextNode(
-        n.isStyle && hackCss
-          ? adaptCssForReplay(n.textContent, cache)
-          : n.textContent,
-      );
+      if (n.isStyle && hackCss) {
+        // support legacy style
+        return doc.createTextNode(adaptCssForReplay(n.textContent, cache));
```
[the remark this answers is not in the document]
eoghanmurray: Check out the new `style.html` which has style elements with hover pseudoclasses; they are tested against with `can record and replay style mutations`.

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -446,6 +446,7 @@ function serializeNode(
      * `newlyAddedElement: true` skips scrollTop and scrollLeft check
      */
     newlyAddedElement?: boolean;
+    blankTextNodes?: boolean;
```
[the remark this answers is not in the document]
eoghanmurray: I've renamed the variable to `cssCaptured`. Is that self evident enough?

packages/rrweb-snapshot/src/rebuild.ts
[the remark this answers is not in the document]
eoghanmurray: Actually your comment tipped me off that there is one place where we still process raw css outside of the `_cssText` attribute; direct text mutations.  I've added a test and fix for this case in https://github.com/rrweb-io/rrweb/pull/1437/commits/2a41d0d50a4816891b165484183ccbc1282f7600 as we were missing `adaptCssForReplay` there.  The whole 'can record and replay style mutations' is was initially aimed to "shows that the `isStyle` attribute on textContent is not needed in a mutation" (but that test does a whole lot more now)

packages/rrweb/test/__snapshots__/record.test.ts.snap
eoghanmurray: (_cssTextSplits are gone in this PR now as a separate attribute in favour of a special `/* rr_split */` marker directly in the `_cssText`)

packages/rrweb-snapshot/src/types.ts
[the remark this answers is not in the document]
eoghanmurray: So I've since removed `_cssTextSplits` and kept `_cssText` as a string, but injected the special marker `/* rr_split */` into the css content where needed.  I think this is the best of all worlds in terms of not adding a new attribute and not modifying the existing type (or replacing it with a new `rr_cssTexts` array attribute as was the other proposal).
The `_cssText` will be demoted in #1475 in favour of an `rr_css_text` which records the id of the style node as the css text will be coming in asynchronously.

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -351,3 +360,53 @@ export function extractFileExtension(
   const match = url.pathname.match(regex);
   return match?.[1] ?? null;
 }
+
+function normalizeCssString(cssText: string): string {
+  // remove spaces
+  // TODO: normalize other differences between css as authored vs. stringifyStylesheet
+  return cssText.replace(/[\s]/g, '');
```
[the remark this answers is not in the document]
eoghanmurray: The normalization is only for making a comparison between the raw text and the serialized css, the normalized css is discarded after we figure out where the splits should go.

packages/rrweb-snapshot/src/rebuild.ts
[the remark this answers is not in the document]
eoghanmurray: (I believe this was a bug even before this PR but haven't looked into that)

packages/rrweb-snapshot/src/types.ts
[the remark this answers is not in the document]
Juice10: I like this change @eoghanmurray I think it's quite elegant, well done!

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -422,3 +436,59 @@ export function absolutifyURLs(cssText: string | null, href: string): string {
...
+    for (let i = 1; i < childNodes.length; i++) {
+      if (
+        childNodes[i].textContent &&
+        typeof childNodes[i].textContent === 'string'
+      ) {
+        const textContentNorm = normalizeCssString(childNodes[i].textContent!);
+        for (let j = 3; j < textContentNorm.length; j++) {
+          // find a  substring that appears only once
+          const bit = textContentNorm.substring(0, j);
+          if (cssTextNorm.split(bit).length === 2) {
```
Juice10, line 468: Wouldn't this happen pretty quick? Example case where one text node is `a { a:1px }` and another is `b { b:1px }`:
```
> bit = "a { a:1px }; b { b:1px }".substring(0,3)
< "a {"
> "a { a:1px }; b { b:1px }".split(bit)
< ["", " a:1px }; b { b:1px }"] (2)
```

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -78,6 +80,79 @@ export function createCache(): BuildCache {
...
+      cssText = cssText.replace(/\/\* rr_split \*\//g, '');
+    }
+    if (hackCss) {
+      cssText = adaptCssForReplay(cssText, cache);
+    }
+    // id will be assigned when these child nodes are
+    // iterated over in buildNodeWithSN
+    scn.textContent = cssText;
+    cssText = remainder;
+  }
```
Juice10, line 119: This seems like it would be a lot simpler by splitting the cssText on `/* rr_split */`:
```suggestion
  const cssTextSplits = cssText.split("/* rr_split */")
  for (let i = 0; i < childTextNodes.length; i++) {
    const childTextNode = childTextNodes[i];
    const cssTextSection = cssTextSplits[i];
    if (childTextNode && cssTextSection) {
      // id will be assigned when these child nodes are
      // iterated over in buildNodeWithSN
      childTextNode.textContent = hackCSS ? adaptCssForReplay(cssTextSection, cache) : cssTextSection;
    }
  }
```

packages/rrweb-snapshot/test/css.test.ts
```diff
@@ -1,6 +1,18 @@
...
+import postcss, { type AcceptedPlugin } from 'postcss';
+import { JSDOM } from 'jsdom';
+import { splitCssText, stringifyStylesheet } from './../src/utils';
+import { applyCssSplits } from './../src/rebuild';
+import {
+  NodeType,
+  serializedElementNodeWithId,
+  BuildCache,
+  textNode,
+} from '../src/types';
```
Juice10, line 15:
```suggestion
import {
  NodeType,
  type serializedElementNodeWithId,
  type BuildCache,
  type textNode,
} from '../src/types';
```

packages/rrweb-snapshot/test/css.test.ts
```diff
@@ -73,3 +85,135 @@ li[attr="has,comma"] a.\\:hover {background: red;}`,
...
+      );
+      // browser .rules would usually omit the vendored versions and modifies the transition value
+      const expected = [
+        '.x { content: "try to keep a newline"; background: red; transition: 4s; }',
+        '.y { transition: 5s; }',
+      ];
+      const browserSheet = expected.join('');
+
+      // can't do this as JSDOM doesn't have style.sheet
+      //expect(stringifyStylesheet(style.sheet!)).toEqual(browserSheet);
```
Juice10, line 136: Because we are running this with [vitest](https://vitest.dev/config/#environment) you can also use [happy-dom](https://github.com/capricorn86/happy-dom) instead of jsdom, as it does have CSSStyleSheet support.

packages/rrweb-snapshot/test/css.test.ts
```diff
@@ -73,3 +85,135 @@ li[attr="has,comma"] a.\\:hover {background: red;}`,
...
+    if (style) {
+      // as authored, with newlines
+      style.appendChild(
+        JSDOM.fragment(`.x {
+  -webkit-transition: all 4s ease;
+  content: 'try to keep a newline';
+  transition: all 4s ease;
+}`),
+      );
+      // TODO: splitCssText can't handle it yet if both start with .x
```
Juice10, line 121: Is this maybe due to the question I asked earlier about `splitCssText`?

packages/rrweb/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "prepare": "npm run prepack",
     "prepack": "npm run build",
-    "retest": "vitest run --exclude test/benchmark",
+    "retest": "PUPPETEER_HEADLESS=true yarn retest:headful",
```
Juice10, line 8: Needed for Windows
```suggestion
    "retest": "cross-env PUPPETEER_HEADLESS=true yarn retest:headful",
```

Juice10 (approved):
Looks good @eoghanmurray, most of my code suggestions are cosmetic/minor, feel free to merge the PR when you think it's ready.

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -422,3 +436,59 @@ export function absolutifyURLs(cssText: string | null, href: string): string {
...
+ */
+export function splitCssText(
+  cssText: string,
+  style: HTMLStyleElement,
+): string[] {
+  const childNodes = Array.from(style.childNodes);
+  const splits: string[] = [];
+  if (childNodes.length > 1 && cssText && typeof cssText === 'string') {
+    const cssTextNorm = normalizeCssString(cssText);
+    for (let i = 1; i < childNodes.length; i++) {
```
Juice10, line 459:
```suggestion
    // starts looking for the second css text node, then third and so forth...
    for (let i = 1; i < childNodes.length; i++) {
```

packages/rrweb-snapshot/src/rebuild.ts
YunFeng0817: The refined code is easier to understand to me

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -422,3 +436,59 @@ export function absolutifyURLs(cssText: string | null, href: string): string {
     },
   );
 }
+
+function normalizeCssString(cssText: string): string {
+  // remove spaces
+  // TODO: normalize other differences between css as authored vs. stringifyStylesheet
+  return cssText.replace(/[\s]/g, '');
```
Juice10, line 443: > // TODO: normalize other differences between css as authored vs. stringifyStylesheet
I think css comments might also be stripped out in `stringifyStylesheet`

packages/rrweb-snapshot/src/rebuild.ts
Juice10: @eoghanmurray it would be great if we could add this code before merging

otan:
i don't have an exact page to reproduce this problem, but i find that `/* rrweb-split */` is introduced in places which creates invalid CSS, e.g. this beast:

```
"_cssText": "@keyframes intercom-lightweight-app-launcher { \n  0% { opacity: 0; transform: scale(0.5); }\n  100% { opacity: 1; transform: scale(1); }\n}@keyframes intercom-lightweight-app-gradient { \n  0% { opacity: 0; }\n  100% { opacity: 1; }\n}@keyframes intercom-lightweight-app-messenger { \n  0% { opacity: 0; transform: scale(0); }\n  40% { opacity: 1; }\n  100% { transform: scale(1); }\n}.intercom-lightweight-app { position: fixed/* rr_split */; width: 0px; height: 0px; font-family: intercom-font, \"Helvetica Neue\", \"Apple Color Emoji\", Helvetica, Arial, sans-serif; }.intercom-lightweight-app-gradient { position: fixed; width: 500px; height: 500px; bottom: 0px; right: 0px; pointer-events: none; background: radial-gradient(at right bottom, rgba(29, 39, 54, 0.16) 0%, rgba(29, 39, 54, 0) 72%); animation: 200ms ease-out 0s 1 normal none running intercom-lightweight-app-gradient; }.intercom-lightweight-app-launcher { position: fixed; border: none; bottom: 20px; right: 20px; max-width: 48px; width: 48px; max-height: 48/* rr_split */px; height: 48px; border-radius: 50%; background: rgb(22, 38, 148); cursor: pointer; box-shadow: rgba(0, 0, 0, 0.06) 0px 1px 6px 0px, rgba(0, 0, 0, 0.16) 0px 2px 32px 0px; transition: transform 167ms cubic-bezier(0.33, 0, 0, 1); box-sizing: content-box; padding: 0px !important; margin: 0px !important; }.intercom-lightweight-app-launcher:hover { transition: transform 250ms cubic-bezier(0.33, 0, 0, 1); transform: scale(1.1); }.intercom-lightweight-app-launcher:active { transform: scale(0.85); transition: transform 134ms cubic-bezier(0.45, 0, 0.2, 1); }.intercom-lightweight-app-launcher:focus { outline: none; }.intercom-lightweight-app-launcher-icon { display: flex; align-items: center; justify-content: center; position: absolute; top: 0px; left: 0px; width: 48px; height: 48px; transition: transform 100ms linear, opacity 80ms linear; }.intercom-lightweight-app-launcher-icon-open { opacity: 1; transform: rotate(0deg) scale(1); }.intercom-lightweight-app-launcher-icon-open svg { width: 24px; height: 24px; }.intercom-lightweight-app-launcher-icon-open svg path { fill: rgb(255, 255, 255); }.intercom-lightweight-app-launcher-icon-self-serve { opacity: 1; transform: rotate(0deg) scale/* rr_split */(1); }.intercom-lightweight-app-launcher-icon-self-serve svg { height: 44px; }.intercom-lightweight-app-launcher-icon-self-serve svg path { fill: rgb(255, 255, 255); }.intercom-lightweight-app-launcher-custom-icon-open { max-height: 24px; max-width: 24px; opacity: 1; transform: rotate(0deg) scale(1); }.intercom-lightweight-app-launcher-icon-minimize { opacity: 0; transform: rotate(-60deg) scale(0); }.intercom-lightweight-app-launcher-icon-minimize svg path { fill: rgb(255, 255, 255); }.intercom-lightweight-app-messenger { position: fixed; overflow: hidden; background-color: white; animation: 250ms cubic-bezier(0, 1, 1, 1) 0s 1 normal none running intercom-lightweight-app-messenger; transform-origin: right bottom; width: 400px; height: calc(100% - 40px); max-height: 704px; min-height: 250px; right: 20px; bottom: 20px; box-shadow: rgba(0, 0, 0, 0.16) 0px 5px 40px; border-radius: 16px; }.intercom-lightweight-app-messenger-header { height: 64px; border-bottom: none; background: rgb(255, 255, 255); }.intercom-lightweight-app-messenger-footer { position: absolute; bottom: 0px; width: 100%; height: 80px; background: rgb(255, 255, 255); font-size: 14px; line-height: 21px; border-top: 1px solid rgba(0, 0, 0, 0.05); box-shadow: rgba(0, 0, 0, 0.05) 0px 0px 25px; }@media print {\n  .intercom-lightweight-app { display: none; }\n}"
```

now fails to replay despite having valid styles.

otan:
here's a repro:

```
const s = document.createElement("style"); s.append(document.createTextNode(".a {")); s.append(document.createTextNode("color: red; }"));
document.head.appendChild(s);
```

PR #1439, opened 2024-04-07
fix: solve the smooth scrolling scene, the y value obtained is the data 100ms ago

xujiujiu:
In the smooth scrolling scene on the mobile phone, the y value obtained by the monitored scroll event is too small, causing the bottom content to be invisible during playback.


xujiujiu:
When listening to the scroll event, the x and y data of the node cannot be obtained because the target cannot be obtained through `throttle`

xujiujiu:
image

xujiujiu:
image
there target is undefined

packages/rrweb/src/utils.ts
```diff
@@ -88,10 +88,19 @@ export function throttle<T>(
       previous = now;
       func.apply(context, args);
     } else if (!timeout && options.trailing !== false) {
+      const argumentsList = args.map((arg: T)=> {
+        return {
+          ...arg,
+          target: arg.target
+        }
+      })
```
eoghanmurray, line 96: Is this argument transformation something specific to MouseEvent/TouchEvent/DragEvent ?

I think this needs a bit more explanation; the `T` type here is supposed to be generic, right?

packages/rrweb/src/utils.ts
```diff
@@ -88,10 +88,19 @@ export function throttle<T>(
...
       timeout = setTimeout(() => {
         previous = options.leading === false ? 0 : Date.now();
-        timeout = null;
-        func.apply(context, args);
+        if (timeout) {
+          clearTimeout(timeout)
```
eoghanmurray, line 100: I can't see why you need to clear the timeout here as it is a 'one-shot' timeout.
clearing it would make sense if it was `setInterval`, but it seems to me that it's already been called and won't be called again?
Is there some concurrency or similar here that is possible?

packages/rrweb/src/utils.ts
eoghanmurray: I'm still wondering why we need to do the transformation?
is it that an explicit reference to the `target` needs to be saved before the timeout?
I would have thought that `args` being in a closure should handle saving the reference?

PR #1440, opened 2024-04-09
better splitting of selectors

daibhin:
As part of the full investigation in https://github.com/PostHog/posthog/pull/21427 I noticed that nested commas in selectors are not very well supported by the current `.split` regex.

Moved to a JS function that only splits root selectors with a comma.

It looks to me like the `css.parse` is only used to generate this AST in one place. Perhaps it's better we move this whole function over to use the already included `cssom` equivalent method. Unsure if that suffers the same issues, so proposing this given I know it works but happy to explore the alternative root if it is preferred. 

@eoghanmurray might be best placed to help here given recent work on hover selectors


eoghanmurray:
This is great, I suggest you create a bug report or PR on the codebase that css.ts was copied from; https://github.com/reworkcss/css/blob/master/lib/parse/index.js

eoghanmurray:
Unfortunately I was able to produce a further failing test by being pathological:

```li[attr="weirdly("] a:hover, li[attr="weirdly)"] a {}```

https://github.com/rrweb-io/rrweb/commit/84f0f224c0e78c87d144d23933d003814fd3d181

This PR is better than what we had before, but do you want to improve further to fix my test case?

File: packages/rrweb-snapshot/src/css.ts [hunk 1]
```diff
@@ -429,17 +429,51 @@ export function parse(css: string, options: ParserOptions = {}): Stylesheet {
...
+        currentPart += char;
+
+        if (char === '(' && !currentStringChar) {
+          nestedLevel++;
+        } else if (char === ')' && !currentStringChar) {
+          nestedLevel--;
+        } else if (char === ',' && nestedLevel === 0) {
+          parts.push(currentPart.slice(0, -1).trim());
+          currentPart = '';
+        } else if (char.match(/"|'|`/)) {
```
eoghanmurray, line 450: the backtick here is not indicated as possible in the CSS spec:
https://www.w3.org/TR/selectors-3/#attribute-selectors

"Attribute values must be CSS [identifiers](https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier) or [strings](https://www.w3.org/TR/CSS21/syndata.html#strings)."

"Strings can either be written with double quotes or with single quotes"

eoghanmurray, line 450:
```suggestion
        } else if (char.match(/"|'/)) {
```

eoghanmurray:
Tests needed a re-run after your latest commit this morning, but I managed to add another test case that breaks things;

it's possible to escape double quotes within those css strings :(

packages/rrweb-snapshot/src/css.ts
eoghanmurray: Or 
`} else if ("'" === char || '"' === char) {` 
if you prefer

eoghanmurray (approved):

packages/rrweb-snapshot/src/css.ts
```diff
@@ -429,17 +429,51 @@ export function parse(css: string, options: ParserOptions = {}): Stylesheet {
...
+        if (char === '(' && !currentStringChar) {
+          nestedLevel++;
+        } else if (char === ')' && !currentStringChar) {
+          nestedLevel--;
+        } else if (char === ',' && nestedLevel === 0) {
+          parts.push(currentPart.slice(0, -1).trim());
+          currentPart = '';
+        } else if ('\'"'.includes(char)) {
+          if (currentStringChar === char) {
+            currentStringChar = null;
```
eoghanmurray, line 452:
```suggestion
            if (!(i > 0 && input[i - 1] === '\\')) {
              currentStringChar = null;
            }
```

daibhin: Stepping through a debugger here it looks like the `\"` is just rendered as a `"` which means the selector is not malformed and hence the tests fail in https://github.com/rrweb-io/rrweb/actions/runs/8661394617/job/23751204467?pr=1440
Screenshot 2024-04-12 at 17 12 29

Going to go ahead and remove this test case given it's a very edge case of a very edge case and getting this through is already an improvement over what exists

packages/rrweb-snapshot/test/css.test.ts
```diff
@@ -119,6 +119,39 @@ describe('css parser', () => {
...
+    const weirderresult = parse(
+      `
+li[attr="weirder\"("] a:hover, li[attr="weirder\")"] a {
+  background-color: red;
+}
+`,
+    );
+    expect(
+      (weirderresult.stylesheet!.rules[0] as Rule)!.selectors!.length,
+    ).toEqual(2);
```
daibhin, line 152:
```suggestion
```

daibhin:
I'm not permitted to merge this PR so I'll leave it up to a library maintainer

eoghanmurray:
@daibhin I added a question above: "Was this removed accidentally?"

daibhin:
@eoghanmurray I couldn't find your latest question. Did you mean the final test case? I left a reply in https://github.com/rrweb-io/rrweb/pull/1440#discussion_r1562787952 that hopefully explains things

eoghanmurray:
I've just noticed #1401 which has been merged into trunk and seems to solve a similar issue; I'll try to run the new tests from this PR against master now...

eoghanmurray:
`trickresult` test still fails on master ... so we still need this PR!
The merge is not straightforward though...

eoghanmurray:
I've added back the `weirderresult` test again btw and tweaked it so that escaping was effective to demonstrate the problem

.changeset/modern-doors-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'rrweb-snapshot': patch
+---
+
+better nested css selector splitting
```
eoghanmurray, line 5:
```suggestion
better nested css selector splitting when commas or brackets happen to be in quoted text
```

PR #1441, opened 2024-04-09
perf: Avoid an extra function call and object clone during event emission

eoghanmurray:
See #1339 - perf: remove a nested function call and an object clone per event


Juice10 (approved):

eoghanmurray:
427acff3d6aeb90b77f81a495fc20ef6ad996b2b removed the timestamp from the 'event' type and added the 'eventWithTime' .
with this change, 'event' is now called 'eventWithoutTime'

in future I'm also thinking of renaming `eventWithTime` to just `event` (or `rrEvent`)

Issue #1450, opened 2024-04-17
[Bug]: echart color error


Belen-Luo:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

v 2.0.0-alpha.12

### Expected Behavior

Expected normal rendering echarts

### Actual Behavior

During the recording of echarts, if the echarts are operated, the echarts color will turn gray

### Steps to Reproduce

1. renderer echarts
2. Start recording
3. Mouse in echarts
4. End View the recording result

### Additional Information

<div id="echarts" style="width: 200px; height: 200px;"></div>
const option = {
        color: ['#80FFA5', '#00DDFF', '#37A2FF', '#FF0087', '#FFBF00'],
        title: {
          text: 'Gradient Stacked Area Chart'
        },
        tooltip: {
          trigger: 'axis',
          axisPointer: {
            type: 'cross',
            label: {
              backgroundColor: '#6a7985'
            }
          }
        },
        legend: {
          data: ['Line 1', 'Line 2', 'Line 3', 'Line 4', 'Line 5']
        },
        toolbox: {
          feature: {
            saveAsImage: {}
          }
        },
        grid: {
          left: '3%',
          right: '4%',
          bottom: '3%',
          containLabel: true
        },
        xAxis: [
          {
            type: 'category',
            boundaryGap: false,
            data: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun']
          }
        ],
        yAxis: [
          {
            type: 'value'
          }
        ],
        series: [
          {
            name: 'Line 1',
            type: 'line',
            stack: 'Total',
            smooth: true,
            lineStyle: {
              width: 0
            },
            showSymbol: false,
            areaStyle: {
              opacity: 0.8,
              color: new echarts.graphic.LinearGradient(0, 0, 0, 1, [
                {
                  offset: 0,
                  color: 'rgb(128, 255, 165)'
                },
                {
                  offset: 1,
                  color: 'rgb(1, 191, 236)'
                }
              ])
            },
            emphasis: {
              focus: 'series'
            },
            data: [140, 232, 101, 264, 90, 340, 250]
          },
          {
            name: 'Line 2',
            type: 'line',
            stack: 'Total',
            smooth: true,
            lineStyle: {
              width: 0
            },
            showSymbol: false,
[...]

PR #1451, opened 2024-04-17
fix: Fixed the color of echarts turning gray when clicked or moved in

Belen-Luo:
[Bug]: echart color error [1450](https://github.com/rrweb-io/rrweb/issues/1450)

### DIFF
--- a/packages/rrweb/src/record/observer.ts
+++ b/packages/rrweb/src/record/observer.ts
@@ -1314,7 +1314,7 @@ export function initObservers(
       //
     };
   }
-  console.warn(o)
+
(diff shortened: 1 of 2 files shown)


eoghanmurray:
This seems too general a change; see `should execute all mutations after args are parsed` failing test.  This PR would need a good justification/explanation for why we throw away commands (at least that's what I assume is happening from reading the diff)

is there something specific about the canvas command in echarts that is causing failure?  Is there a specific exception?

Belen-Luo:
Okay, I'll look into it

PR #1458, opened 2024-04-24
feat: add new css parser - postcss

daibhin:
`rrweb-snapshot` parses stylesheets to adapt the CSS for playback

In certain cases the parsing wasn't always correct:
https://github.com/PostHog/posthog/pull/21427
https://github.com/rrweb-io/rrweb/issues/1379

Which led to improved CSS selector efforts first included in `rrweb@2.0.0-alpha.13` (https://github.com/rrweb-io/rrweb/pull/1401)

Sadly this caused playback issues that consumed 100% CPU and froze playback
https://github.com/PostHog/posthog/issues/21791
profiler

----

CSS parsing is performed in the https://github.com/rrweb-io/rrweb/blob/master/packages/rrweb-snapshot/src/css.ts which was pulled from https://github.com/reworkcss/css (which hasn't been updated in over 3 years)

I propose moving to https://github.com/csstree/csstree which creates an actual AST which can be used to effectively debug via https://astexplorer.net

I couldn't figure out some import issues in the `test/integration.test.ts`
```
Evaluation failed: ReferenceError: require is not defined
```

All `test/css.test.ts` tests still pass which confirms this is as good as the trickiest cases previously handled without any of the slowness introduced in https://github.com/rrweb-io/rrweb/pull/1401


packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -70,34 +78,18 @@ export function adaptCssForReplay(cssText: string, cache: BuildCache): string {
...
-    if ('media' in rule && rule.media && MEDIA_SELECTOR.test(rule.media)) {
-      medias.push(rule.media);
-    }
-    if ('rules' in rule && rule.rules) {
-      rule.rules.forEach(getSelectors);
-    }
-  }
-  getSelectors(ast.stylesheet);
+  const mediaFeatures = css
+                        .findAll(newAst, (node) => node.type === 'MediaFeature')
```
daibhin, line 88: Possibly some further optimizations we could even make here by using the `css.walk` method to only traverse the AST once (rather than twice like I assume `findAll` is doing)

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -70,34 +78,18 @@ export function adaptCssForReplay(cssText: string, cache: BuildCache): string {
...
   if (cachedStyle) return cachedStyle;
 
-  const ast = parse(cssText, {
-    silent: true,
-  });
-
-  if (!ast.stylesheet) {
-    return cssText;
-  }
+  const newAst = css.parse(cssText);
```
pauldambra, line 81: i'm a little allergic to `new` in variable names... does it have meaning here? or could we `cssAST` instead?

we're calling `findAll` and `map` twice as well, since we know we had perf issues here already it might be worth trying to loop once... (although very much wondering towards premature optimization so treat as nitpicking)

daibhin: Ah good spot. That was actually a hangover from when I had the old code running alongside to test things. Will revert to the pre-existing names

packages/rrweb-snapshot/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "rrweb-snapshot",
-  "version": "2.0.0-alpha.13",
+  "version": "2.0.0-alpha.12",
```
marandaneto, line 3: Should probably be `13`?

eoghanmurray:
This is great and something I thought we might get to 'some day' after the recent fixes to css.ts.

Would you be able to post a before/after of the built file sizes from this change; i.e.

```$ ls -al packages/rrweb/dist/*.js packages/rrweb/dist/record/*.js packages/rrweb/dist/replay/*.js```

One of the reasons css.ts was ported over was so that we could tree shake it a bit.

**Comment by daibhin** [comment]
Something isn't right here because I'm getting the same values when running `yarn build:dev && ls -al packages/rrweb/dist/*.js packages/rrweb/dist/record/*.js packages/rrweb/dist/replay/*.js` on both branches.

Possibly need to fix the import errors I mentioned in the description first 🤔 🤷 

```
-rw-r--r--@ 7 david  staff   26568 29 Apr 18:24 packages/rrweb/dist/record/rrweb-record-pack.js
-rw-r--r--@ 7 david  staff   10418 29 Apr 18:24 packages/rrweb/dist/record/rrweb-record-pack.min.js
-rw-r--r--@ 7 david  staff  165350 29 Apr 18:24 packages/rrweb/dist/record/rrweb-record.js
-rw-r--r--@ 7 david  staff   68811 29 Apr 18:24 packages/rrweb/dist/record/rrweb-record.min.js
-rw-r--r--@ 7 david  staff  209951 29 Apr 18:24 packages/rrweb/dist/replay/rrweb-replay-unpack.js
-rw-r--r--@ 7 david  staff   92849 29 Apr 18:24 packages/rrweb/dist/replay/rrweb-replay-unpack.min.js
-rw-r--r--@ 7 david  staff  193751 29 Apr 18:24 packages/rrweb/dist/replay/rrweb-replay.js
-rw-r--r--@ 7 david  staff   88049 29 Apr 18:24 packages/rrweb/dist/replay/rrweb-replay.min.js
-rw-r--r--@ 7 david  staff  401587 29 Apr 18:24 packages/rrweb/dist/rrweb-all.js
-rw-r--r--@ 7 david  staff  170294 29 Apr 18:24 packages/rrweb/dist/rrweb-all.min.js
-rw-r--r--@ 7 david  staff  351097 29 Apr 18:24 packages/rrweb/dist/rrweb.js
-rw-r--r--@ 7 david  staff  152936 29 Apr 18:24 packages/rrweb/dist/rrweb.min.js
```

**Comment by daibhin** [comment]
In a significant update here I moved to using `postcss` once I figured out how their plugin system works. I've moved around some test files to better separate concerns.

Some of the tests were "testing the framework" given they had been copied over from https://github.com/reworkcss/css/blob/master/test/parse.js. I removed any that were no longer relevant but left any of the cases added in https://github.com/rrweb-io/rrweb/pull/1440/files#diff-ca07555d1c1965f9c7ef3b5df3772a19953138b8eac62deb7786e549b25b455d as we know they've specifically broken in the past and we would like to avoid breaking them in the future

**Comment by eoghanmurray** [comment]
Super! What sort of filesize numbers are there with the `postcss` version?

**Comment by daibhin** [comment]
Sorry, probably wasn't clear but I updated my earlier comment with latest build numbers having run `yarn build:dev && ls -al packages/rrweb/dist/*.js packages/rrweb/dist/record/*.js packages/rrweb/dist/replay/*.js`

`master`
```
-rw-r--r--@ 7 david  staff   26568 29 Apr 18:24 packages/rrweb/dist/record/rrweb-record-pack.js
-rw-r--r--@ 7 david  staff   10418 29 Apr 18:24 packages/rrweb/dist/record/rrweb-record-pack.min.js
-rw-r--r--@ 7 david  staff  165350 29 Apr 18:24 packages/rrweb/dist/record/rrweb-record.js
-rw-r--r--@ 7 david  staff   68811 29 Apr 18:24 packages/rrweb/dist/record/rrweb-record.min.js
-rw-r--r--@ 7 david  staff  209951 29 Apr 18:24 packages/rrweb/dist/replay/rrweb-replay-unpack.js
-rw-r--r--@ 7 david  staff   92849 29 Apr 18:24 packages/rrweb/dist/replay/rrweb-replay-unpack.min.js
-rw-r--r--@ 7 david  staff  193751 29 Apr 18:24 packages/rrweb/dist/replay/rrweb-replay.js
-rw-r--r--@ 7 david  staff   88049 29 Apr 18:24 packages/rrweb/dist/replay/rrweb-replay.min.js
-rw-r--r--@ 7 david  staff  401587 29 Apr 18:24 packages/rrweb/dist/rrweb-all.js
-rw-r--r--@ 7 david  staff  170294 29 Apr 18:24 packages/rrweb/dist/rrweb-all.min.js
-rw-r--r--@ 7 david  staff  351097 29 Apr 18:24 packages/rrweb/dist/rrweb.js
-rw-r--r--@ 7 david  staff  152936 29 Apr 18:24 packages/rrweb/dist/rrweb.min.js
```

`fix/css-selectors`
```
-rwxrwxr-x@ 3 david  staff   26584 30 Apr 16:33 packages/rrweb/dist/record/rrweb-record-pack.js
-rwxrwxr-x@ 3 david  staff   10437 30 Apr 16:33 packages/rrweb/dist/record/rrweb-record-pack.min.js
-rwxrwxr-x@ 3 david  staff  157698 30 Apr 16:33 packages/rrweb/dist/record/rrweb-record.js
-rwxrwxr-x@ 3 david  staff   68728 30 Apr 16:33 packages/rrweb/dist/record/rrweb-record.min.js
-rwxrwxr-x@ 3 david  staff  191575 30 Apr 16:33 packages/rrweb/dist/replay/rrweb-replay-unpack.js
-rwxrwxr-x@ 3 david  staff   87904 30 Apr 16:33 packages/rrweb/dist/replay/rrweb-replay-unpack.min.js
-rwxrwxr-x@ 3 david  staff  175375 30 Apr 16:33 packages/rrweb/dist/replay/rrweb-replay.js
-rwxrwxr-x@ 3 david  staff   83086 30 Apr 16:33 packages/rrweb/dist/replay/rrweb-replay.min.js
-rwxrwxr-x@ 3 david  staff  375719 30 Apr 16:33 packages/rrweb/dist/rrweb-all.js
-rwxrwxr-x@ 3 david  staff  165248 30 Apr 16:33 packages/rrweb/dist/rrweb-all.min.js
-rwxrwxr-x@ 3 david  staff  325229 30 Apr 16:34 packages/rrweb/dist/rrweb.js
-rwxrwxr-x@ 3 david  staff  147850 30 Apr 16:34 packages/rrweb/dist/rrweb.min.js
```

File: packages/rrweb-snapshot/src/rebuild.ts [hunk 4]
```diff
@@ -9,6 +9,9 @@ import {
 } from './types';
 import { isElement, Mirror, isNodeMetaEqual } from './utils';
 
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-var-requires
+const postcss = require('postcss');
```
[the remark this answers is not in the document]
eoghanmurray: This is causing tests to fail with `require is not defined`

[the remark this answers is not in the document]
eoghanmurray: Is it that we need a typescript version of https://github.com/giuseppeg/postcss-pseudo-classes
 ?

eoghanmurray:
Hopefully the following doesn't derail the conversation, and even if we go ahead with the following, that doesn't detract from the worth of the current PR...

So okay I just had a thought; all this CSS processing is on the replay side, right?
I'm wondering why we don't just iterate over the final sheet.rules directly after the stylesheet has been rendered.  We will be rendering it as an inline <style> so the .sheet.rules will be accessible.  I don't think we'd need any 3rd party library for that approach, and it would have another advantage in that we could apply the hover classes to stylesheets that are only referenced in the recording (assuming the .sheet.rules are accessible).  I'm planning to do this approach anyway so that we can apply our hover classes to CDN stylesheets that are currently recorded only by reference — if you replay the <link> with `crossorigin="anonymous"` present, then it's possible to access and modify the rules for the sheet.  If that succeeds, it could enable further paring back on what needs to be inlined at record time; we could avoid inlining anything that has a 'versioned' url and a permanent web presence.

pauldambra:
> a 'versioned' url and a permanent web presence.

I don't think we can know that something has a permanent web presence only because it has a versioned URL - e.g. I was just dealing with a customer bug report that centred around the assets only living for 15 days and not inlining at capture

We either need very robust inlining or asset capture at recording time (ideally both :)). Loading over the net back to the customer is only a fallback in a perfect world.

(I appreciate that's a (this) consumer of the library position and not necessarily a developer of the library position 😊)

eoghanmurray:
Yes, this could be configurable, and likely a robust solution will involve deploy of your own proxy/cache for assets.  We replicate the '\.hover' modification server side in our proxy cache so applying those modifications at replay on all `.sheet` (rather than only the captured cssText) would also make a web proxy easier to run.

packages/rrweb-snapshot/src/rebuild.ts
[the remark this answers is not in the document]
Juice10: This require business might be fixed with this PR: https://github.com/rrweb-io/rrweb/pull/1033

eoghanmurray:
There is a new test in #1481 which adds coverage to the css parser; hopefully it will pass (or fail) as-is without having to modify the test itself.

eoghanmurray:
Also re. rrweb-snapshot/test/rebuild.test.ts

if you revert the whitespace changes to the expected css, you could instead cherry-pick the following commit which should do loose matching by ignoring whitespace and semicolon:
https://github.com/rrweb-io/rrweb/pull/1480/commits/85db6c21583746637e7c54b9b5782fe0f88f4868

eoghanmurray:
Let me know if you wish me to execute the above two for you.
Still awaiting someone to figure out how to pull in a typescript version of https://github.com/giuseppeg/postcss-pseudo-classes

packages/rrweb-snapshot/package.json
```diff
@@ -58,6 +58,10 @@
     "rollup-plugin-typescript2": "^0.31.2",
     "ts-jest": "^27.0.5",
     "ts-node": "^7.0.1",
-    "tslib": "^1.9.3"
+    "tslib": "^1.9.3",
+    "@types/postcss-js": "^4.0.4"
```
billyvg, line 62: I don't think this is needed, it's a different package than `postcss`

marandaneto: good point, this comes from https://github.com/postcss/postcss-js but the goal is to move to https://github.com/csstree/csstree
if this can be removed altogether would be great.

eoghanmurray: if someone wants to author an alternative PR using csstree, then go for it!

eoghanmurray: alternatively, we could continue this one but omit [postcss-pseudo-classes](https://github.com/giuseppeg/postcss-pseudo-classes) altogether; it does more than just modify the `:hover` part so if it's possible to write the `:hover` -> `.:hover` modification in another way using just postcss, then that might be more straightforward.

eoghanmurray: So I think this PR was originally using csstree but the author moved to use postcss subsequently 
https://github.com/PostHog/posthog/issues/18520

Juice10: Just double checked, @billyvg is indeed correct. Postcss comes with its own types, no need to add these types

daibhin: Cleaned up the types in the latest commit

packages/rrweb-snapshot/src/rebuild.ts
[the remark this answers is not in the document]
Juice10: @daibhin, PR has been merged, although what you have done (replacing require by import) seems perfectly good

packages/rrweb-snapshot/src/css.ts
```diff
@@ -1,987 +1,90 @@
...
-    for (const char of input) {
-      const hasStringEscape = currentSegment.endsWith('\\');
-
-      if (currentStringChar) {
-        if (currentStringChar === char && !hasStringEscape) {
-          currentStringChar = null;
+  },
+};
+
+// Adapted from https://github.com/giuseppeg/postcss-pseudo-classes/blob/master/index.js
```
eoghanmurray, line 20: This is great, and I think I questioned somewhere whether we needed to import this.

I'm having some doubts now though, as just notice the following is possible with that plugin:


```
  restrictTo: ['hover'],
```
did you consider that? 
This approach of copying over a subset of the code has it's attractions too.

packages/rrweb-snapshot/src/css.ts
```diff
@@ -1,987 +1,90 @@
...
+          const selectorParts = selector.replace(/\n/g, ' ').split(' ');
+          const pseudoedSelectorParts: string[] = [];
 
-    const ret = pos({
-      type: 'declaration',
-      property: prop.replace(commentre, ''),
-      value: val ? trim(val[0]).replace(commentre, '') : '',
-    });
+          selectorParts.forEach(function (selectorPart) {
+            const pseudos = selectorPart.match(/::?([^:]+)/g);
```
eoghanmurray, line 41: could possibly be more specific and include the `hover` string here given we know that's what we will be matching against

eoghanmurray: as in that's the only one we need to support

packages/rrweb-snapshot/test/css.test.ts
```diff
@@ -1,255 +1,83 @@
...
+        expect(
+          parse(
+            mediaSelectorPlugin,
+            `@media only screen and (${first}-device-${second}: 1200px) { .a { width: 10px; }}`,
+          ),
+        ).toEqual(
+          `@media only screen and (${first}-${second}: 1200px) { .a { width: 10px; }}`,
+        );
+      });
+      expect(
```
eoghanmurray, line 34: this one seems to be covered by the above `forEach`?

eoghanmurray (approved):
Broadly approve but have a few questions before merging.

packages/rrweb-snapshot/src/css.ts
daibhin: I considered adding the package but thought it easier just to copy the code. When I inlined it there were a bunch of options we didn't really need e.g. `restrictTo` so just simplified the whole thing down. We can always go back and add the original in future, but I chose to go with the simpler (no import / no extra options) until we knew if we needed any more

packages/rrweb-snapshot/test/css.test.ts
daibhin: Ah yes, that's the one I used to write the above test. Forgot to remove it

packages/rrweb-snapshot/src/css.ts
daibhin: This actually finds all pseudo elements on a selector, not just the hover

daibhin:
@Juice10 thanks for fixing up the imports. Do you know if the failing test is related to my changes somehow?

Juice10:
@daibhin failing test was for https://github.com/rrweb-io/rrweb/pull/1458/commits/2872d0ee4fa0455b76836a6616317b561c330344 was due to an upgraded package, thats fixed now (by locking the package version). The failure for  https://github.com/rrweb-io/rrweb/pull/1458/commits/c7153b7620a630eaadeee0159d0cc7b1767516c2 seems to be due to jest being extended where we switched over to vitest. I just fixed that issue, but it looks like a chunk of the :hover tests are failing (you'll see that soon for  https://github.com/rrweb-io/rrweb/pull/1458/commits/9df9946d7befa28a9d2633aecbce5e6ca11be9cf )

Juice10 (approved):
This looks really good, and is so much easier to adapt!

packages/rrweb-snapshot/src/css.ts
eoghanmurray: Did that commit make it in to the merge?  I can't see anything by following the link above?  I'm still seeing `const pseudos = selectorPart.match(/::?([^:]+)/g);` in current version

eoghanmurray: So I wrote up a simplified version in #1535, which I'm just looking at as it has a failing test; however I'm not so sure that the test failure isn't more correct!

We added the test in question in #1440 and it looks like the following:

```
body > ul :is(li:not(:first-of-type) a:hover, li:not(:first-of-type).active a) {background: red;}
```

The focus then was on ensuring that the above wasn't parsed as two separate selectors because of the internal comma, however that hasn't broken, but I think we missed that there was also a `:hover` in there which should mean that the whole selector should be duplicated (which is how the test fails).

In the version from this PR, everything between the brackets of `:is` is bundled in with the `:is`, so the internal `:hover` is effectively ignored.  I haven't investigated why it was ignored previously.  I'll duplicate this comment over in #1535 if you'd prefer to continue conversation there.

PR #1459, opened 2024-04-24
repro: adding node to <style> clears injected styles

billyvg:
Here is a repro of a case where a mutation that adds a node to a `<style>` (at least I think this is what `parentId=-2` is) will clear the previously injected styles. This ends up clearing the `noscript` css so that the contents of the `noscript` elements will be displayed in the replay.


eoghanmurray:
```parentId === -2``` means the parent node is ignored in terms of slimDom settings.

There is `parentId: -1` in this test case which shouldn't happen I believe (if that is possible we should add a record-time test too)

I haven't tried to recreate yet..

billyvg:
Superceded by https://github.com/rrweb-io/rrweb/pull/1357

PR #1461, opened 2024-04-26
fix: safe pause

marandaneto:


marandaneto:
cc @Juice10

eoghanmurray:
If it's complaining about RRMediaElement, then the fix likely should be in rrdom?

I don't understand the abstract base class stuff too well, but maybe it just needs a concrete implementation of the `pause` function here:
https://github.com/rrweb-io/rrweb/blob/master/packages/rrdom/src/index.ts#L152
?

marandaneto:
> If it's complaining about RRMediaElement, then the fix likely should be in rrdom?
> 
> I don't understand the abstract base class stuff too well, but maybe it just needs a concrete implementation of the `pause` function here: [`master`/packages/rrdom/src/index.ts#L152](https://github.com/rrweb-io/rrweb/blob/master/packages/rrdom/src/index.ts?rgh-link-date=2024-04-26T13%3A15%3A49Z#L152) ?

https://github.com/rrweb-io/rrweb/blob/ae6908dcdcd7c732c1ce79eea19de5240bec1151/packages/rrdom/src/document.ts#L576-L578 should be the implementation since the class extends the base class with the method implementation, so no, I don't think so but not a JS/TS expert either.

Juice10:
Any chance we could figure out what the type of element is that's being sent through? I'd like us to have a deeper fix if possible.

Juice10:
@marandaneto I'd prefer this fix although if we don't know exactly what's triggering it, it might not be helpful for your use case. https://github.com/rrweb-io/rrweb/pull/1462 Any chance you could check?

marandaneto:
> @marandaneto I'd prefer this fix although if we don't know exactly what's triggering it, it might not be helpful for your use case. #1462 Any chance you could check?

I could not reproduce the issue so I cannot check it but your PR makes sense to me, I added a comment with a possible improvement though, I think this is worth the shot.

marandaneto:
closing in favor of https://github.com/rrweb-io/rrweb/pull/1462
cc @daibhin

PR #1464, opened 2024-05-01
avoid using 2nd argument of Array.from that prototype.js does not support

colingm:
We ran into an issue where rrweb running on a site that uses prototype.js breaks prototype overrides native Array.from to not support the 2nd argument (the mapFn). We struggled with if it was right to open this pr against upstream due to it only being necessary to support a mostly abandoned tool but thought we would open it anyways and get your opinions on it.


eoghanmurray:
Prior to me examining this further:
- See #1272 for an example of where second argument of Array.from was introduced for performance reasons.
- See #1196 for a similar guard against an old 3rd party js library which in that case defined it's own `Date.now` (before there was a standard Date.now)

eoghanmurray:
Having had a quick look, we'll need something better than reverting #1272 which is kind of what this is ... 

 - Could make our own definition of array_from which checks whether it's safe to use the inbuilt one first 
 - Would also need to check whether our own one doesn't slow things down
 - There's some trick I remember where you create a new 'clean' document context, pull out the native Array.from function from that context, and re-overwrite the one in the top level context (which has been doctored by prototypejs)

Notwithstanding #1196, I think that's more suitable for code outside of rrweb, as prototype.js is old and doesn't do this anymore right?

colingm:
Also "there isn't really a good reason or a sustainable way to do this in rrweb" is a totally reasonable final decision. This was a combination of me going through our current changes that have diverted from upstream and making sure I at least saw if there was a good way to do it.

colingm:
@eoghanmurray good idea. I have done that with other tools before just for some reason didn't think about trying that here. Probably was trying to be as surgical as possible which might have made sense for a quick hotfix on our part but definitely doesn't make as much sense for the long term evolution of rrweb.

.changeset/eleven-bobcats-peel.md
```diff
@@ -0,0 +1,6 @@
+---
+'rrweb-snapshot': patch
+'rrweb': patch
+---
+
+try to provide clean implementation for Array.from to support older libraries that might drop the 2nd parameter
```
eoghanmurray, line 6:
```suggestion
better support for coexistence with older libraries (e.g. MooTools & Prototype.js) which modify the in-built `Array.from` function
```

eoghanmurray:
Thanks for working out the typing issues!  I'll leave it to another to review as I won't have seen problems in the code I suggested.

packages/rrweb/src/record/index.ts
```diff
@@ -46,6 +46,20 @@ let takeFullSnapshot!: (isCheckout?: boolean) => void;
...
 
+// Multiple tools (i.e. MooTools, Prototype.js) override Array.from and drop support for the 2nd parameter
+// Try to pull a clean implementation from a newly created iframe
+try {
+  if (Array.from([1], (x) => x * 2)[0] !== 2) {
+    const cleanFrame = document.createElement('iframe');
+    document.body.appendChild(cleanFrame);
+    // eslint-disable-next-line @typescript-eslint/unbound-method -- Array.from is static and doesn't rely on binding
+    Array.from = cleanFrame.contentWindow?.Array.from || Array.from;
+    document.body.removeChild(cleanFrame);
```
Juice10, line 57: Is there anyway we can do this without attaching the iframe? I'm afraid we might pick it up in recordings although the chance is super low.

eoghanmurray: We have the following in mutation.ts:
```
            if (!this.unattachedDoc) {
              try {
                // avoid upsetting original document from a Content Security point of view
                this.unattachedDoc =
                  document.implementation.createHTMLDocument();
              } catch (e) {
                // fallback to more direct method
                this.unattachedDoc = this.doc;
              }
            }
            const old = this.unattachedDoc.createElement('span');
```
            
Did you try something like that?

eoghanmurray: OK nevermind, looks like the unattachedDoc doesn't get the `contentWindow` created

eoghanmurray (approved):

daibhin:
Fwiw, I've debugged the exact same issue this week where a site had overridden the `Array.from` implementation. Would love to see protection against this but also have no problem telling people that the issue is on their end 😅

PR #1465, opened 2024-05-01
rrweb sends message to cross-origin iframe to force snapshot

colingm:
We often manually take snapshots in cross origin frame situations and were running into problems with no longer having the content of cross origin frames. We noticed that if we sent along a message to tell the cross origin frame to take a new snapshot whenever the parent frame took one then it resolved this issue.

### DIFF
--- a/packages/rrweb/src/types.ts
+++ b/packages/rrweb/src/types.ts
@@ -216,6 +216,7 @@ export type CrossOriginIframeMessageEventContent<T = eventWithTime> = {
   // The origin of the iframe which originally emits this message. It is used to check the integrity of message and to filter out the rrweb messages which are forwarded by some sites.
   origin: string;
   isCheckout?: boolean;
+  snapshot?: boolean;
 };
 export type CrossOriginIframeMessageEvent =
   MessageEvent<CrossOriginIframeMessageEventContent>;
--- a/.changeset/rotten-squids-run.md
+++ b/.changeset/rotten-squids-run.md
@@ -1,0 +1,5 @@
+---
+'rrweb': patch
+---
+
+rrweb sends message to cross-origin iframe to force snapshot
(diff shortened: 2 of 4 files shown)


eoghanmurray:
Is there a possibility that this will duplicate the regular way to take an iframe snapshot (when the contentDocument is accessible)?
I'm also curious as to if there is a problem with communicating the snapshot back up to the parent context when the contentDocument can't be accessed.
Some tests to demonstrate the novel scenario and prevent future regression would be appreciated.

colingm:
@eoghanmurray I am not quite sure if this can be triggered before the inner frame ends up being ready, will need to test that specifically. In practice we don't run into that problem since we don't try to initialize rrweb until the document is ready ourselves so it is worth double checking.

packages/rrweb/src/types.ts
```diff
@@ -216,6 +216,7 @@ export type CrossOriginIframeMessageEventContent<T = eventWithTime> = {
   // The origin of the iframe which originally emits this message. It is used to check the integrity of message and to filter out the rrweb messages which are forwarded by some sites.
   origin: string;
   isCheckout?: boolean;
+  snapshot?: boolean;
```
eoghanmurray, line 219: This note is a reminder that we need to add a test

eoghanmurray (changes requested):

colingm:
Adding a note here as doing some testing with timing of things and have found some oddities where it might not be behaving as expected due to the ordering of definition for `takeFullSnapshot` and `iframeManager`. Working through some testing and other options with the team.

dobrynin:
fwiw we also have this use case!

Juice10 (commented):
Looks really good @colingm! Curious what (edge) cases you've come across, and it would indeed be nice to have a test to make sure this stays working.

Easy-ish way to test this, is load rrweb (but don't record) in a page with cross origin iframe, then after everything is loaded: start recording. Iframe won't be recorded in the way its currently setup, but will with your fix.

colingm:
@Juice10 yeah sorry I have kind of let this languish unfortunately. So we actually found with the original version we had a weird race condition where it wasn't doing what we thought most of the time. We changed how we passed in the function (a setter now being called later) and that seemed to address the odd behavior we saw. Now just to write the tests 😅

PR #1468, opened 2024-05-03
inlineImages: Setting of `image.crossOrigin` is not always necessary

eoghanmurray:
Setting of `image.crossOrigin` is not necessary for same origin images, and causes an immediate image reload necessitating the load event listener (albeit from cache)

Changeset:
"inlineImages: during snapshot avoid adding an event listener for inlining of same-origin images (async listener mutates the snapshot which can be problematic)"


croqaz:
Hi, thanks for the PR! I didn't use this project anymore, so I won't test the code on my own.
But I understand the problem you're trying to solve and I can confirm that setting the crossOrigin triggers a browser request.
This fix looks great!
I would test it with some pages that have images stored in a CDN which is different than the website host, from what I remember this is what triggers the crossOrigin request. This would be just to make sure the feature is not broken. You could make a static HTML page and inject some random images from random domains and record that page to see that it works.

croqaz (approved):

eoghanmurray:
>  I would test it with some pages that have images stored in a CDN which is different than the website host, from what I remember this is what triggers the crossOrigin request. This would be just to make sure the feature is not broken. 

Thank you so much for suggesting this; in fact the new code had a bug not present in the prior version which I've fixed here: https://github.com/rrweb-io/rrweb/pull/1468/commits/236cd54144fb258be58014cffacd49e04e370e3f

packages/rrweb-snapshot/test/html/picture.html
```diff
@@ -1,9 +1,11 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
   <body>
     <picture>
+      <!-- these are 404 - not sure if that's intentional -->
       <source type="image/webp" srcset="assets/img/characters/robot.webp" />
       <img src="assets/img/characters/robot.png" />
     </picture>
     <img src="/images/robot.png" alt="This is a robot" />
+    <img src="https://avatars.githubusercontent.com/u/43396833?s=20&v=4" alt="CORS restricted but has access-control-allow-origin: *" />
```
eoghanmurray, line 9: I was also thinking of using https://www.rrweb.io/favicon.png here but that isn't served with the requisite `access-control-allow-origin: *` — @Yuyz0112 if you wish to add that header I can update the test to point to our own resource

Juice10: You could also check out how we do cross origin tests and do it in a similar fashion: https://github.com/rrweb-io/rrweb/blob/03b5216a9403f1509b4f69d1d71ef9874277fe91/packages/rrweb/test/record/cross-origin-iframes.test.ts#L138

Juice10 (approved):

packages/rrweb-snapshot/test/html/picture.html
eoghanmurray: I'll check if I can get our test server to set the access-control headers.

eoghanmurray: That's done in https://github.com/rrweb-io/rrweb/pull/1468/commits/d2df9bb673a13e5c78969b566c1d3b6161609fff
I can't believe it but it worked first time.

eoghanmurray: I didn't explicitly set or check that the access-control-origin header is actually being set; but I verified that failing to set the  `crossorigin="anonymous"` on the `` causes the test to fail

PR #1469, opened 2024-05-03
capture bitmap errors

daibhin:
Creating a bitmap image can throw an error. This wraps the call in `callbackWrapper` that will silence the error in cases where the `errorHandler` is specified


eoghanmurray:
Is the test failure unrelated or does it need to be addressed?

PR #1471, opened 2024-05-08
yarn format - prettier improvements & add .editorconfig

eoghanmurray:
- add .editorconfig (prettier falls back to config from that file)
 - fix that the .changeset/*.md files were being corrected by prettier, although the CHANGESETS plugin was generating them with double quotes
 - add test html files to yarn:format and fix some malformed HTML
 - .prettierignore: ignore deliberately malformed HTML files used for testing
 - convenience: add a `yarn format:head` command to just run prettier on the files you've recently changed


.changeset/format-head-prettier.md
```diff
@@ -0,0 +1,6 @@
+---
+"rrweb-snapshot": patch
+"rrweb": patch
+---
+
+development: add .editorconfig and improve how yarn format (prettier) works
```
Juice10, line 6: I generally leave empty changeset file as development changes aren't really needed for the Changelog

```suggestion
---
---
```

Juice10 (approved):

.prettierignore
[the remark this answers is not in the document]
eoghanmurray: It does, but the long list is deliberate; basically I want to ignore _old_ changeset files but apply to new ones

.prettierignore
```diff
@@ -0,0 +1,106 @@
+.changeset/attribute-text-reductions.md
```
eoghanmurray, line 1:
```suggestion
# list of old changeset files that were mutated to use single quotes by a previous version of prettier:
.changeset/attribute-text-reductions.md
```

Issue #1457, opened 2024-04-24
[Bug]: Uncaught SyntaxError: Regular expression too large


mrfullset:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

v2.0.0-alpha.13

### Expected Behavior

Live mode replay starts

### Actual Behavior

rrweb fails with an error in a console: `Uncaught SyntaxError: Invalid regular expression: (huge CSS contents) Regular expression too large`

### Steps to Reproduce

```js
rrweb.record({
        emit(event) {
          // send via websocket
        },
        recordCanvas: true,
        dataURLOptions: {
          type: 'image/webp',
          quality: 0.6,
        },
      });
```

```js
rrweb.Replayer([],
        { root: this.$refs.container,
          liveMode: true,
          UNSAFE_replayCanvas: true,
          mouseTail: {
            strokeStyle: "#d0d4e1"
          }
        })
```

### Additional Information

The same code as in "Steps to Reproduce" section worked with `v1.1.3`.
CSS inside of a regex are generated by Tailwind.

dobrynin:
We are also seeing this.

dobrynin:
It seems to be coming from https://github.com/rrweb-io/rrweb/blame/26c331b3b7e1a45091904fd1a6a875e16c9b777f/packages/rrweb-snapshot/src/rebuild.ts#L112-L118

colingm:
@priyankblitzz we didn't have any workaround other than rolling back rrweb to alpha.11

eoghanmurray:
I'm not sure what has changed such that it's now generating a bigger CSS regex on the first `selectorMatcher`, as #1431 didn't modify that.
Does anyone have a test website that reliably generates the issue? Is that `<SVG` embedded in CSS or is it in the document and somehow mistakenly going through the adaptCssForReplay function?

There's also #1458 which makes improvements and might be able to fix it although not quite currently ready for merge.

eoghanmurray:
Could you guys review or apply #1481 to verify that it fixes it?

priyankblitzz:
just tried the latest version(alpha 14) in my web app and it is working fine now. The issue is resolved for me now. Thank you so much for taking care of the issue.

PR #1478, opened 2024-05-16
Add a failing test for fixing

eoghanmurray:
Bug report in #1457

This bug was caused by changes in #1431

Run the test on it's own as follows:

`packages/rrweb-snapshot$ yarn retest --testPathPattern rebuild.test.ts -t tricky`


eoghanmurray:
I've isolated this failing css in a much smaller test case in #1481 which also fixes the issue

PR #1480, opened 2024-05-17
replay `:hover` rewrite - instead mutate stylesheets

eoghanmurray:
switch to model where we adapt the replay stylesheets programmatically after they have been added to the page instead of parsing the text

 - means we don't have to rely on regexp parsing of raw `cssText` text value to ensure we are substituting css selectors and media statements correctly
 - could be adapted in future for use against external (non inlined) `<link>` stylesheets during replay

#1458 points out failings in the current css parsing, and switches to `postcss`. This is an alternative to that and removes all 3rd party dependencies.

This PR could also be a contender as a fix for #1478 (although I suspect something else is the matter there, as I couldn't create a stylesheet with [the CSS text from that](https://github.com/rrweb-io/rrweb/blob/d864875ca0fe682c10943821b21e65e7d48f49ee/packages/rrweb-snapshot/test/tailwindCssText.ts))

How to help:

- [ ] satisfy eslint by removing the `ignore` line and upgrading typescript to a version that incorporates https://github.com/Microsoft/TypeScript/issues/23406 bugfix
- [ ] ensure we are applying `adaptStylesheetForReplay` after a virtual dom operation (i.e. fastforward in replayer which has a diff of adding a new stylesheet)
- [ ] add back the caching support (I'm not sure how as the stylesheet for which we've adapted things is mutable)
- [ ] verify replay performance is better or isn't affected


packages/rrweb-snapshot/test/rebuild.test.ts
```diff
@@ -4,12 +4,76 @@
...
+  adaptStylesheetForReplay,
   buildNodeWithSN,
   createCache,
 } from '../src/rebuild';
 import { NodeType } from '../src/types';
-import { createMirror, Mirror } from '../src/utils';
+import { createMirror, Mirror, stringifyStylesheet } from '../src/utils';
+
+/*
+function toEqualCss(actual: string, expected: string) {
```
eoghanmurray, line 15: I made an attempt to write my own `expect` matcher, but gave up due to typescript overhead and wrote `norm` instead

PR #1482, opened 2024-05-22
Removing global document references

AlfieJones:
In the rrweb-snapshot package, there are a few references to a global document object instead of using the doc parameter.
This pr moves them to the doc parameter inline with the rest of the package.

There's room for improving the typescript support to throw an error when accessing the document element but I wasn't able to figure out a clean way to achieve this.

### DIFF
--- a/.changeset/eleven-bobcats-peel.md
+++ b/.changeset/eleven-bobcats-peel.md
@@ -1,6 +1,6 @@
 ---
-'rrweb-snapshot': patch
-'rrweb': patch
+"rrweb-snapshot": patch
+"rrweb": patch
 ---
 
 better support for coexistence with older libraries (e.g. MooTools & Prototype.js) which modify the in-built `Array.from` function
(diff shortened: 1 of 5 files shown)


eoghanmurray:
Hi Alfie, thanks for your PR!

Can you explain if there was a situation where the use of the global `document` failed here, or is it just for completeness?
If it can fail maybe I can help you create a test for the situation.

AlfieJones:
I was running the rebuild outside of a dom environment using jsdom so any references to the global document would fail.

eoghanmurray:
Thank you.
I've added a cursory test to ensure we don't get a regression in the main code paths.
The one you fixed though was related to `canvas` and I couldn't recreate a failure using the jsdom in the tests without more effort.
Feel free to add to that test if you wish it to prevent regressions in that sense, but overall it's probably not necessary.  
If you think it's worthwhile to share what you tried in terms of typescript enforcement, I'd be interested to see how that might work.

eoghanmurray:
I also explored a test on the rebuild side, but don't think it's worth adding;
https://gist.github.com/eoghanmurray/52d0730eb8d7a209e27f29de0ddde455

eoghanmurray (approved):

PR #1485, opened 2024-05-23
Add options to mask values based on inclusion list instead of exclusion list

vincent-psarga:
Currently, the options to handle privacy are based on an exclusion list. This does not work fine for our use-case. The idea behind this pull request is to anonymize everything by default and allow the end user to specify which part of the recorded app should not be anonymized.

It would look something like this for example:

```
record({
  maskTextExcludeSelector: 'header, footer, nav.menu'
})
```

With such configuration, all content of the web page would be anonymized except for the header, footer and main navigation.

It also adds the possibility to have some text not masked inside a masked block.


eoghanmurray:
Thanks for your contribution!

One minor point is that `allowList` should be renamed to `maskTextAllowList` (or `maskTextWhitelist` maybe)

But; do you think your usecase could be handled by:

```
record({
  maskTextSelector: ':not(header, footer, nav.menu)'
})
```
?

This would avoid any need for code changes; I'm not sure this is a common enough scenario to warrant an extra config option, and I'm hoping to merge the `maskTextClass` and `maskTextSelector` and 'maskAllInputs' options (and similar options for 'block') at some point soon...

vincent-psarga:
I think this PR can be closed (I don't work anymore in the company for which I had to start this PR).

But to answer your question, using `:not(...)` does not work. Let's consider this HTML snippet:
```html
<html>
  <body>
    <header>Whatever...</header>
  </body>
</html>
```

and the following configuration:
```
record({
  maskTextSelector: ':not(header)'
})
```

it's going to mark everything, as the masking system checks if any ancestor of the current node matches the given CSS query. As the `body` tag matches `:not(header)`, all the page will be masked (including the header by the way :D )

PR #1490, opened 2024-06-03
fix: pass maskInputFn to snapshot

lewgordon-amplitude:
I saw the same change proposed in https://github.com/rrweb-io/rrweb/pull/1386, but not merged. I can't see why the tests were failing. I'd like to help get this through.


eoghanmurray:
Thanks @lewgordon-amplitude I'll close this as a duplicate of the other one you mention; I think that one is close to merging, I've just got a few questions on it.

lewgordon-amplitude:
Sounds good, thanks @eoghanmurray

PR #1493, opened 2024-06-04
No neg lookbehind

eoghanmurray:
Safari 16 doesn't support it

- this is a fixup for negative lookbehind which was recently added in  #1481
- instead I'm asserting the non-presence of a slash before a quote; which means these quoted bits can't be at the start of the content, which I think is okay.

A css construction that might now fail (I haven't tested this):

```
"my-quoted-css-rule" { color: blue; } 
```


Juice10 (approved):

Juice10:
Thanks @eoghanmurray!

PR #1501, opened 2024-06-10
Chore: Fix yarn dev

Juice10:
This pull request includes an upgrade to the turbo package and fixes an issue with the yarn dev command. The upgrade ensures that the latest version of turbo is used, while the fix addresses an issue with the dist directory being cleared which broke the dependants dev commands.


eoghanmurray:
I don't fully understand these changes but happy to approve :)

eoghanmurray (approved):

turbo.json
```diff
@@ -1,21 +1,42 @@
 {
   "$schema": "https://turborepo.org/schema.json",
-  "baseBranch": "origin/master",
-  "pipeline": {
+  // These root workspace files are reused in workspaces and may affect their build output
+  "globalDependencies": [
+    ".eslintrc.js",
+    ".prettierrc",
+    "lerna.json",
```
YunFeng0817, line 7: @Juice10 Do you think it's OK to remove lerna config since we are not using it anymore?

Juice10: Good catch! I've created a PR for it here: https://github.com/rrweb-io/rrweb/pull/1506

PR #1503, opened 2024-06-10
Support top-layer <dialog> recording & replay

### LINKED ISSUE #1381: [Feature Request]: Support top-layer / dialog HTML component
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a feature request that matches the one I want to file, without success.

rrweb

### Problem Description

RRWeb incorrectly positions the `dialog` modal.

Created a reproducible example where the modal should appear center vertical / horizontal in the body (as is circled using the mouse) but is aligned top in the replay:
https://rrwebdebug.com/play/index.html?url=http://jsonblob.com/1191711996356976640&version=2.0.0-alpha.11&virtual-dom=on&play=on

### Proposed Solution

Support the [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer) to layout the modal separately to the rest of the dom

### Alternatives Considered

None

### Additional Information

_No response_

Juice10:
Fixes #1381.

Adds support for recording & replaying [Top Layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer) `<dialog>` elements. These elements float over any other elements no matter the z-index.

What this PR does not do:
- trigger `showModal()` on rrweb-snapshot's `rebuild`
  - when rebuild is called we have no guarantee the element is added to the dom, and `.showModal()` has to be attached to be called, so the responsibility to do this is hoisted to rrweb & rrdom.
- add support for `popover` elements.
- add support for `requestFullScreen` elements.
- register order of top layer dialog elements, this is impossible to do without monkey patching `HTMLDialogElement.showModal`, and although that is possible I didn't have time to implement it. However most of the time this PR should sufficient. Unfortunately there is no way to query the dom for the order of elements in the top layer, as that would have made this doable without monkey patch, and recordable in rrweb-snapshot.
  - exact case where this breaks: 
  - ```ts
      const dialog1 = document.createElement('dialog') as HTMLDialogElement;
      dialog1.className = 'dialog1';
      document.body.appendChild(dialog1);
      const dialog2 = document.createElement('dialog') as HTMLDialogElement;
      dialog1.className = 'dialog2';
      document.body.appendChild(dialog2);
      dialog2.showModal(); // <== Note that dialog TWO is being triggered first, if dialog 1 was triggered first, this would have replayed as expected.
      dialog1.showModal();


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -704,6 +705,16 @@ function serializeElementNode(
...
     }
   }
+
+  if (
+    tagName === 'dialog' &&
+    (n as HTMLDialogElement).open &&
+    n.matches('dialog:modal')
+  ) {
+    (attributes as DialogAttributes).rr_open = 'modal';
+    delete attributes.open; // prevent default `show()` behavior which blocks `showModal()` from working
```
eoghanmurray, line 715: this could do with an expanded multi line comment
it looks like you are mutating the original website?

packages/rrweb/test/replay/dialog.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  hideMouseAnimation,
+  ISuite,
+  launchPuppeteer,
+  startServer,
+  waitForRAF,
+} from '../utils';
+
+expect.extend({ toMatchImageSnapshot });
+
+// TODO: test the following:
```
daibhin, line 29: Should this be removed before merging?

daibhin (approved):
Looks good!! Thanks for taking this on :)

packages/rrweb/test/replay/dialog.test.ts
Juice10: I'll remove the `// TODO`, but I'd like to keep some sort of history mainly to show that we don't yet have any tests or support replaying multiple dialogs in a specific order.
Maybe I should add a skipped test instead, what do you think?

packages/rrweb-snapshot/src/types.ts
```diff
@@ -103,6 +103,23 @@ export type mediaAttributes = {
...
 
+export type DialogAttributes = {
+  open: string;
+  /**
+   * Represents the dialog's open mode.
+   * `modal` means the dialog is opened with `showModal()`.
+   * `non-modal` means the dialog is opened with `show()` or
+   * by adding an `open` attribute.
+   */
+  rr_open: 'modal' | 'non-modal';
```
Juice10, line 114: To capture the state of a dialog element this PR is adding an extra attribute on dialog elements with the following name: `rr_open` (reason being its a companion to the dialog's `open` attribute).
Currently there are two possible values, `modal` to reflect it was called with `showModal` as opposed to `show` or just adding the `open` attribute. And `non-modal`.
Maybe there are better names for these. Especially `non-modal` maybe that should be `regular`, `plain` or `normal`?

daibhin: `non-modal` does seem like a weird name but I prefer that it is aligned with [mdn docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) rather than going with some of the other suggestions.

Given the `open` attribute is used for a slightly different purpose in the HTML spec would it make more sense to rename `rr_open` to something different like `presentation_style` or `mode`?

packages/rrweb/test/replay/dialog.test.ts
Juice10: I removed the `// TODO:` and replaced it with a skipped test

packages/rrweb-snapshot/src/types.ts
Juice10: Since it is a companion attribute to the `open` attribute I did want to have naming reflect that, so I ended up renaming it to `rr_openMode` for more clarity.

packages/rrweb/src/replay/index.ts
```diff
@@ -948,7 +957,8 @@ export class Replayer {
     });
     afterAppend(iframeEl.contentDocument! as Document, mutation.node.id);
 
-    for (const { mutationInQueue, builtNode } of collected) {
+    collectedDialogs.forEach((d) => applyDialogToTopLevel(d));
```
YunFeng0817, line 960: Would it be better to apply dialog after applied collected iframes?

packages/rrweb/src/replay/dialog/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  node: HTMLDialogElement | Node | RRNode,
+  attributeMutation?: attributeMutation,
+): void {
+  if (node.nodeName !== 'DIALOG' || node instanceof RRNode) return;
+  const dialog = node as HTMLDialogElement;
+  const isOpen = dialog.open;
+  const isModal = isOpen && dialog.matches('dialog:modal');
+  const rrOpen = dialog.getAttribute('rr_open_mode');
+
+  const shouldBeOpen =
```
YunFeng0817, line 20: Can you explain a bit more about this variable? I don't fully understand what is it here.

packages/rrweb/src/replay/index.ts
```diff
@@ -247,6 +248,7 @@ export class Replayer {
               this.applyStyleDeclaration(data, styleSheet);
           },
           afterAppend: (node: Node, id: number) => {
+            applyDialogToTopLevel(node);
```
YunFeng0817, line 251: I think this is redundant if we have added support to rrdom

YunFeng0817 (commented):
Thanks for implementing this. Overall, it looks good

packages/rrweb/src/replay/dialog/index.ts
Juice10: just renamed it all to `oldIsOpen` and `newIsOpen`, same for all the other variable names, let me know if this is more clear for you

packages/rrdom/src/document.ts
```diff
@@ -474,7 +474,7 @@ export class BaseRRElement extends BaseRRNode implements IRRElement {
   }
 
   public getAttribute(name: string): string | null {
-    return this.attributes[name] || null;
+    return this.attributes[name] ?? null;
```
eoghanmurray, line 477: had to read up about this operator (which is good to know about) but maybe a more explicit check for `undefined` -> returne null would be clearer?

Juice10:
```suggestion
    if (this.attributes[name] === undefined) return null;
    return this.attributes[name];
```

YunFeng0817 (approved):

PR #1507, opened 2024-06-12
fix: extension doesn't work after vite bump

YunFeng0817:
In #1033, the dependency vite-plugin-web-extension for extension package was upgraded as well along with vite
There are some breaking changes for vite-plugin-web-extension.


Juice10 (approved):

Juice10:
Good catch, @YunFeng0817!

PR #1509, opened 2024-06-14
Reverse monkey patch built in methods to support LWC

Juice10:
LWC monkey patches [a lot of](https://github.com/salesforce/lwc/blob/11058232634a18c530dc7aeee884694a81df7a9e/packages/%40lwc/synthetic-shadow/src/faux-shadow/node.ts#L256-L415) [built in](https://github.com/salesforce/lwc/blob/11058232634a18c530dc7aeee884694a81df7a9e/packages/%40lwc/synthetic-shadow/src/polyfills/mutation-observer/polyfill.ts#L288-L292) [methods](https://github.com/salesforce/lwc/tree/11058232634a18c530dc7aeee884694a81df7a9e/packages/%40lwc/synthetic-shadow/src/faux-shadow).

This PR gets around that issue by pulling new references from a temporary iframe, which hasn't been monkey patched yet.


eoghanmurray:
I've asked a SO question here to see if there are any ideas on how we might achieve the goals of this PR with less invasive code changes:
https://stackoverflow.com/questions/78633141/typescript-transform-property-to-function-call

Could we add a new (optional) output target, which pulls in the added functions, and also includes a rollup plugin to find/replace all the needed property accesses?

https://www.npmjs.com/package/rollup-plugin-modify#example-using-regexp-for-find-and-a-function-for-replace

eoghanmurray:
From SO, the first replier mentions https://www.npmjs.com/package/ts-patch

eoghanmurray:
I've just noticed that I've added code in `stylesheet-assets` branch that runs over `n.childNodes` but can accept either a `serializedElementNodeWithId` or a `HTMLStyleElement`:
https://github.com/rrweb-io/rrweb/pull/1475/files#diff-0d9a03e75eacff8f6be89336b49419b210455c1a81d8f46fae7cea2d83a8f93eR156

Presumably I'll have to do some type checking to figure out whether to call `n.childNodes` or `dom.childNodes(n)` etc. :(

YunFeng0817 (approved):
Great feature for recording LWS based websites!

PR #1511, opened 2024-06-17
Assuming all jest should have been removed in #1033

eoghanmurray:
Also added `retest:update`


Juice10 (approved):

YunFeng0817 (approved):

packages/rrweb-snapshot/package.json
```diff
@@ -5,10 +5,11 @@
   "scripts": {
     "prepare": "npm run prepack",
     "prepack": "npm run build",
-    "retest": "jest",
-    "test": "vitest run",
+    "retest": "vitest run",
+    "test": "yarn build && vitest run",
```
Juice10, line 9: We should probably decouple these a bit, turbo.json already has the relationship between build and test. If we add it here, build ends up being called twice

eoghanmurray: turbo.json is only in root; will we be adding one in this package too?
the difference between `test` and `retest` that I introduced was the idea that you have changed the code and want to check the tests, vs. modifying a test and you want to check if the modification fixes things (no need to build, hence 'retest').
Is there a way that can work with turbo, e.g. automatically it knows that there's no need to build as the files haven't changed?

Juice10: Turbo looks at the config file in the root to understand the relationships.

When you run a command with turbo for example `turbo run test -F PACKAGE` it will run build only if the code was unmodified, otherwise it will use the cached output.

PR #1515, opened 2024-06-18
Added support for deprecated addRule & removeRule methods

okejminja:
We recently had a client on our fork who still uses addRule & removeRule (bad practice, I know, but we had to support 'em)

This PR add support for deprecated methods [addRule](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/addRule) & [removeRule](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/removeRule)

Instead of writing duplicate code to support these or make some sort of an abstraction I just "transpiled" the props from addRule to what insertRule accepts and called the original insertRule or deleteRule accordingly – let you guys' methods handle the rest

Also expanded the test for capturing stylesheets to test this

LMK what you think


packages/rrweb/test/__snapshots__/record.test.ts.snap
```diff
@@ -2058,144 +2058,6 @@ exports[`record > captures style property changes 1`] = `
 ]"
 `;
 
-exports[`record > captures stylesheet rules 1`] = `
```
Juice10, line 2061: Looks like you removed this snapshot instead of committing the updated code with the changes you added to the `captures stylesheet rules` test.

Juice10 (changes requested):

packages/rrweb/test/__snapshots__/record.test.ts.snap
okejminja: Must've missed staging the file

packages/rrweb/src/record/observer.ts
```diff
@@ -657,6 +657,17 @@ function initStyleSheetObserver(
     ),
   });
 
+  // Support for deprecated addRule method
+  win.CSSStyleSheet.prototype.addRule = function (
+    this: CSSStyleSheet,
+    selector: string,
+    styleBlock: string,
+    index?: number,
```
eoghanmurray, line 665: If `index` is undefined, the default will be zero (the `insertRule` default)
However `addRule` default is cssRules.length https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/addRule#index
Could we supply this new default before passing it off to insertRule?

eoghanmurray (changes requested):

packages/rrweb/src/record/observer.ts
okejminja: Good catch! 
When supplying cssRules.length

```javascript
  win.CSSStyleSheet.prototype.addRule = function (
    this: CSSStyleSheet,
    selector: string,
    styleBlock: string,
    index: number = this.cssRules.length,
  ) {
    const rule = `${selector} { ${styleBlock} }`;
    return win.CSSStyleSheet.prototype.insertRule.apply(this, [rule, index]);
  };
```

I am getting an error when running the test, specifically when deleting rules
`DOMException: Failed to execute 'deleteRule' on 'CSSStyleSheet': The index provided (3) is larger than the maximum index (2).`

Will investigate and update

okejminja: I ended up separating the tests since idx kept moving around when calling deleteRule and removeRule in sequence

eoghanmurray (approved):

Juice10 (approved):

PR #1527, opened 2024-06-27
[Feature] Include takeFullSnapshot function in rrweb

arredgroup:
Added takeFullSnapshot function to allow retake a snapshot if user decide pause recording (avoid emit function) and after resume with an updated screen

### DIFF
--- a/.changeset/giant-rats-chew.md
+++ b/.changeset/giant-rats-chew.md
@@ -1,0 +1,5 @@
+---
+"rrweb": patch
+---
+
+Export takeFullSnapshot function for a recording process
--- a/packages/rrweb/src/index.ts
+++ b/packages/rrweb/src/index.ts
@@ -24,11 +24,13 @@ export type { recordOptions, ReplayPlugin } from './types';
 
 const { addCustomEvent } = record;
 const { freezePage } = record;
+const { takeFullSnapshot } = record;
 
 export {
   record,
   addCustomEvent,
   freezePage,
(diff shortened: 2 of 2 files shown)


Juice10:
Hi @arredgroup, thanks for the contribution. After reading your description I'm still not sure I fully understand the use case. Could you elaborate a bit on this topic?

Also maybe this is a better fit for @rrweb/record instead of rrweb itself, but that depends on the use case I guess.

arredgroup:
> Hi @arredgroup, thanks for the contribution. After reading your description I'm still not sure I fully understand the use case. Could you elaborate a bit on this topic?
> 
> Also maybe this is a better fit for @rrweb/record instead of rrweb itself, but that depends on the use case I guess.

Sure, we are using rrweb, but in some specific paths, for data security we need to pause the recording and continue, to do that, we decide to stop the emit event, with a return before data is enqueued, but we need to take a full snapshot to refresh the screen recording when pause is finished.

the idea is use this method to avoid stop the recording and start other.

YunFeng0817 (approved):
I agree with exporting this function explicitly. This function is defined similarly to `addCustomEvent` and `freezePage`, both of which were exported long ago. Therefore, it makes sense to also export takeFullSnapshot.

eoghanmurray:
Anything exported in `packages/rrweb/src/record/index.ts` should also be exported by `packages/record/src/index.ts` I think?

PR #1529, opened 2024-07-02
fix: console assert logging capture behaviour

pauldambra:
fixes https://github.com/PostHog/posthog-js/issues/1280

when someone calls console.assert the console recorder always captures as if the log was shown in the console

but...

> The console.assert() static method writes an error message to the console if the assertion is false. If the assertion is true, nothing happens.

from https://developer.mozilla.org/en-US/docs/Web/API/console/assert_static

this changes the recorder to only capture when the assertion is false, and not to log the first argument, which brings the recorded logs into line with what the browser would do

technically this is a breaking change since folk could be depending on this incorrect behaviour, i chose not to wrap it in config allowing people to opt in to this new behaviour to keep the change simpler (but open to adding that if necessary)

my educated guess is that use of `console.assert` is pretty rare


pauldambra:
ugh, the changeset command is very unhappy with something to do with how I've created this from our fork... I'll recreate

PR #1531, opened 2024-07-02
feat: allow custom truncation suffix

pauldambra:
when adding https://github.com/rrweb-io/rrweb/pull/1530 I realised that there's code in the console plugin for truncation that we're copying so our users have to download twice (not much but every little helps)

the only difference is we use (and some code expects) a different truncation marker

so, this 

* lets you add a custom truncation marker 
* and actively truncates strings not just objects that are being toString-ed

### DIFF
--- a/packages/plugins/rrweb-plugin-console-record/test/stringify.test.ts
+++ b/packages/plugins/rrweb-plugin-console-record/test/stringify.test.ts
@@ -6,7 +6,7 @@ import { stringify } from '../src/stringify';
 import { describe, it, expect } from 'vitest';
 
 describe('console record plugin', () => {
-  const aLongString = "a".repeat(100);
+  const aLongString = 'a'.repeat(100);
 
   it('can stringify bigint', () => {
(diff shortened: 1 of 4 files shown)


pauldambra:
@Juice10 noticed this while in there earlier... it's useful for us, but happy to drop it if not useful for the library generally :)

packages/plugins/rrweb-plugin-console-record/test/stringify.test.ts
```diff
@@ -6,7 +6,28 @@ import { stringify } from '../src/stringify';
 import { describe, it, expect } from 'vitest';
 
 describe('console record plugin', () => {
+  const aLongString = 'a'.repeat(100);
```
eoghanmurray, line 9:
```suggestion
  const aLongString = 'a'.repeat(100) + 'b';
```
(Ensure that we are testing that we're removing the end of the string and not the start)

pauldambra:
@eoghanmurray ping on this for next week (i'm just watering my PRs while #4 daughter naps :))

PR #1533, opened 2024-07-11
fix: nested stylesheets should have absolute URLs

jeffdnguyen:
hello 👋 - I identified an issue with nested stylesheets while looking into the [asset manager PR](https://github.com/rrweb-io/rrweb/pull/1475).

The PR has a [comment that outlines the problem in more detail](https://github.com/rrweb-io/rrweb/pull/1475#discussion_r1674425640)

The tl;dr is that for style tags, we are assuming any URLs inside of a style tag must share the same file path as the document the style tag is on. This causes nested stylesheets (think style tag that imports another stylesheet) to have incorrect urls in the snapshot. In order to fix this, I decided the best route would be to determine those absolute paths sooner when stringifying the stylesheets.


packages/rrweb-snapshot/src/utils.ts
```diff
@@ -351,3 +353,65 @@ export function extractFileExtension(
...
+  }
+  origin = origin.split('?')[0];
+  return origin;
+}
+
+const URL_IN_CSS_REF = /url\((?:(')([^']*)'|(")(.*?)"|([^)]*))\)/gm;
+const URL_PROTOCOL_MATCH = /^(?:[a-z+]+:)?\/\//i;
+const URL_WWW_MATCH = /^www\..*/i;
+const DATA_URI = /^(data:)([^,]*),(.*)/i;
+export function absoluteToStylesheet(
```
eoghanmurray, line 372: Mostly unrelated, but we could take the opportunity to rename this to e.g. `absolutifyURLs` or something else of your choice here.

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -351,3 +353,65 @@ export function extractFileExtension(
   const match = url.pathname.match(regex);
   return match?.[1] ?? null;
 }
+
+function extractOrigin(url: string): string {
+  let origin = '';
```
eoghanmurray, line 358: Also unrelated but we are moving towards `new URL` for this sort of thing; although we still probably want a `try/catch` around it.

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -96,11 +96,13 @@ export function escapeImportStatement(rule: CSSImportRule): string {
...
   try {
     const rules = s.rules || s.cssRules;
-    return rules
-      ? fixBrowserCompatibilityIssuesInCSS(
-          Array.from(rules, stringifyRule).join(''),
-        )
-      : null;
+    const stringifiedRules = Array.from(rules, stringifyRule)
+      .map((rule) => {
+        return s.href ? absoluteToStylesheet(rule, s.href) : rule;
```
eoghanmurray, line 101: Hmmm, why is this transformation not included in `stringifyRule`? e.g. after `importStringified` is created?

eoghanmurray: (the second argument to `Array.from` is a map function so with this addition we are running over the rules twice)

packages/rrweb-snapshot/src/utils.ts
jeffdnguyen: Love it! I honestly had the same thought too.

packages/rrweb-snapshot/src/utils.ts
jeffdnguyen: I think I'll leave this function alone for now and just stick with moving it into utils.ts.

packages/rrweb-snapshot/src/utils.ts
jeffdnguyen: Great point, we can avoid the extra map here. I didn't think of that tbh, I was just avoiding the effort to pass in the sheet's href down to stringify. Made the change!

Let me know if I overstepped with adding dom.iterable to the tsconfig. I was using [this](https://developer.mozilla.org/en-US/docs/Web/API/StyleSheetList#get_all_css_rules_for_the_document_using_array_methods) as an example. Alternatively I could just use a for loop.

packages/rrweb-snapshot/test/__snapshots__/integration.test.ts.snap
```diff
@@ -500,7 +500,7 @@ exports[`integration tests > [html file]: with-style-sheet-with-import.html 1`]
...
...css2?family=Open+Sans:wght@400;600;700&amp;family=Roboto:wght@100;300;400;500;700&amp;display=swap\\\\\\"\\");body { margin: 0px; background: url(\\"http://localhost:3030/a.jpg\\"); border-image: url(\\"data:image/svg+xml;utf8,&lt;svg xmlns=\\\\\\"http://www.w3.org/2000/svg\\\\\\" x=\\\\\\"0px\\\\\\" y=\\\\\\"0px\\\\\\" viewBox=\\\\\\"0 0 256 256\\\\\\"&gt;&lt;g&gt;&lt;g&gt;&lt;polygon points=\\\\\\"79.093,0 48.907,30.187 146.72,128 48.907,225.813 79.093,256 207.093,128\\\\\\"/&gt;&lt;/g&gt;&lt;/g&gt;&lt;/svg&gt;\\") 100% / 1 / 0 stretch; }p { color: red; background: url(\\"http://localhost:3030/css/b.jpg\\"); }body &gt; p { color: yellow; }body { margin: 0px; background: url(\\"http://localhost:3030/c.jpg\\"); border-image: url(\\"data:image/svg+xml;utf8,&lt;svg xmlns=\\\\\\"http://www.w3.org/2000/svg\\\\\\" x=\\\\\\"0px\\\\\\" y=\\\\\\"0px\\\\\\" viewBox=\\\\\\"0 0 256 256\\\\\\"&gt;&lt;g&gt;&lt;g&gt;&lt;polygon points=\\\\\\"79.093,0 48.907,30.187 146.72,128 48.907,225.813 79.093,256 207.093,128\\\\\\"/&gt;&lt;/g&gt;&lt;/g&gt;&lt;/svg&gt;\\") 100% / 1 / 0 stretch; }p { color: red; background: url(\\"http://localhost:3030/alt-css/d.jpg\\"); }body &gt; p { color: yellow; }</style>
```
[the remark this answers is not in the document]
eoghanmurray: (I'll do this in a direct commit)

eoghanmurray:
So something in https://github.com/rrweb-io/rrweb/pull/1533/commits/27432e2ec311396d185d265176e805aa286eb511 has caused the test to fail; unfortunately they didn't run automatically after you pushed to the PR, not sure why

eoghanmurray:
I've also got 3 commits in https://github.com/eoghanmurray/rrweb/tree/jn-correct-absolute-urls-for-nested-stylesheets that I'd like you to cherry-pick / pull in here first ... one of them adds a new test

jeffdnguyen:
> So something in [27432e2](https://github.com/rrweb-io/rrweb/commit/27432e2ec311396d185d265176e805aa286eb511) has caused the test to fail; unfortunately they didn't run automatically after you pushed to the PR, not sure why

I think I know why, moving the `absolultifyURLs` call into `stringifyRule` means I need to use the import rule's stylesheet's href and not the parent stylesheet href. Also need to call it not just after `importStringified` is populated in the case of a stylesheet with urls that wasn't imported (in this case use the parent href).

jeffdnguyen:
> I've also got 3 commits in https://github.com/eoghanmurray/rrweb/tree/jn-correct-absolute-urls-for-nested-stylesheets that I'd like you to cherry-pick / pull in here first ... one of them adds a new test. I'd need some sort of access credentials for `pendo-io` to be able to commit directly, not sure why that's not the case; maybe it's the 'allow access to maintainers' checkbox in your right hand column for this PR?

Weird...I'm not seeing that checkbox but will look into that. In the meantime happy to cherry pick!

packages/rrweb-snapshot/test/__snapshots__/integration.test.ts.snap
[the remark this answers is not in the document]
eoghanmurray: Actually I didn't get around to doing that in those mentioned commits.

eoghanmurray:
Also I'm urging the rest of the core team to merge in the following soon:
https://github.com/rrweb-io/rrweb/pull/1437

I imagine there would be extensive conflicts with this one, so if you are at all going back to the drawing board, then we could keep the tests from this PR and build upon that first.  If that's too much let's go ahead with this one and I'll handle the rebase on the other.

jeffdnguyen:
Update

- Cherry picked your commits (thanks for making those!)
- Handled that non import case
- Clarified with the jpg names as you suggested

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -96,36 +96,54 @@ export function escapeImportStatement(rule: CSSImportRule): string {
...
+
+      if (rule.styleSheet.href) {
+        importStringified = absolutifyURLs(
+          importStringified,
+          rule.styleSheet.href,
+        );
+      }
+
+      return importStringified;
     } catch (error) {
```
eoghanmurray, line 130: We shouldn't be adding new code within this try/catch (which should be better documented). I just checked, and `rule.cssText` is still available on CssImport rules so we can continue to fallback to that instead if we haven't managed to serialize here (the prior version did this)

eoghanmurray: Here's how that function was originally authored (for reference):

```
  let cssStringified = rule.cssText;
  if (isCSSImportRule(rule)) {
    try {
      cssStringified = getCssRulesString(rule.styleSheet) || cssStringified;
    } catch {
      // ignore
    }
  }
  return cssStringified;
```

eoghanmurray: the try/catch was probably to do with protecting against errors thrown by `rule.styleSheet`; presumably there could be restrictions on cross domain imports

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -96,36 +96,54 @@ export function escapeImportStatement(rule: CSSImportRule): string {
...
+    let importStringified;
     try {
       importStringified =
         // for same-origin stylesheets,
         // we can access the imported stylesheet rules directly
         stringifyStylesheet(rule.styleSheet) ||
         // work around browser issues with the raw string `@import url(...)` statement
         escapeImportStatement(rule);
+
+      if (rule.styleSheet.href) {
```
eoghanmurray, line 122: Great work I wouldn't have thought of this!
Works within <style> elements too!

eoghanmurray (changes requested):
Just the moving of code out of the try/catch and I think we'll be good to merge

packages/rrweb-snapshot/src/utils.ts
jeffdnguyen: Ah, makes sense to me. Thank you for the context!

jeffdnguyen:
Made the change! 🤞

eoghanmurray (approved):

eoghanmurray:
I've made a few more changes on top, including playing around with a benchmark as I was worried about the now-per-rule calls to `absolutifyURL` (which turned out fine)

https://github.com/pendo-io/rrweb/compare/jn-correct-absolute-urls-for-nested-stylesheets...eoghanmurray:rrweb:jn-correct-absolute-urls-for-nested-stylesheets

jeffdnguyen:
Cherry-picked your commits in. Nice job with deduping the fallback behavior, that feels much better! Glad the benchmark turned out fine, did not think to do that 😅

jeffdnguyen:
@eoghanmurray Please let me know if there is anything left I need to do to or can help with to get this merged 😄

Juice10:
Can we use postcss instead of regex to replace/absolutify the css? We've had a lot of issues with regular expressions not being the same as the spec and having some buggy edge cases. We've added postcss recently which should be much more stable.

jeffdnguyen:
> Can we use postcss instead of regex to replace/absolutify urls in the css? We've had a lot of issues with regular expressions not being the same as the spec and having some buggy edge cases. We've added postcss recently which should be much more stable.

@Juice10 I love the idea! Would you be open to me opening a seperate/followup PR to explore that instead? Mainly since I wanted to keep the scope of this PR to changing when we try to `absolutify` and not the how/implementation of the function.

eoghanmurray:
@Juice10 this PR is for the record side; postcss is used on the replay side, so big -1 from me for introducing postcss dependency for recording.  We've been using the absoluteURLs regex for ages already (this doesn't change that) and nobody has reported any issues.  

I did explore a couple of further commits on top of this PR which isolate the url regex replace to just the rule bit of the css (and not the selector bit). I figured this PR was complete enough so will introduce those changes properly in a new PR once this is complete.

Side note: postcss could be made redundant at replay time if I complete #1480 successfully.

Juice10 (approved):

Juice10:
> @Juice10 this PR is for the record side; postcss is used on the replay side, so big -1 from me for introducing postcss dependency for recording. We've been using the absoluteURLs regex for ages already (this doesn't change that) and nobody has reported any issues.

Fair point @eoghanmurray! Merging this as is, thanks for contributing @jeffdnguyen

PR #1535, opened 2024-07-16
Simplify the hover replacement function

eoghanmurray:
See #1458 and https://github.com/giuseppeg/postcss-pseudo-classes/blob/master/index.js for the full version; we didn't need all the features so we added `if (pseudoToCheck !== ':hover') {` but this PR goes further to reduce code.

### DIFF
--- a/.changeset/simplifify-hover-replacement.md
+++ b/.changeset/simplifify-hover-replacement.md
@@ -1,0 +1,6 @@
+---
+"rrweb-snapshot": patch
+"rrweb": patch
+---
+
+Slight simplification to how we replace :hover after #1458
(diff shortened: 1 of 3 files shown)


eoghanmurray:
This PR has resulted in a test change which wasn't anticipated, however instead of reverting I believe updating the tests is actually the correct path!

We added the test in question in #1440 and it looks like the following:

```
body > ul :is(li:not(:first-of-type) a:hover, li:not(:first-of-type).active a) {background: red;}
```

The focus then was on ensuring that the above wasn't parsed as two separate selectors because of the internal comma (which is actually a comma of the `:is` expression). I think we missed that there was also a `:hover` in there which should mean that the whole selector should be duplicated.  I've added a new test `doesn't ignore :hover within :is brackets` to formalize this expectation.

With the pseudoClassPlugin postcss plugin, and also in the version prior to that, everything between the brackets of `:is` was bundled with the `:is`, so the internal `:hover` was being ignored.

daibhin (approved):

Juice10 (approved):

PR #1536, opened 2024-07-18
perf: improve blocked and masking logic checks

JonasBa:
Porting the masking and block changes from https://github.com/rrweb-io/rrweb/pull/1271 

Related to https://github.com/rrweb-io/rrweb/pull/1271

### DIFF
--- a/packages/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb-snapshot/src/snapshot.ts
@@ -282,16 +282,16 @@ export function needMaskingText(
   } else {
     el = dom.parentElement(node)!;
   }
-    if (typeof maskTextClass === 'string') {
-      if (el.classList.contains(maskTextClass)) return true;
-      if (checkAncestors && el.matches(`.${maskTextClass} *`)) return true;
-    } else {
-      if (classMatchesRegex(el, maskTextClass, checkAncestors)) return true;
-    }
-    if (maskTextSelector) {
-      if (el.matches(maskTextSelector)) return true;
-      if (checkAncestors && el.matches(`${maskTextSelector} *`)) return true;
-    }
+  if (typeof maskTextClass === 'string') {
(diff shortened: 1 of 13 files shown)


packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -340,26 +340,16 @@ export function needMaskingText(
   } else {
     el = node.parentElement;
   }
-  try {
```
JonasBa, line 343: try/catch is an optimization [bailout](https://github.com/vhf/v8-bailout-reasons#trycatchstatement) reason in v8. It's unclear exactly why this code is here, but since this is an extremely hot path, we should remove it

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -340,25 +340,15 @@ export function needMaskingText(
...
-        if (el.closest(`.${maskTextClass}`)) return true;
-      } else {
-        if (el.classList.contains(maskTextClass)) return true;
-      }
-    } else {
-      if (classMatchesRegex(el, maskTextClass, checkAncestors)) return true;
-    }
-    if (maskTextSelector) {
-      if (checkAncestors) {
-        if (el.closest(maskTextSelector)) return true;
```
JonasBa, line 355: The closest check is unnecessary as we dont care about the actual node, we just want to know if it matches or not

eoghanmurray: Is the matches with the descendant selector more performant?

JonasBa: Yes, though lets make sure to run the benchmarks once @Juice10 merges the benchmark PR. the main difference is that the engine only applies the selector check to the element that it is testing as opposed to testing it on each parent node until it finds a match

PR #1539, opened 2024-07-19
ref: refactor pushadd to class member method

JonasBa:
I _think_ there is a good chance this gets optimized by the engines if the function reference is stable

related to https://github.com/rrweb-io/rrweb/pull/1271


eoghanmurray:
Can you remember why you closed this?

PR #1541, opened 2024-07-19
perf: coerce blockSelector and blockClass to a single concept and reduce branching

JonasBa:
this builds on top of https://github.com/rrweb-io/rrweb/pull/1536, the idea is to create a single blocking selector and remove some runtime checking and branching inside isBlocked

This is a stacked PR btw, once upstream is merged I will rebase so it'll only contain the related changes


JonasBa:
Discussed with maintainers: we can make this more efficient by constructing a validation function that is then passed down to the serializers

JonasBa:
Closing this in favor of a better alternative

eoghanmurray:
Do you have the other alternative as a PR @JonasBa ?

PR #1542, opened 2024-07-20
fix: benchmarks

JonasBa:
Fixes the benchmark runs and adds a title so its easier to tell what suite was profiled and make sense of the profiles

### DIFF
--- a/packages/rrweb/test/benchmark/dom-mutation.test.ts
+++ b/packages/rrweb/test/benchmark/dom-mutation.test.ts
@@ -146,7 +146,9 @@ describe('benchmark: mutation observer', () => {
       };
 
       // generate profile.json file
-      const profileFilename = `profile-${suite.title}-${new Date().toISOString()}.json`;
+      const profileFilename = `profile-${
+        suite.title
+      }-${new Date().toISOString()}.json`;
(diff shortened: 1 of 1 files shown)


JonasBa:
I dont think this needs a changelog since we dont require a package bump. Happy to open one if needed though

packages/rrweb/test/benchmark/dom-mutation.test.ts
```diff
@@ -110,12 +99,18 @@ describe('benchmark: mutation observer', () => {
...
         if ('html' in suite) {
           await page.goto('about:blank');
+          const code = fs
+            .readFileSync(path.resolve(__dirname, '../../dist/rrweb.umd.cjs'))
+            .toString();
+          await page.setContent(`<html>
+            <script>
+            ${code.toString()}
+            </script>
+            </html>`);
```
Juice10, line 109: What is the reason for moving this (back) to an inlined script element? This should make the profiles harder to read as the source map references are no longer intact.

JonasBa: @Juice10 I honestly dont remember, but afaik it had something to do with benchmark not working. I'm going to close the PR since you are working on the benchmarks anyways

PR #1544, opened 2024-07-20
ref: comptime lookup for tagName

JonasBa:
Attempt to bypass tagName regexp execution by performing a lookup for the most common static tagName values. Also skips a call to uppercase since tagName is already uppercase

Related to https://github.com/rrweb-io/rrweb/pull/1271


packages/rrdom/src/index.ts
```diff
@@ -202,7 +202,7 @@ function getValidTagName(element: HTMLElement): string {
   if (element instanceof HTMLFormElement) {
     return 'FORM';
   }
-  return element.tagName.toUpperCase();
+  return element.tagName;
```
Juice10, line 205: I believe for XML the tagNames are not always uppercased. But maybe this is more correct in that case. Also I'm not sure in what cases we record XML (maybe for some older XHTML websites?).

JonasBa: >I believe for XML the tagNames are not always uppercased.
Correct. The only think I'm unsure of if if this change has any downstream effects. I dont know if we rely on this to be uppercase, but my knowledge here is pretty limited.

eoghanmurray:
Could this be made with less boilerplate code if we populate HTML_TAGNAMES dynamically after first call for a given tag name?
i.e. would it give a similar benefit if this is a very hot path if the regex is called just once per given type of element?

PR #1545, opened 2024-07-20
ref: sort options obj and remove optional params

JonasBa:
Feel free to discard this PR, it's more of a suggestion than anything else and would require a new semver major due to the breaking API change, but I found it hard to track the options object around the different call sites (especially as a lot of field are optional and it's hard to know exactly what is being passed vs what is default initialized).

This PR changes that by removing changing all optional field of the serialize options params and sorting it. The goal of the change is to simplify the control flow by not having to look at what is default initialized as well as hopefully reducing the polymorphic argument values.


JonasBa:
@eoghanmurray lets skip it then and I will reopen it when your PR is merged. The idea is to reduce the number of argument shapes passed to the function which allows JS engines to make certain [optimizations](https://stackoverflow.com/questions/76167593/js-v8-compiler-optimization-monomorph-polymorph-sealed-object-and-classes-o) they otherwise wouldnt be able to.

eoghanmurray:
Thanks @JonasBa for the tricky work and hope we can revisit.
From my own POV (a programming pov) I dislike the many places you have to touch to get a new argument propagated down the call chain.  This is a problem when there are multiple open PRs which touch the arguments.  I don't know if your solution here also has that sort of simplification benefit.

PR #1561, opened 2024-08-27
fix(rrdom): Ignore invalid DOM attributes when diffing

billyvg:
We encountered an issue where replays with invalid attributes (e.g. `@click`) would break rendering the replay after seeking. The exception bubbles up to
[here](https://github.com/rrweb-io/rrweb/blob/62093d4385a09eb0980c2ac02d97eea5ce2882be/packages/rrweb/src/replay/index.ts#L270-L279), which means the replay will continue to play, but the replay mirror will be incomplete.

### DIFF
--- a/.changeset/chilled-penguins-sin.md
+++ b/.changeset/chilled-penguins-sin.md
@@ -1,0 +1,5 @@
+---
+"rrdom": patch
+---
+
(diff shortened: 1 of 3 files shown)


YunFeng0817 (approved):
@billyvg Thanks for the fix

PR #1568, opened 2024-09-13
feat: add CI for publishing chrome extension

YunFeng0817:
Changes:
1. Because we are upgrading svelte, the link for the rrweb-player css file needs to be updated.
2. Remove a permission that is not used in the extension
3. Add a CI to automatically publish the extension to the chrome web store
4. Improve conversion of rrweb version to chrome extension version


.github/workflows/extension-release.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: NODE_OPTIONS='--max-old-space-size=4096' DISABLE_WORKER_INLINING=true yarn turbo run prepublish --filter=@rrweb/web-extension
+
+      - uses: mnao305/chrome-extension-upload@v5.0.0
+        with:
+          extension-id: 'pdaldeopoccdhlkabbkcjmecmmoninhe'
+          file-path: ./packages/web-extension/dist/chrome.zip
+          client-id: ${{ secrets.CWS_CLIENT_ID }}
+          client-secret: ${{ secrets.CWS_CLIENT_SECRET }}
+          refresh-token: ${{ secrets.CWS_REFRESH_TOKEN }}
+          publish: true
```
Juice10, line 29: As far as I can tell from here: https://github.com/changesets/action?tab=readme-ov-file#custom-publishing

You need to move the following code (including the new `if` statement) to `.github/workflows/release.yml`:

```
      - uses: mnao305/chrome-extension-upload@v5.0.0
        if: steps.changesets.outputs.hasChangesets == 'false'
        with:
          extension-id: 'pdaldeopoccdhlkabbkcjmecmmoninhe'
          file-path: ./packages/web-extension/dist/chrome.zip
          client-id: ${{ secrets.CWS_CLIENT_ID }}
          client-secret: ${{ secrets.CWS_CLIENT_SECRET }}
          refresh-token: ${{ secrets.CWS_REFRESH_TOKEN }}
          publish: true
```

Juice10 (changes requested):

Juice10 (approved):

mohammedzamakhan:
How would I even use this fix in my chrome extension that uses rrweb package? This fix only seems to fix chrome extension for this repo!!

ruiconti:
Question: why do this as a CI step that only fixes it for `web-extension`, instead of fixing this for all packages published by this repo that ends up importing the worker?

https://github.com/rrweb-io/rrweb/issues/1578#issuecomment-2612855699

Juice10:
> @Juice10 @YunFeng0817
> 
> Question: why do this as a CI step that only fixes it for `web-extension`, instead of fixing this for all packages published by this repo that ends up importing the worker?
> 
> [#1578 (comment)](https://github.com/rrweb-io/rrweb/issues/1578#issuecomment-2612855699)

All (or at least most) of the test we have rely on the rrweb recording and the web worker being bundled together into a `rrweb.js` file that gets injected into puppeteer via `page.evaluate(rrwebCode)`. I wanted to keep this simple when I introduced the web-workers so I decided to inline it.

A secondary consideration was the reports of people copy pasting the `rrweb.(c)js` files into their codebase and expecting it to be a contained package. 
Because of these two reasons, but mainly not wanting to complicate our test setup & build setup, I decided we should keep it a self contained package, at least on the recording side. (replay side also has a stylesheet that needs including).

I'm very much open to reconsidering this, just not sure if I have to time to redo our test setup as of right now.

Issue #1575, opened 2024-09-26
[Bug]: 


otan:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb-player

### Version

v2.0.0-alpha.11

### Expected Behavior

No crashes when a style sheet has multiple child nodes where the individual nodes don't form valid CSS, e.g.

```js
const s = document.createElement("style"); s.append(document.createTextNode("a {")); s.append(document.createTextNode("color: red; }"));
document.head.appendChild(s);
```

### Actual Behavior

Crashes as `postcss` encounters invalid CSS.
Encountered on multiple websites in the wild after https://github.com/rrweb-io/rrweb/pull/1437 was merged by @eoghanmurray 

### Steps to Reproduce

Add this test to `packages/rrweb-snapshot/test/css.test.ts`

```
  it('sad applies css splits correctly', () => {
    // sad path
    const arranHalfCssText = '.a { background-color';
    const arranOtherHalfCssText = ': red; }';
    const markedCssText = [arranHalfCssText, arranOtherHalfCssText].join('/* rr_split */');
    applyCssSplits(sn, markedCssText, true, mockLastUnusedArg);
    expect((sn.childNodes[0] as textNode).textContent).toEqual(arranHalfCssText);
    expect((sn.childNodes[1] as textNode).textContent).toEqual(
      arranOtherHalfCssText,
    );
  });
```

Vadman97:
+1, something we're seeing as well

eoghanmurray:
Fixed between #1600 and #1580

shadowusr:
@eoghanmurray Thanks a lot for the fix and your work on this package! We are suffering from this bug as well, and it's very common in our setting. Is there any way to get the bleeding-edge version of rrweb that has this fix included or maybe there are plans to release alpha 19 any time soon? Thanks!

eoghanmurray:
We are gearing up for another alpha release but it requires a bit of co-ordination among the team as to what is in vs. out.
In the meantime it's perfectly fine to compile from source, I've been doing so from the start; I'm not sure what your setup is but I'd recommend that approach if this bug is affecting you.

PR #1580, opened 2024-10-08
fix: Handle exceptions thrown from postcss when calling adaptCssForReplay

guntherjh:
This PR adds a `try/catch` around `postcss().process()` so that exceptions won't cause playback to crash. I was encountering issues where invalid CSS was being passed to `postcss` which would result in [CssSyntaxError](https://postcss.org/api/#csssyntaxerror-column). I am still investigating why invalid CSS is passed to [adaptCssForReplay](https://postcss.org/api/#csssyntaxerror-column) in the first place, but for now this change should at least make sure any exceptions from `postcss` are at least handled and logged. I am currently using `warn` to log the exceptions but am happy to switch that to something else if need be. 

Related discussion in rrweb slack: https://rrweb.slack.com/archives/C01BYDC5C93/p1727975312250899

### DIFF
--- a/.changeset/angry-turtles-provide.md
+++ b/.changeset/angry-turtles-provide.md
@@ -1,0 +1,5 @@
+---
+"rrweb-snapshot": patch
+---
+
+Handle exceptions thrown from postcss when calling adaptCssForReplay
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -63,7 +63,7 @@ function getTagName(n: elementNode): string {
 export function adaptCssForReplay(cssText: string, cache: BuildCache): string {
   const cachedStyle = cache?.stylesWithHoverClass.get(cssText);
   if (cachedStyle) return cachedStyle;
-  
+
   let result = cssText;
   try {
     const ast: { css: string } = postcss([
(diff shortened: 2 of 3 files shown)


otan:
heads up, i think the root cause is https://github.com/rrweb-io/rrweb/issues/1575

guntherjh:
@otan Good call out! That definitely seems like one of the ways invalid CSS can be passed to `postcss`. I'm also looking at another potential cause and will report back here if I find anything useful (or if it turns out to be the same problem as the one described in the issue you linked).

For this particular PR, I just wanted to simply make sure exceptions from `postcss` were handled so it wouldn't cause playback to crash.

daibhin (approved):

eoghanmurray:
#1600 should fix that those types of fragments get passed into postcss in the first place.  Happy to add this try/except but hopefully it shouldn't be encountered after #1600 (assuming the underlying css is fine)

guntherjh:
@eoghanmurray thanks for taking a look at my PR! I'll need to double check the original issue I was encountering to see if https://github.com/rrweb-io/rrweb/pull/1600 fixes it (I thought I checked and it didn't but I honestly can't remember 😓). In any case, I am inclined to keep the try/catch just to prevent any future exceptions from postcss from going unhandled. I'm happy to make any updates to how the exceptions are logged (or any other issue you see). Thanks again for the review!

eoghanmurray (approved):

PR #1588, opened 2024-10-22
Constructed stylesheets snapshotted, rebuilt with `rrweb-snapshot`

skitterm:


eoghanmurray:
Is it possible to construct a test case to both better illustrate exactly how to produce the scenario you are recording, as well as better illustrate the PR?

PR #1597, opened 2024-11-19
fix: angular wrapped mutationobserver detection

pauldambra:
When Angular wraps MutationObserver and rrweb tracks mutations you get trapped in a fight with Angular's changedetection and performance suffers. We had multiple reports at posthog of performance degrading with recording active - even to the point of freezing the page entirely

We discovered that the checks for whether to load mutation observer from an iFrame were not correctly detecting the "taint" of angular wrapping the mutation observer.

It turns out detecting whether `Zone` is on the global object is a good enough proxy for Angular having tainted something

(there is some hand-waving here since you can configure zone.js to use a different global object name, but my guess is most people don't, and anyone with enough tech savvy to do that on purpose can also run the recorder outside of the angular zone)

This is confirmed as fixing performance issues in posthog-js v 1.187.1

You can see a wrapped MutationObserver from angular in this screenshot

Screenshot 2024-11-28 at 19 05 39

a posthog customer provided this repro https://github.com/typedb/angular-posthog-freeze
with that when session recording started the page froze immediately
with this fix it does not

### DIFF
--- a/.changeset/moody-experts-build.md
+++ b/.changeset/moody-experts-build.md
@@ -1,0 +1,5 @@
+---
+"@rrweb/record": patch
+---
+
+correctly detect when angular has wrapped mutation observer"
--- a/.changeset/moody-experts-build.md
+++ b/.changeset/moody-experts-build.md
@@ -2,4 +2,4 @@
 "@rrweb/record": patch
 ---
 
-correctly detect when angular has wrapped mutation observer"
+Correctly detect when angular has wrapped mutation observer
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -28,23 +28,21 @@ const testableMethods = {
 
 const untaintedBasePrototype: Partial<BasePrototypeCache> = {};
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const isFunction = (x: unknown): x is (...args: any[]) => any => {
-  return typeof x === 'function';
-};
-
 /*
- When angular patches things they pass the above `isNativeFunction` check
+ When angular patches things - particularly the MutationObserver -
+ they pass the `isNativeFunction` check
  That then causes performance issues
- because angular's change detection
+ because Angular's change detection
  doesn't like sharing a mutation observer
+ Checking for the presence of the Zone object
+ on global is a good-enough proxy for Angular
+ to cover most cases
+ (you can configure zone.js to have a different name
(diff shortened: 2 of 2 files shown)


packages/utils/src/index.ts
```diff
@@ -63,7 +82,11 @@ export function getUntaintedPrototype<T extends keyof BasePrototypeCache>(
       ),
   );
 
-  if (isUntaintedAccessors && isUntaintedMethods) {
+  if (
+    isUntaintedAccessors &&
+    isUntaintedMethods &&
+    !isAngularZonePatchedFunction(defaultObj)
```
pauldambra, line 88: considered only checking this for the known "bad" case of `MutationObserver` but it's a fast enough check to always be sure

packages/utils/src/index.ts
```diff
@@ -28,6 +28,25 @@ const testableMethods = {
...
+ because angular's change detection
+ doesn't like sharing a mutation observer
+ */
+export const isAngularZonePatchedFunction = (x: unknown): boolean => {
+  if (!isFunction(x)) {
+    return false;
+  }
+  const prototypeKeys = Object.getOwnPropertyNames(x.prototype || {});
+  return prototypeKeys.some((key) => key.indexOf('__zone'));
+};
```
pauldambra, line 48: ah, no, i've just had to change this in prod at posthog

will wait for a few more customers to report it works now

turns out I'm still an idiot, `-1` is truthy, and this always returns true

pauldambra: interestingly, that meant that all browsers and sessions were treating all things as tainted and loading them from an iframe

in safari and mobile safari particularly this meant often there was no mutation observer

which suggests that the iframe loading code is bugged in some way (or just not guaranteed maybe)

pauldambra: yep, lots of reports that this change has fixed things for folk who aren't using Angular
And for folk using Angular there's no more reports of perfomance issues

But I still see folk where they are using Angular and they can't get untainted things from an iframe when using Safari

.changeset/moody-experts-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rrweb/record": patch
+---
+
+correctly detect when angular has wrapped mutation observer"
```
Juice10, line 5:
```suggestion
Correctly detect when angular has wrapped mutation observer
```

Juice10 (approved):

Issue #1596, opened 2024-11-08
[Bug]: genTextAreaValueMutation doesn't seem to mask attribute.value


sdemjanenko:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb-snapshot

### Version

2.0.0-alpha.17

### Expected Behavior

I have a React app that controls a `textarea` by setting the `value` prop. RRWeb records the attribute change as well as the `change` event. The data looks like

Attribute change event:
```
{
    "adds": [],
    "attributes": [
        {
            "attributes": {
                "value": "(903) 993-7305"
            },
            "id": 121
        }
    ],
    "removes": [],
    "source": 0,
    "texts": []
} 
```

Input change event:
```
{
    "id": 121,
    "isChecked": false,
    "source": 5,
    "text": "**************"
}
```

I am expecting the `attributes.value` to be masked.

### Actual Behavior

The `attributes.value` is unmasked. The issue appears to be in `genTextAreaValueMutation`, specifically the code

```
item.attributes.value = Array.from(
  dom.childNodes(textarea),
  (cn) => dom.textContent(cn) || '',
).join('');
```

There doesn't appear to be any masking logic that is applied to this value (or upstream of this function).

### Steps to Reproduce

Create a controlled `textarea` in React.

eoghanmurray:
Could you verify that #1596 fixed things?
My tests only recreate the bug when setting the textarea value using `innerText`
So this will only fix things if that's what React does when you say 'setting the value prop' .

sdemjanenko:
This was coming up when I wrapped a textarea with https://www.npmjs.com/package/typed.js. In that package it has logic that looks like
```
replaceText(str) {
    if (this.attr) {
      this.el.setAttribute(this.attr, str);
    } else {
      if (this.isInput) {
        this.el.value = str;
      } else if (this.contentType === 'html') {
        this.el.innerHTML = str;
      } else {
        this.el.textContent = str;
      }
    }
  }
```
which matches your `textContent` comment.

PR #1599, opened 2024-11-29
Fix for #1596 - we missed a path where masking could be skipped

eoghanmurray:
See #1596, although I'm not fully convinced that is the case I've fixed here

### DIFF
--- a/packages/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/src/record/mutation.ts
@@ -533,7 +533,7 @@ export default class MutationBuffer {
       this.attributes.push(item);
       this.attributeMap.set(textarea, item);
     }
-    let value = Array.from(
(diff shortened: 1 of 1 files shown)


packages/rrweb/src/record/mutation.ts
```diff
@@ -533,10 +533,18 @@ export default class MutationBuffer {
       this.attributes.push(item);
       this.attributeMap.set(textarea, item);
     }
-    item.attributes.value = Array.from(
+    let value = Array.from(
```
Juice10, line 536: You're getting an eslint error here. Should be the following to fix it:
```suggestion
    const value = Array.from(
```

Juice10 (approved):

YunFeng0817 (approved):

PR #1613, opened 2024-12-17
fix: remote CSS does not get rebuilt properly

billyvg:
This fixes an issue where inlined CSS from a remotely loaded `<link>` does not get applied properly due to object reference mutation. The incremental mutation of the link element was then being skipped because the diff-er did not detect a change. This leads to the CSS not being loaded at all.


billyvg:
Superceded by https://github.com/rrweb-io/rrweb/pull/1618

Issue #228, opened 2020-06-02
Replay Error: Uncaught TypeError: Cannot read property 'querySelectorAll' of null


mpstv:
Hi.
Sometimes when replay record, i getting an error:
```
Uncaught TypeError: Cannot read property 'querySelectorAll' of null
``` 
I did not reveal the pattern of error, but at a high level of review, this happens on this [line](https://github.com/rrweb-io/rrweb/blob/ee73bd4919b5db1d380981fd9249a7452cdef474/src/replay/index.ts#L747).

Yuyz0112:
Hi @mpstv, could you provide the events for debugging?

mpstv:
I introduced some changes to [repository](https://github.com/mpstv/angular-components-rrweb-problem) and now you can use it to reproduce this error.

![querySelectorAll_reproduce](https://user-images.githubusercontent.com/54762420/83517236-36127e80-a4e1-11ea-9b31-00165d34c0b1.gif)

For using this app you should clone repository and run command:
```
npm i && ng serve
``` 

After it app will be available on `localhost:4200`

PR #1614, opened 2024-12-17
fix(snapshot): Change to ignore all `link[rel="modulepreload"]` (#228)

billyvg:
... removes requirement to have `as="script"` attribute as it is optional (also `modulepreload` implies script).

Also fixes `prefetch` to extract file extension instead of checking end of string since there can be URL parameters.

### DIFF
--- a/.changeset/short-hounds-confess.md
+++ b/.changeset/short-hounds-confess.md
@@ -1,0 +1,6 @@
+---
+"rrweb-snapshot": patch
+"rrweb": patch
+---
+
+Change to ignore all link[rel="modulepreload"] instead of including only those with `as="script"`
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -269,7 +269,7 @@ function buildNode(
             continue;
           } else if (
             tagName === 'link' &&
-            (n.attributes.rel === 'preload' ||
+            ((n.attributes.rel === 'preload' && n.attributes.as === 'script') ||
               n.attributes.rel === 'modulepreload')
           ) {
             // ignore
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -7,7 +7,12 @@ import {
   type legacyAttributes,
 } from '@rrweb/types';
 import { type tagMap, type BuildCache } from './types';
-import { isElement, Mirror, isNodeMetaEqual } from './utils';
+import {
+  isElement,
+  Mirror,
+  isNodeMetaEqual,
+  extractFileExtension,
+} from './utils';
 import postcss from 'postcss';
 
 const tagMap: tagMap = {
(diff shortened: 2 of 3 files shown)


packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -265,15 +270,14 @@ function buildNode(
...
-            n.attributes.as === 'script'
+              n.attributes.rel === 'modulepreload')
           ) {
             // ignore
           } else if (
             tagName === 'link' &&
             n.attributes.rel === 'prefetch' &&
             typeof n.attributes.href === 'string' &&
-            n.attributes.href.endsWith('.js')
+            extractFileExtension(n.attributes.href) === 'js'
```
billyvg, line 280: This is what is done in [`snapshot.ts`](https://github.com/getsentry/rrweb/blob/fix-snapshot-link-modulepreload-ignore-all-up/packages/rrweb-snapshot/src/snapshot.ts#L828)

packages/rrweb-snapshot/src/snapshot.ts
```diff
@@ -820,8 +820,7 @@ function slimDOMExcluded(
         // (module)preload link
         (sn.tagName === 'link' &&
           (sn.attributes.rel === 'preload' ||
```
eoghanmurray, line 822:
```suggestion
          ((sn.attributes.rel === 'preload' && sn.attributes.as === 'script') ||
```
The rel="preload" and as="script" originally went together before https://github.com/rrweb-io/rrweb/pull/1128/files

eoghanmurray (changes requested):

packages/rrweb-snapshot/src/snapshot.ts
billyvg: For `preload`, should we check: `as="script"` OR file extension == "js"? 

(Users don't always write HTML according to spec)

eoghanmurray: I believe that the `as` part is necessary for the preload to 'work' correctly:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#as
So I'd be happy to conform to that.  
Otherwise we could add it to the following clause as in 

```
-        sn.attributes.rel === 'prefetch' &&
+        `['prefetch', 'preload'].includes(sn.attributes.rel) &&
```

but personally I'm not convinced that's necessary

eoghanmurray (approved):

.changeset/short-hounds-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+"rrweb-snapshot": patch
+"rrweb": patch
+---
+
+Change to ignore all link[rel="modulepreload"] instead of including only those with `as="script"`
```
eoghanmurray, line 6:
```suggestion
Change to ignore all link[rel="modulepreload"] instead of including only those with `as="script"`
Properly check file extension on prefetch <link> elements to correctly ignore url params
```

Issue #1603, opened 2024-12-06
[Bug]: splitCssText causes degraded performance when recording


guntherjh:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

v2.0.0-alpha.18

### Expected Behavior

rrweb should not have a significant impact on an application/site when recording data. 

### Actual Behavior

Whenever `markCssSplits` is called, we have observed significantly degraded performance caused by splitCssText maxing out the JS heap. 

### Steps to Reproduce

**Reproduction steps with browser extensions:**
In the wild, we were seeing this issue with customers who were using chrome extensions that are injecting CSS in to the page. We've seen the issue with our customers using the Grammarly browser extension and Sider AI browser extension. To reproduce the issue this way:

1. Install one these two extensions.
2. Open a page that has rrweb recordind data.
3. In the case of Grammarly, you just need to click in an input in a form and that should trigger the extension to start injecting code that will trigger the issue. For Sider AI, you need to open the extension and a side bar will appear that triggers the issue. 

**Simple reproduction steps (not using a browser extension):**

1. Serve the following HTML file using a simple http server. For example. I used [http-server](https://github.com/http-party/http-server). 
```html
<html>
<head>
    // test.css can really contain anything, it's super important for this test
    <link href="test.css"rel="stylesheet">
    // benchmark.css is the css file used for benchmarking in /packages/rrweb-snaphsot/test/css
    // choosen b/c it's pretty large
    <link href="benchmark.css"rel="stylesheet">
    <title>Home</title>
    <textarea></textarea>
    <script type="module">
        // This is the record.js file from /packages/record/dist created after running yarn && yarn build:all in the root of the repo
        // commented out for now
        // import { record } from './record.js';
        
        const stylesheet = document.getElementsByTagName('link')[1].sheet; 

        let cssString = '';
        // Check if the stylesheet is loaded
        if (stylesheet.rules) {
            // Get all the CSS rules as a string
            for (let i = 0; i < stylesheet.cssRules.length; i++) {
            cssString += stylesheet.cssRules[i].cssText;
            }
        }

        var recorder = record({
            emit(event) {
                console.log(event);
            },
        });
        
        const newStyleEl = document.createElement('style');
        document.body.appendChild(newStyleEl);
[...]

PR #1615, opened 2024-12-18
Fix performance of splitCssText 

eoghanmurray:
See #1603 for an excellent bug report.

I didn't include the benchmark from that report as it didn't demonstrate the pathological cases that were being experienced in the wild, but rather just the degree of slowdown of the 'split' vs. 'no split' code paths.

----

See #1437 for the context as to why `splitCssText` exists; 
To recap, a <style> element can have multiple text nodes. We currently serialize by processing styleEl.sheet.cssRules into a single string, but if one of the text nodes is programmatically modified (via a text mutation), then we want to be able to map the mutation back to only modify the relevant part, and not blow away the entire css text if we were not to do the split in the first place.

This PR massively improves the performance of the splitting in the case where we need to search through large strings to find similar parts, we need to compare after normalization so there's a lot of back and forth.  This PR changes that process to more like a binary search rather than a crawling search which was producing the pathological performance.

### DIFF
--- a/.changeset/efficiently-splitCssText-1603.md
+++ b/.changeset/efficiently-splitCssText-1603.md
@@ -1,0 +1,6 @@
+---
+"rrweb-snapshot": patch
+"rrweb": patch
+---
+
+Improve performance of splitCssText for <style> elements with large css content - see #1603
--- a/.puppeteerrc.cjs
+++ b/.puppeteerrc.cjs
@@ -6,5 +6,5 @@ const { join } = require('path');
 module.exports = {
   // Changes the cache location for Puppeteer.
   cacheDirectory: join(__dirname, '.cache', 'puppeteer'),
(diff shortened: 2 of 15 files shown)


eoghanmurray:
There's a lot of commits in here that can be ignored as I was iterating trying to fix an issue in Github Actions

eoghanmurray:
See #1437 for the context as to why `splitCssText` exists; 
To recap, a <style> element can have multiple text nodes. We currently serialize by processing styleEl.sheet.cssRules into a single string, but if one of the text nodes is programmatically modified (via a text mutation), then we want to be able to map the mutation back to only modify the relevant part, and not blow away the entire css text if we were not to do the split in the first place.

Juice10 (approved):

Juice10:
Lets add a PR for the GitHub Actions change and merge this one after that has happened

Issue #1560, opened 2024-08-27
[Feature Request]: Is there a way to merge events within a specified period of time?


wqcstrong:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a feature request that matches the one I want to file, without success.

### What package is this feature request for?

rrweb

### Problem Description

![image](https://github.com/user-attachments/assets/e8d69f0b-6b27-4858-ba66-dfe78ebc163b)

### Proposed Solution

/

### Alternatives Considered

/

wfk007:
@wqcstrong I have the idea to solve it, but In what scenarios would this be used?

wfk007:
@wqcstrong 
I have published a npm package named `rrweb-events-merge` powered by `rrweb` and `rrweb-snapshot`.

You can try it with:

1. install dependencies

```shell
npm install rrweb-events-merge
```

2. import and generate merged events

```ts
import { mergeEvents } from "rrweb-events-merge";

const mergedEvents = mergeEvents({
  events: events,
  startTimeStamp: startTimeStamp,
  endTimeStamp: endTimeStamp,
  iframe: iframe,
  snapshotOptions: {
    mirror,
  },
  onError: (error: string) => {
    console.error(error);
  },
});
```

Here is the repo: https://github.com/wfk007/rrweb-events-merge

wqcstrong:
@wfk007 Thx, I will take a look ❤️

> but In what scenarios would this be used?

This repository https://github.com/HuolalaTech/page-spy-web/ provides the ability to replay runtime information on browser applications through log files. It now also supports log trimming, but now is achieved by manipulating time information, meaning the file size of rrweb remains unchanged. If events could be merged, it would be a significant breakthrough.

wfk007:
@wqcstrong I opened a PR: https://github.com/rrweb-io/rrweb/pull/1616. Please help to confirm if this PR can solve your problem at your convenience.

wqcstrong:
> @wqcstrong I opened a PR: #1616. Please help to confirm if this PR can solve your problem at your convenience.

Your implementation converts the complete events into `mergedEvents` in replay and pass it to Replayer. What I am hoping for is the ability to manipulate them after the recording is finished, before upload / download events and  there is no `iframe`. The data flow is:
1. User open the page: start recording;
2. User finish some actions and click "Stop recording" button, let's say the size of total events is 10Mb;
[...]

PR #1616, opened 2024-12-23
feat: merge events

wfk007:
feat for: https://github.com/rrweb-io/rrweb/issues/1560


eoghanmurray:
The following fixes the zygote issue in the tests:
https://github.com/rrweb-io/rrweb/pull/1615/commits/c8b05950002524970b9abb5018f1bf5769ebb21c

eoghanmurray:
I think this does not solve the problem in #1560 
I do have another plan to implement a 'frames per second' limit on mutations.  The motivation there is where multiple mutations are operating on the same DOM element, e.g. javascript based animation.  The fruit of such a plan would be the ability to merge events at record time, which might allow just a single, presumably smaller, mutation to be used for the scenario the OP wants.  This approach I have in mind of directly manipulating and merging the mutation events might also be a candidate to replace `rrdom` (not that that isn't already doing a good job).

PR #1618, opened 2025-01-06
fix: remote CSS does not get rebuilt properly

billyvg:
This fixes an issue where inlined CSS from a remotely loaded `<link>` does not get applied properly due to object reference mutation.

### DIFF
--- a/.changeset/red-peaches-explode.md
+++ b/.changeset/red-peaches-explode.md
@@ -1,0 +1,5 @@
+---
+"rrweb": patch
+---
+
+This fixes an issue where inlined CSS from a remotely loaded `<link>` does not get applied properly due to object reference mutation.
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -7,7 +7,7 @@ concurrency: ${{ github.workflow }}-${{ github.ref }}
 jobs:
   release:
     name: Tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v3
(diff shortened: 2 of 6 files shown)


eoghanmurray:
The Object.assign bit was introduced in [#995](https://github.com/rrweb-io/rrweb/pull/995)
I'm wondering if it had special significance there in terms of how the link element is updated ... have pinged Yun.

YunFeng0817:
@billyvg The only difference is that Object.assign overwrites the attribute value in the mirror. Your change does not change the meta in the mirror.
Can you please help me understand why this difference can cause the bug?

billyvg:
@YunFeng0817 Yeah so I'm not sure if we are relying on changing the meta in the mirror elsewhere. The reason for this change is as follows:

* [buildNodeWithSN() is called with the meta from mirror](https://github.com/getsentry/rrweb/blob/fix-snapshot-remote-css-rebuild/packages/rrweb/src/replay/index.ts#L1795)
* [Inside of buildNodeWithSN, we check if the meta in mirror is equal to _itself_](https://github.com/rrweb-io/rrweb/blob/dc20cd45cc63058325784444af6bd32ed2cace48/packages/rrweb-snapshot/src/rebuild.ts#L435-L438)

YunFeng0817:
@billyvg Thank you for the clean explanation. I now understand the cause of the issue. 
[buildNodeWithSN() is called with the meta from mirror](https://github.com/getsentry/rrweb/blob/fix-snapshot-remote-css-rebuild/packages/rrweb/src/replay/index.ts#L1795) I made this for preventing duplicate nodes in the replayer. But a remotely loaded <link> is an exception for the check.

Can you add 
```
Object.assign(
  newSn.attributes,
  mutation.attributes as attributes,
);
```
to the place after the newNode is created? This can keep the mirror sync with Dom status

billyvg:
@YunFeng0817 Updated!

YunFeng0817 (approved):
LGTM

Issue #1619, opened 2025-01-08
[Bug]:  I am getting a white screen after an incremental snapshot and it gets restored after the next snapshot. I am not getting this issue every time I record the session. Warning- [replayer] Exception in mutation Failed to execute 'appendChild' on 'Node': Nodes of type '#document' may not be inserted inside nodes of type 'DIV'.


sandeshmurdia:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

v2.0.0-alpha.7

### Expected Behavior

It should work fine and not give white screen

![image](https://github.com/user-attachments/assets/d72e2f6b-8873-4407-a746-275863bb61be)

### Actual Behavior

giving white screen

### Steps to Reproduce

it is happening randomly

mohitmaliLL:
I'm also facing this issue.

Issue #1414, opened 2024-02-21
[Bug]: [replayer] Exception in mutation Failed to execute 'appendChild' on 'Node': Nodes of type '#document' may not be inserted inside nodes of type 'META'.


macalinao:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

v2.0.0-alpha.11

### Expected Behavior

Should not crash the replayer

### Actual Behavior

Replayer crashes

### Steps to Reproduce

Use RRWeb on etsy.com

### Testcase Gist URL

https://rrwebdebug.com/play/index.html?url=https%3A%2F%2Fgist.github.com%2Fmacalinao%2Fbb2505ada4a5536a28879cfd670109ae%2Fraw%2Fb256cc21c3094ecb6c8be48afef8e3eb503ada1b%2Fbotstudio-virtual-browser-session-b807712e-492d-4a0e-98a2-716b3e00403d.json&version=2.0.0-alpha.11&virtual-dom=on&play=on

mindaugasrukas:
We observe the same issue with all versions up to v2.0.0-alpha.17 (on many sites). It seems like the bug is related to iframes.

charlie632:
Is there any solution to this?

eoghanmurray:
So in the example recording on rrwebDebug, there is a `<meta content="en-PH">` node with ID of 15
There is an iframe mutation event which should correspond to the loading of the iframe contents.
The parent ID for this event is the meta element, instead of an `<iframe>` element.
There doesn't appear to be any iframes on the page.
I think something must have gotten messed up either at recording time or in later processing.
The recording appears to be from etsy.com; was it recorded with an extension?

PR #1624, opened 2025-01-13
Fix/workaround iframe parent

salomita:
This is a workaround to avoid the replayer crashing when there's an error when trying to attach a #document to a non-iframe.

The replayer has a bug in which all the recording looks blank. This has been reported a couple of times:
https://github.com/rrweb-io/rrweb/issues/1414
https://github.com/rrweb-io/rrweb/issues/1619

It is happening because the replayer is trying to appendChild(#document) onto a non-iframe. The workaround prevents it from doing it and prevents the replay from going blank. Still, this is a workaround and further investigation is needed to check if the event is corrupted or if there's a mismatch in the order

### DIFF
--- a/.changeset/brave-planes-begin.md
+++ b/.changeset/brave-planes-begin.md
@@ -1,0 +1,5 @@
+---
+"rrweb": patch
+---
+
+Workaround for avoid the replayer from crashing when there's an error when trying to attach a document to a non-iframe
--- a/packages/rrweb/src/replay/index.ts
+++ b/packages/rrweb/src/replay/index.ts
@@ -1516,7 +1516,7 @@ export class Replayer {
         console.warn(
           '[Replayer] Skipping invalid document append to a non-iframe parent.',
           mutation,
-          parent
+          parent,
         );
         return;
       }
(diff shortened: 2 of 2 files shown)


eoghanmurray:
There are failing tests there and I believe the changeset has broken something (rather than it being flaky tests)

YunFeng0817:
@salomita Can you please add a change set to the PR https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md
The workaround looks good. Can you just give us more information about this case?
For example, what is the case you encountered and is it a bug on the recorder side or a corner case that needs to be handled on the replayer side?

salomita:
> @salomita Can you please add a change set to the PR https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md The workaround looks good. Can you just give us more information about this case? For example, what is the case you encountered and is it a bug on the recorder side or a corner case that needs to be handled on the replayer side?

Added more information, let me know if I should add more

eoghanmurray:
This PR potentially plasters over an underlying problem where perhaps the id of the parent is incorrect?
Was there a recording on https://rrwebdebug.com/ that exhibits the crashing behaviour?

PR #1594, opened 2024-11-02
fix: don't bundle everything into console recorder

pauldambra:
| before | after |
|-|-|
|before|Screenshot 2024-11-03 at 00 13 35|

the use of `utils.patch` from `rrweb` in the `console` plugin was including the whole of rrweb in the console recorder.

move the patch function into the existing `@rrweb/utils` package fixes this

(I've only proven the bundle is fixed - i'm hoping CI proves the recorder still works 😅 )

just looking at bundle size this reduces the console recorder plugin from 255kb to 15kb

PR #1630, opened 2025-01-17
chore(ci): track bundle size

pauldambra:
split out of https://github.com/rrweb-io/rrweb/pull/1594 which I managed to break :/

this adds two useful CI tracking mechanisms

1) it adds a bundle tracker which comments back onto PRs showing change in size of bundled artefacts

we've found this super useful in the posthog SDK 
i don't think i've access such that it can comment back on this PR, so i'm guessing we'd need to merge it and then see what output it has and correct (maybe)

2) it outputs an HTML file for each artefact showing what makes up the size of each bundle

very useful when you see bundle change is too high or changes to investigate what is causing the problem 

image

(I didn't include a changeset, since it's just a CI change)


pauldambra:
cc @Juice10 since you reviewed the original PR that i have now messed up 🙈

Juice10 (approved):

Juice10:
Thanks Paul! Just merged it

Issue #1627, opened 2025-01-15
[Bug]: Injected iframe cleanup on error


lapritchett:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

Other (specify below)

### Version

v2.0.0-alpha.18

### Expected Behavior

@rrweb/record and rrweb-snapshot should clean up any iframes that are injected for the purposes of getting clean prototypes.

### Actual Behavior

In @rrweb/record and rrweb-snapshot, `getUntaintedPrototype` contains code that injects an iframe but doesn't always properly clean it up (in my case, when calling const win = iframeEl.contentWindow; doesn't work within an iframe):

```
  try {
    const iframeEl = document.createElement("iframe");
    document.body.appendChild(iframeEl);
    const win = iframeEl.contentWindow;
    if (!win) return defaultObj.prototype;
    const untaintedObject = win[key].prototype;
    document.body.removeChild(iframeEl);
    if (!untaintedObject) return defaultPrototype;
    return untaintedBasePrototype[key] = untaintedObject;
  } catch {
    return defaultPrototype;
  }
  ```

### Steps to Reproduce

Enable @rrweb/record
Go to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe
Paste the following into the HTML input:
```
<iframe
  id="inlineFrameExample"
  title="Inline Frame Example"
  width="300"
  name="Example Website"
  src="https://example.com"
  title="Example.com"
  ref={iframeRef}
  sandbox="allow-scripts"
  height="200"
  src="">
</iframe>
```
Click on "Example Domain" when it renders in the iframe
Note a new blank iframe is injected and not removed
![CleanShot 2025-01-14 at 18 16 42@2x](https://github.com/user-attachments/assets/2c69ed53-afea-446d-9029-79de44fbd92c)

### Additional Information

  I suggest adding a finally clause to handle the iframe cleanup:
  
  ```
    let iframeEl;
  try {
    iframeEl = document.createElement("iframe");
    document.body.appendChild(iframeEl);
    const win = iframeEl.contentWindow;
    if (!win) return defaultObj.prototype;
    const untaintedObject = win[key].prototype;
    if (!untaintedObject) return defaultPrototype;
    return untaintedBasePrototype[key] = untaintedObject;
  } catch {
    return defaultPrototype;
  } finally {
    try {
[...]

PR #1633, opened 2025-01-18
fix: untainted prototype access in Angular

pauldambra:
we're seeing multiple reports of issues with the mutation observer in Safari

this is most often in relation to Angular apps 

considering we recently improved detection of need to load untainted observers when using Angular

when testing locally I can always see that Safari gives a mutation observer from an iframe - but I never see that mutation observer notify of any mutations. 

this PR moves back to the approach that it turns out rrweb had in V1

The Angular zone stores unpatched things that it has patched in a discoverable location. This checks for that and if it is present uses it. 

This means that in practice Angular won't trigger safari to load a mutation observer from an iframe and so (in testing) reliably captures mutations in safari for angular apps

### DIFF
--- a/.changeset/tall-poems-develop.md
+++ b/.changeset/tall-poems-develop.md
@@ -1,0 +1,5 @@
+---
+"@rrweb/utils": patch
+---
+
+safer access to iframes in safari for untainted prototypes
--- a/.changeset/tall-poems-develop.md
+++ b/.changeset/tall-poems-develop.md
@@ -2,4 +2,4 @@
 "@rrweb/utils": patch
 ---
 
-safer access to iframes in safari for untainted prototypes
+load unpatched versions of things from Angular zone when present
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -80,29 +80,34 @@ export function getUntaintedPrototype<T extends keyof BasePrototypeCache>(
       ),
   );
 
-  if (isUntaintedAccessors && isUntaintedMethods && !isAngularZonePresent()) {
-    untaintedBasePrototype[key] = defaultObj.prototype as BasePrototypeCache[T];
-    return defaultObj.prototype as BasePrototypeCache[T];
+  const isUntainted = isUntaintedAccessors && isUntaintedMethods && !isAngularZonePresent();
+  // we're going to default to what we do have
+  let impl: BasePrototypeCache[T] = defaultObj.prototype as BasePrototypeCache[T]
+  // but if it is tainted
+  if (!isUntainted) {
+    // try to load a fresh copy from a sandbox iframe
+    let iframeEl: HTMLIFrameElement | undefined = undefined;
+    try {
+      iframeEl = document.createElement('iframe');
+      iframeEl.hidden = true;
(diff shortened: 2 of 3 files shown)


packages/utils/src/index.ts
```diff
@@ -80,29 +80,36 @@ export function getUntaintedPrototype<T extends keyof BasePrototypeCache>(
...
+  let impl: BasePrototypeCache[T] =
+    defaultObj.prototype as BasePrototypeCache[T];
+  // but if it is tainted
+  if (!isUntainted) {
+    // try to load a fresh copy from a sandbox iframe
+    let iframeEl: HTMLIFrameElement | undefined = undefined;
+    try {
+      iframeEl = document.createElement('iframe');
+      iframeEl.hidden = true;
+      document.head.appendChild(iframeEl);
```
pauldambra, line 95: functionally these two lines are the change

the rest _i believe_ is just refactoring to reduce the number of returns

pauldambra:
i think also fixes https://github.com/rrweb-io/rrweb/issues/1627

pauldambra:
customer testing shows this isn't the whole picture... will keep digging

eoghanmurray:
I'm sceptical about adding it to `document.head` ([against spec](https://stackoverflow.com/questions/26861329/is-the-iframe-tag-allowed-in-the-head-tag-according-to-the-w3c)?) but the sentry precedence is powerful so will approve when you are ready with any more updates.

pauldambra:
the change to head didn't fix it... or didn't fix it enough 🤣 

i'm hoping the customer is still willing to test more - i think i'll try polling for the content window next to see if it's not ready when we try to read from it

it may all be a red herring here ofc 🙈

arredgroup:
Hi! Exists any news related to this fix? I had the same issue with dynamic iframes in my react app

pauldambra:
I need to update this PR. In the end I went back to how rrweb v1 used to do this, since Angular Zone advertises unpatched versions in a known location

Safari was perfectly capable of giving me a mutation observer from an iframe but I could never get that observer to trigger anything

.changeset/tall-poems-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rrweb/utils": patch
+---
+
+safer access to iframes in safari for untainted prototypes
```
pauldambra, line 5:
```suggestion
load unpatched versions of things from Angular zone when present
```

packages/utils/src/index.ts
```diff
@@ -80,29 +97,35 @@ export function getUntaintedPrototype<T extends keyof BasePrototypeCache>(
       ),
   );
 
-  if (isUntaintedAccessors && isUntaintedMethods && !isAngularZonePresent()) {
-    untaintedBasePrototype[key] = defaultObj.prototype as BasePrototypeCache[T];
-    return defaultObj.prototype as BasePrototypeCache[T];
+  const isUntainted = isUntaintedAccessors && isUntaintedMethods;
+  // we're going to default to what we do have
+  let impl: BasePrototypeCache[T] =
```
pauldambra, line 102: i kept the reorganization here since I did find it difficult to track what was happening previously

PR #1636, opened 2025-01-23
Fix cross-origin-iframes test failure (currently a problem in multiple PRs)

eoghanmurray:
Fix up the 'should replace the existing DOM nodes on iframe navigation with `isAttachIframe`' test

 - It was working for me when the test was run in isolation (`-t` option), but when the entire cross-origin-iframes test was run, the change of iframe contents didn't seem to happen in time
 - I've had to investigate what the full snapshot should look like, and have saved a good snapshot to an out-of-band snapshot file in case anyone needs to debug in future, rather than updating `__snapshots__/cross-origin-iframes.test.ts.snap` as I like that the original test was minimal


packages/rrweb/test/record/cross-origin-iframes.test.ts
```diff
@@ -213,15 +213,22 @@ describe('cross origin iframes', function (this: ISuite) {
     it('should replace the existing DOM nodes on iframe navigation with `isAttachIframe`', async () => {
       await ctx.page.evaluate((url) => {
         const iframe = document.querySelector('iframe') as HTMLIFrameElement;
-        iframe.src = `${url}/html/form.html?2`;
+        iframe.src = `${url}/html/empty.html`;
```
eoghanmurray, line 216: I've just changed it to `empty` in order to reduce the size of the __snapshot__ file above, but also at one point I suspected that the iframe load event wasn't coming in as a new version of puppeteer was 'caching' it or the content wasn't otherwise changing (rrdom?!).  I'm pretty sure now that that was the case, but that it was just a case of the `waitForRAF` not waiting long enough.

eoghanmurray: The `empty` html file does have some content that we now depend on, namely `div#one`

PR #1638, opened 2025-01-27
update record and playback to account for parent node missing when it…

mikeyagner:
… will be added later


mikeyagner:
Performance testing the dom mutations since we are adding a recursive loop. No significant increase in execution time
image

PR #1640, opened 2025-01-29
Fix splitCssText again

eoghanmurray:
See also https://github.com/PostHog/posthog-js/issues/1668 for the downstream bug report, and further discussion in Slack (Thanks Paul D'Ambra for report and pointers)

This is a further improvement after performance fixes in #1615 

This covers new scenarios as outlined in the tests.  Test cases were recreated from a very large inline style node in https://hiring.workfully.com/signin which looks like it was in a shadow root `:host(.productfruits--container)` although I can't quite find it there now.  I pulled the examples files from a breakpoint and have them locally, but the test cases here incorporate the important bit, including the split in the middle of a statement.
Ultimately the problem with the content which triggered this case was that `margin-top: 0;` as authored, gets serialized to `margin-top: 0px;`, which was preventing us finding the right point to split between normalized/unnormalized.


packages/rrweb-snapshot/src/utils.ts
```diff
@@ -489,31 +501,62 @@ export function splitCssText(
...
+            j = jLimit + 1; // trigger end of search
+          } else if (j === textContentNorm.length - 1) {
+            // we're about to end loop without a split point
+            splitNorm = cssTextNorm.indexOf(startSubstring);
+          }
+          if (cssNormSplits.length >= 2 && j > jLimit) {
+            const prevTextContent = childNodes[i - 1].textContent;
+            if (prevTextContent && typeof prevTextContent === 'string') {
+              // pick the first matching point which respects the previous chunk's approx size
+              const prevMinLength = normalizeCssString(prevTextContent).length;
```
pauldambra, line 538: i guess it's somewhere here that means if you run a test twice (or some multiple times) then you don't get the same output

(at least in my experience of testing whether this was deterministic when trying to figure out what was happening)

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -450,8 +450,15 @@ export function absolutifyURLs(cssText: string | null, href: string): string {
...
-export function normalizeCssString(cssText: string): string {
-  return cssText.replace(/(\/\*[^*]*\*\/)|[\s;]/g, '');
+export function normalizeCssString(
+  cssText: string,
+  _testNoPxNorm = false,
+): string {
+  if (_testNoPxNorm) {
+    return cssText.replace(/(\/\*[^*]*\*\/)|[\s;]/g, '');
+  } else {
+    return cssText.replace(/(\/\*[^*]*\*\/)|[\s;]/g, '').replace(/0px/g, '0');
```
pauldambra, line 460: i briefly tested a very naive loop parser and it was slower than regex replace - i guess because browsers/v8 are doing some magic to optimise this already

but I didn't test it over a range of inputs

this is (from my testing) at best O(n) for whitespace - and for clarity since i'm not much of a comp sci person. if you insert whitespace into the input then this gets slower the more whitespace is present

pauldambra (commented):
i am really not sure how i could roll this out in prod
the last version basically doubled our support load and i'm ending this week exhausted as a result

i have no idea of how to test if this fixes for all cases or just for a specific case

we seem to be writing a css parser, and i'm wondering if adopting a css parser would be safer

packages/rrweb-snapshot/src/utils.ts
eoghanmurray: The performance of the normalization function could be improved.  I've moreso tried to ensure it's not called repeatedly on the same piece of css (with the 'binary search' style changes in #1615 ).

eoghanmurray:
We're not attempting to write a CSS parser at record time, we are using `cssRules` to parse the CSS which uses the native browser capabilities.  

The `splitCssText` function just passes through the output for <style> elements that have only one child, so these fixes are very much for edge cases.  The mutation issues that this splitting solves are also mostly theoretical, so it's possible to patch/short-circuit the `splitCssText` function (directly `return [cssText]`) for guaranteed performance, at the expense of correctness of the edge cases documented in the 'css splitter' suite.   I introduced this splitting approach in #1437 and never dreamed it would cause performance issues so I didn't write it with performance in mind.

> we seem to be writing a css parser 
I appreciate that the algorithm implemented here is not simple; I've thought about abstracting it out to a third party library "split a string according to an array of related substrings which can be matched via a normalization function" ... I haven't looked into whether such a thing already exists.

This PR is definitely an improvement and I believe catches the last pathological case, particularly as now there is now an additional `jLimit` iteration limit.  

There is another [large PR](#1475) to move CSS parsing off the main thread at record time, however that would likely have hidden this problem rather than bringing it to the fore so painfully.

I've also another plan to ditch the whole `cssRules` approach, and hence the need to do any matching of split points, as we'd just use the textContent directly, when we can detect that the style element hasn't been modified programmatically, however I'm waiting for #1475 to get merged before we can look at that.

packages/rrweb-snapshot/src/utils.ts
eoghanmurray: I don't quite understand this comment, but I would say the algorithm is indeed deterministic, but maybe you mean it will behave differently based on different sized inputs because of the `jLimit` bit?

pauldambra: oh, i wrote a test that ran the split multiple times and compared the output and it didn't match
was generally whitespace ending up on different sides of a split

Juice10 (approved):

Fox32:
@Juice10 @eoghanmurray Is it possible to get a new pre release containing this fix? This fix is quite significant and the issues makes the latest release unusable for us. Thanks for fixing this!

packages/rrweb-snapshot/src/utils.ts
raghavsethi: Coming in without much context, please forgive me if this is dumb:

Is this expected to replace only the 0px case or also the 10px case?
```
.replace(/0px/g, '0');
```

PR #1642, opened 2025-02-03
Allow more fine grained control over blocking elements

kevinatown:
This PR introduces a new option `blockElementFN` to `rrweb.record` options. This option allows the consumer to add a callback function to have more control over if the element is blocked or not. This will allow consumers to have a very strict blocking policy and whitelist content.


eoghanmurray:
There was an effort in #1541 to merge the two previous block methods into one argument that could be passed down.  That's not a requirement to get this merged, but just to make note.

kevinatown:
Oh interesting. It seems that they were thinking about something similar: https://github.com/rrweb-io/rrweb/pull/1541#issuecomment-2245420570. Happy to tackle that and remove the other two options, but that would be a breaking change and bigger than just a patch. Also willing to do it as a follow up.

PR #1646, opened 2025-02-04
[chore] Cache yarn packages for CI

Juice10:
TODO:
- [x] use cache in ci-cd.yml
- [x] use cache in releases.yml
- [x] use cache in style-check.yml


eoghanmurray:
I don't understand what dangers there might be with `cache: yarn` but I guess we'll find out!

eoghanmurray (approved):

PR #1651, opened 2025-02-06
fix: address security risks in GitHub Actions workflows

AdnaneKhan:
The recent [pull request](https://github.com/rrweb-io/rrweb/pull/1646) mentions [I don't understand what dangers there might be with cache: yarn but I guess we'll find out!](https://github.com/rrweb-io/rrweb/pull/1646#issuecomment-2634649223).

It turns out, this adds a lot of danger. An attacker can compromise the NPM token and the Chrome extension release secrets fairly easily by leveraging a technique known as GitHub Actions cache poisoning.

This happens because the `style-check.yml` workflow runs on `pull_request_target`. I've updated the workflows to do a few things:

* Remove cache consumption in release job
* Tighten `GITHUB_TOKEN` permissions in other workflows that consume the cache

Ideally, you should move away from running user-controlled code from forks on `pull_request_target`, but that requires a more substantial overhaul to the repository's CI. The changes here will make it impossible to directly compromise the release secrets.

Some background on cache poisoning:

* https://adnanthekhan.com/2024/12/21/cacheract-the-monster-in-your-build-cache/
* https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/

### DIFF
--- a/.github/workflows/style-check.yml
+++ b/.github/workflows/style-check.yml
@@ -5,7 +5,8 @@ on: [push, pull_request_target]
 jobs:
   eslint_check_upload:
     runs-on: ubuntu-latest
-    permissions: read-all
+    permissions:
+      contents: read
     name: ESLint Check and Report Upload
 
     steps:
@@ -55,7 +56,8 @@ jobs:
   prettier_check:
     # In the forked PR, it's hard to format code and push to the branch directly, so the action only check the format correctness.
     if: github.event_name != 'push' && github.event.pull_request.head.repo.full_name != 'rrweb-io/rrweb'
-    permissions: read-all
+    permissions:
+      contents: read
     runs-on: ubuntu-latest
     name: Format Check
     steps:
--- a/.github/workflows/style-check.yml
+++ b/.github/workflows/style-check.yml
@@ -5,6 +5,7 @@ on: [push, pull_request_target]
 jobs:
   eslint_check_upload:
     runs-on: ubuntu-latest
+    permissions: read-all
     name: ESLint Check and Report Upload
 
     steps:
@@ -36,6 +37,8 @@ jobs:
   annotation:
     # Skip the annotation action in push events
     if: github.event_name == 'pull_request_target'
+    permissions:
+      checks: write
     needs: eslint_check_upload
     runs-on: ubuntu-latest
     name: ESLint Annotation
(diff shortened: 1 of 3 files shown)


Juice10:
Thanks for sharing this @AdnaneKhan and putting in the time to find this exploit, and find the recent PR. We have been using `cache: yarn` for a while now on some of the other GitHub actions in the repo, my recent change was to add cache in the spots we were missing. I'm guessing those caches also need removing/having their permissions reduced. Would you agree?

AdnaneKhan:
> Thanks for sharing this @AdnaneKhan and putting in the time to find this exploit, and find the recent PR. We have been using `cache: yarn` for a while now on some of the other GitHub actions in the repo, my recent change was to add cache in the spots we were missing. I'm guessing those caches also need removing/having their permissions reduced. Would you agree?

That’s your decision - so the root cause (that allows an exploit) here is the workflow that runs code on pull_request_target from fork PRs combined with the release workflow consuming the cache. Ideally, you should look at overhauling the CI so that you use the pull_request trigger (not pull_request_target) to run the untrusted code, then use a second workflow on workflow_run to upload the results/format code, etc. This pattern is called out in https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/. Thats a more involved change, though.

The changes in _this_ PR protect your release secrets from immediate risk. Even if someone did try a cache poisoning attack they would not be able to cause real harm (such as stealing the NPM secret - which we can agree would lead to a big miss if rrweb was backdoored).

I can work on contributing a follow-up PR to design a more robust release process for the project once this one is implemented. Better to remove the immediate risk first, though.

.github/workflows/style-check.yml
```diff
@@ -5,6 +5,7 @@ on: [push, pull_request_target]
 jobs:
   eslint_check_upload:
     runs-on: ubuntu-latest
+    permissions: read-all
```
Juice10, line 8:
```suggestion
    permissions:
      contents: read
```

.github/workflows/style-check.yml
```diff
@@ -52,6 +55,7 @@ jobs:
   prettier_check:
     # In the forked PR, it's hard to format code and push to the branch directly, so the action only check the format correctness.
     if: github.event_name != 'push' && github.event.pull_request.head.repo.full_name != 'rrweb-io/rrweb'
+    permissions: read-all
```
Juice10, line 58:
```suggestion
    permissions:
      contents: read
```

Juice10 (changes requested):
Thanks for all these suggested changes, I went through them to try to understand and double check everything. Looks really good, I thought it would make sense to go as strict as we can since we're looking at it right now so I suggested some stricter permissions

Juice10 (approved):

PR #1652, opened 2025-02-11
Improve mutation processing performance  

eoghanmurray:
Refactor record time mutation processing to improve performance
 
- smarter 'iteration' through the `this.addedSet` so that we start with nodes which already have the required `nextId` and `parentId`
- take advantage of siblings sharing the same parent, so only do parent related calculations once instead of repeatedly for every child node
- inline pushAdd (and getNextId) as we only need a single run over them

This solves pathological cases where nodes from the addedSet were pushed onto the secondary addList, possibly multiple times as `pushAdd` was called again each time the nextId/parentId requirements weren't met.

Previous efforts in this direction were 
- #1398 "traverse children in reverse order"
- #1302 "optimize random shuffled addList"

Also related is #1277 "refactor recursive procedure to iterative" which we could also incorporate on top of this as it should provide orthogonal performance gains — in order to 'grasp the nettle' all at once in terms of possible breakage.


eoghanmurray:
I've cloned the plunker from #1302 from @mdellanoce 
https://plnkr.co/edit/QgOXGFAFOguN2Zpa

The microtasks run after the jstree work is done, and I'm clocking 3.2s for those before this PR vs. 2.2s after.

eoghanmurray:
Here's the benchmark output (please replicate on other machines):

```			
                                                                                                      
title                                                   before     after   improvement   html                                             
                                                                                                      
1000x 1 DOM nodes with deeply nested children       10  2107.5    1974.1      7.0%       benchmark-dom-mutation-deep-nested.html        
1000x10 DOM nodes                                   10   384.9     370.7      3.7%       benchmark-dom-mutation.html                    
1000x10x2 DOM nodes and remove a bunch of them      10   526.6     331.9     36.0%       benchmark-dom-mutation-add-and-remove.html     
1000 DOM nodes and append into its previous looped   5   132.2      79.6     39.8%       benchmark-dom-mutation-multiple-descendant-add.h
10000 DOM nodes and move it to new container         5   586       301       48.6%       benchmark-dom-mutation-add-and-move.html       
modify attributes on 10000 DOM nodes                10    96.7      87.5      9.5%       benchmark-dom-mutation-attributes.html
```

packages/rrweb/src/replay/index.ts
```diff
@@ -1506,7 +1506,7 @@ export class Replayer {
           // is newly added document, maybe the document node of an iframe
           return this.newDocumentQueue.push(mutation);
         }
-        return queue.push(mutation);
+        return legacy_queue.push(mutation);
```
pauldambra, line 1518: fly by that the other item marked legacy here only made me wonder _why_ it was legacy

so there's a signal that maybe it's not current 
but I don't know what the consequences are
e.g. is the code only going to run on data recorded prior to version X

i don't know where the best place for that info is
but i worry it's in someone's head and could be lost :)

packages/rrweb/src/record/mutation.ts
```diff
@@ -272,45 +172,137 @@ export default class MutationBuffer {
...
+        nextSibling = n; // n is a good next sibling
+        n = n.previousSibling as Node;
+      } else {
+        n = this.addedSet.values().next().value as Node; // pop
+
+        // eslint-disable-next-line no-constant-condition
+        while (true) {
+          parentNode = dom.parentNode(n);
+          if (this.addedSet.has(parentNode as Node)) {
+            // start at top of added tree so as not to serialize children before their parents (parentId requirement)
```
pauldambra, line 216:
```suggestion
            // keep searching for the top of added tree so as not to serialize children before their parents (parentId requirement)
```

nit... "start at the top" threw me here and made me think 🙈

packages/rrweb/src/record/mutation.ts
```diff
@@ -272,45 +172,137 @@ export default class MutationBuffer {
...
+              isSelfOrAncestorInSet(this.droppedSet, parentNode) ||
+              this.removesSubTreeCache.has(parentNode);
+
+            if (ancestorBad && isSelfOrAncestorInSet(this.movedSet, n)) {
+              // not bad, just moved
+              ancestorBad = false;
+            }
+          }
+
+          if (this.addedSet.has(parentNode.lastChild as Node)) {
```
pauldambra: i guess the casts to Node are to avoid `!!parentNode.lastChild && x.has(parentNode.lastChild)`

i think safe here but `as X` is such a source of bugs that it always scares me 🙈

pauldambra (commented):
ugh, my youngest is awake...

I promised i'd do a review but i've not looked at mutation code much before so this first pass is much more to check my understanding than anything else

i have a customer site that i know is causing us problems with missing nodes at playback, will try and figure out if there's a way i can test this code on that site to compare the output (i don't think i can without customer changing their code)

that site is a stock ticker so maybe i can make a minimal example myself quickly

(maybe it goes without saying but feel free to tell me where i'm being a fool, i'm keen to understand this all more)

pauldambra:
i tested this locally today

i'd made a stock ticker website that allows me to set a high mutation rate on elements to try and replicate the performance impact of recording that customers are sometimes reporting to use with high complexity sites (although I'm struggling to replicate!)

the performance was not obviously better although may have used less RAM but recorded correctly and was at least not worse

packages/rrweb/src/record/mutation.ts
```diff
@@ -272,45 +172,137 @@ export default class MutationBuffer {
...
           // genTextAreaValueMutation already called via parent
-          return;
-        } else if (parentTag === 'STYLE' && this.addedSet.has(parent)) {
+          continue;
+        } else if (parentTag === 'STYLE' && addedIds.has(parentId)) {
           // css content will be recorded via parent's _cssText attribute when
           // mutation adds entire <style> element
           cssCaptured = true;
         }
       }
```
pauldambra, line 297: ah, posthog is still based on 2.0 alpha 17 (since the perf issues in ~Jan of this year) so we don't have this `cssCaptured` 

is this essential here (i'm travelling and my brain is slow 🫠)

packages/rrweb/src/record/mutation.ts
eoghanmurray: Ya, typescript sees `new Set<Node>()`, so wants to only allow Nodes in the `has` function, whereas .lastChild can return `null`.  However the `.has(null)` on a Javascript Set will return `false` which is fine.  We don't want `null`s in the set, so preventing `.add(null)` makes sense, however the error on the `.has` is a bit too strict for my taste.

packages/rrweb/src/record/mutation.ts
eoghanmurray: I think these bits (Lines 277-289 in rhs) are just routine updates based on variable name change, and change to `continue` as we're no longer in a function

packages/rrweb/src/replay/index.ts
```diff
@@ -1474,7 +1474,7 @@ export class Replayer {
     const legacy_missingNodeMap: missingNodeMap = {
       ...this.legacy_missingNodeRetryMap,
     };
-    const queue: addedNodeMutation[] = [];
+    const legacy_queue: addedNodeMutation[] = [];
```
eoghanmurray, line 1486:
```suggestion
    // this structure should not now normally be needed after #1652
    // but is definitely needed for recordings produced prior to that
    const legacy_queue: addedNodeMutation[] = [];
```

packages/rrweb/src/replay/index.ts
eoghanmurray: Thanks, I've updated with a comment above the `const legacy_queue`. Feel free to amend that.

.changeset/refactor-mutation-pushadd.md
```diff
@@ -0,0 +1,5 @@
+---
+"rrweb": patch
+---
+
+Improvements in efficiency of mutation handling; has been identified as a problem numerous times by mdellanoce, JonasBa and others
```
Juice10:
```suggestion
Improvements in efficiency of mutation handling; has been identified as a problem numerous times by @mdellanoce, @JonasBa and others
```

packages/rrweb/test/utils.ts
```diff
@@ -318,7 +319,21 @@ export async function assertSnapshot(
...
+    const file = long_fname.split(' > ')[0].replace('.test.ts', '');
+    if (typeof useOwnFile !== 'string') {
+      useOwnFile = long_fname.substring(long_fname.indexOf(' > ') + 3);
+    }
+    useOwnFile = useOwnFile.replace(/ > /g, '.').replace(/\s/g, '_');
+
+    const fname = `./__${file}.snapshots__/${useOwnFile}.json`;
+    expect(stringifySnapshots(snapshots)).toMatchFileSnapshot(fname);
+  } else {
+    expect(stringifySnapshots(snapshots)).toMatchSnapshot();
```
Juice10, line 335:
```suggestion
    const longFname = expect.getState().currentTestName.split('/').pop();
    const file = longFname.split(' > ')[0].replace('.test.ts', '');
    if (typeof useOwnFile !== 'string') {
      useOwnFile = longFname.substring(longFname.indexOf(' > ') + 3);
    }
    useOwnFile = useOwnFile.replace(/ > /g, '.').replace(/\s/g, '-');

    const fname = `./__${file}.snapshots__/${useOwnFile}.json`;
    expect(stringifySnapshots(snapshots)).toMatchFileSnapshot(fname);
  } else {
    expect(stringifySnapshots(snapshots)).toMatchSnapshot();
```

packages/rrweb/test/record/mutation.test.ts
```diff
@@ -0,0 +1,395 @@
...
+
+  describe('Large list 1 perf', () => {
+    const mutationBuffer = new MutationBuffer();
+    const options: MutationBufferParam = {
+    };
+    mutationBuffer.init(options);
+
+    mutationBuffer.processMutations([]);
+  });
+*/
```
Juice10, line 18:
```suggestion
```

packages/rrweb/test/integration.test.ts
```diff
@@ -1071,18 +1071,22 @@ describe('record integration tests', function (this: ISuite) {
   it('should record moved shadow DOM', async () => {
     const page: puppeteer.Page = await browser.newPage();
     await page.goto('about:blank');
+    page.on('console', (msg) => console.log(msg.text()));
```
Juice10, line 1074:
```suggestion
    // page.on('console', (msg) => console.log(msg.text()));
```

packages/rrweb/src/record/mutation.ts
```diff
@@ -272,45 +172,137 @@ export default class MutationBuffer {
...
-        ns = ns && ns.nextSibling;
-        nextId = ns && this.mirror.getId(ns);
+    while (this.mapRemoves.length) {
+      this.mirror.removeNodeFromMap(this.mapRemoves.shift()!);
+    }
+
+    for (const n of this.movedSet) {
+      const parentNode = dom.parentNode(n);
+      if (
+        parentNode && // can't be removed if it doesn't exist
```
Juice10, line 182:
```suggestion
        parentNode && // can't be moved if it doesn't exist
```

Is this what you meant?

packages/rrweb/src/record/mutation.ts
```diff
@@ -272,45 +172,137 @@ export default class MutationBuffer {
...
+      ) {
+        continue;
       }
-      return nextId;
-    };
-    const pushAdd = (n: Node) => {
-      const parent = dom.parentNode(n);
-      if (!parent || !inDom(n)) {
-        return;
+      this.addedSet.add(n);
```
Juice10: So if it was moved, and its new parent wasn't removed, then it was "added", eg. we add it to `addedSet`?

Juice10:
Newest version looks good and performs well.

| Workload | Main | This PR |
| -- | -- | -- | 
Deep nesting | 157.6 | 141.9 | 
1000×10 nodes | 33.5 | 30.8 |
Add/remove | 47.6 | 44.3 | 
Chained appends | 11.1 | 9.3 |
Move nodes | 67.2 | 46.0 | 
Attributes | 10.2 | 10.4 | 

There is still some minor cleanup (I added some inline code suggestions), but apart from that this is ready for merge IMO

eoghanmurray:
Enabled benchmarking of Firefox/WebKit via Playwright:

  Firefox (median ms) — pre-PR 2.1.4 → final PR 2.1.5


| Case | pre | final PR |Δ |
| -- | -- | -- |  -- | 
| deep-nested        | 906 |      724 | +20.1% |
| 1000×10            | 216 |      186 | +13.8% |
| add-and-remove     | 294 |      264 | +10.1% |
| append-into-previous |  83 |       77 |  +7.4% |
| add-and-move       | 304 |      235 | +22.7% |
| attributes         |  53 |       61 |   −14% |

 WebKit (median ms)

|         Case       | pre | final PR |   Δ  |
| -- | -- | -- |  -- |
| deep-nested        | 941 |      834 | +11.4% |
| 1000×10            | 148 |      145 |  +1.9% |
| add-and-remove     | 221 |      223 |  −1.2% |
| append-into-previous |  90 |       86 |  +4.0% |
| add-and-move       | 240 |      195 | +18.7% |
| attributes         |  40 |       37 |  +5.3% |

packages/rrweb/src/record/mutation.ts
eoghanmurray: the 'can't be removed' was actually referring to the next line `removesSubTreeCache.has`
the guard is actually redundant and was probably added to document what happens if parentNode is null but it's clearly adding confusion, so will remov

.changeset/refactor-mutation-pushadd.md
```diff
@@ -0,0 +1,5 @@
+---
+"rrweb": patch
```
eoghanmurray:
```suggestion
"rrweb": patch
"@rrweb/record": patch
```

Juice10 (approved):

PR #1543, opened 2024-07-20
ref: isParentRemoved to cache subtree

JonasBa:
isParentRemoved is called super often for each node that we add. I think the current logic results in an On^2 time complexity as each visit to a node traverser the entire chain of parents (please correct me if I'm wrong).

The motivation behind this change is to iterate on the graph once ahead of time which can be done in On time and store the parent nodes which were removed. That way we can query the removed set in O1 time and only perform this lookup once per node

Related to https://github.com/rrweb-io/rrweb/pull/1271

PR #1653, opened 2025-02-13
Performance refactor to avoid 2 function calls

eoghanmurray:
A bridging PR between #1543 and #1652


eoghanmurray:
@JonasBa I think you and @Juice10 worked on #1543, so would like your feedback as to whether this makes sense

PR #1676, opened 2025-04-08
Feat/hanlde frames

salomita:


packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -325,7 +325,14 @@ function buildNode(
               'rrweb-original-srcset',
               n.attributes.srcset as string,
             );
-          } else {
+          }
+          // Set the sandbox attribute on the iframe element will make it lose its contentDocument access and therefore cause additional playback errors.
```
hannadrehman, line 329: could this be related to this https://github.com/rrweb-io/rrweb/issues/1720 ?

PR #1687, opened 2025-04-25
Change `.umd.cjs` dist files to `.umd.js` to satisfy jsdelivr

eoghanmurray:
The verdict from jsdelivr is that they are not going to change the `application/node` mime type for  (see https://github.com/jsdelivr/jsdelivr/issues/18584 ), so we need to revert the extension to `.umd.js` in order that these files can be used in a browser via jsdelivr as per docs (prior to #1033 we shipped these as `.js`)


eoghanmurray:
#1704 is probably better than this given the 'Dual Package' hazard
https://nodejs.org/en/learn/modules/publishing-a-package#the-dual-package-hazard

PR #1688, opened 2025-04-25
fix: prevent invalid media processing

juliecheng:
👋 hello! i'm seeing `h.pause is not a function` errors/warnings during playback and after looking into the issue some more, it looks like during playback we can end up adding non-media elements to the `mediaMap` via `mediaMutation`, which is what's causing some errors during playback. 

to summarize the changes in this pr:
1. protect against MediaInteraction events (mutations) being applied to non-media elements otherwise we'll see errors during playback
2. exit early out of `addMediaElements` if a media element is blocked since it won't have any media attributes anyway

### DIFF
--- a/.changeset/tiny-rabbits-chew.md
+++ b/.changeset/tiny-rabbits-chew.md
@@ -1,0 +1,5 @@
+---
+"rrweb": patch
+---
+
+Make sure MediaInteraction events / mutations are only processed on MediaElements + exit early in addMediaElements if a MediaElement is a blocked element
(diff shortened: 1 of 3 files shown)


juliecheng:
@eoghanmurray hi eoghan! just a heads up that my pr looks like it might be trying to resolve the same (or at least similar) issue as your pr here https://github.com/rrweb-io/rrweb/pull/1673 if my changes here look good i'm happy to close this one out and merge the changes into your pr instead if that's preferred (or vice versa). thanks!

PR #1694, opened 2025-05-27
Implement configurable throttling on mutation emission

eoghanmurray:
using the new `sampling.mutation` ms setting

 - can prevent overly frequent (javascript based) animation per second on the same attribute; the mutation buffer will emit a single change at the end if throttled changes are against the same element


eoghanmurray:
#1933 supersedes

PR #1695, opened 2025-05-27
fix: rrweb recorder may throw error when stopping recording after an iframe becomes cross-origin

YunFeng0817:
The CI failure can be solved by #1696


eoghanmurray:
I've checked and the MutationObserver can detect external changes like `iframeEl.src = 'xyz.com';` 
However what I think you are referring to is if the iframe itself changes the frame location using internal `document.location = 'redirect.com'` ?
Point taken in any case

packages/rrweb/src/record/index.ts
```diff
@@ -617,7 +617,25 @@ function record<T = eventWithTime>(
...
+      handlers.forEach((handler) => {
+        try {
+          handler();
+        } catch (error) {
+          const msg = String(error).toLowerCase();
+          /**
+           * https://github.com/rrweb-io/rrweb/pull/1695
+           * This error can occur in a known scenario:
+           * If an iframe is initially same-origin and observed, but later its 
+           src attribute is changed to a cross-origin URL,
```
eoghanmurray:
```suggestion
           location is changed in an opaque way to a cross-origin URL (perhaps within the iframe via its `document.location` or a redirect) 
```

packages/rrweb/test/record.test.ts
```diff
@@ -990,6 +990,21 @@ describe('record', function (this: ISuite) {
...
+  it('does not throw error when stopping recording after iframe becomes cross-origin', async () => {
+    await ctx.page.evaluate(async () => {
+      const { record } = (window as unknown as IWindow).rrweb;
+      const stopRecord = record({
+        emit: (window as unknown as IWindow).emit,
+      });
+      const iframe = document.createElement('iframe');
+      document.body.appendChild(iframe);
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+      iframe.src = 'https://www.example.com'; // Change the same origin iframe to a cross origin iframe after it's recorded
```
eoghanmurray: this should generate a mutation I believe

PR #1700, opened 2025-06-05
fix: use document.baseURI for sheetHref when handling inline styles

megboehlert:
Prioritize using the href from base tag (if present) when stringifying urls in inline stylesheets. Document.baseURI does just this. If a base tag is not present, it falls back to location.href.

### DIFF
--- a/packages/rrweb-snapshot/src/utils.ts
+++ b/packages/rrweb-snapshot/src/utils.ts
@@ -118,9 +118,9 @@ export function stringifyStylesheet(s: CSSStyleSheet): string | null {
       return null;
     }
     let sheetHref = s.href;
-    if (!sheetHref && s.ownerNode && s.ownerNode.ownerDocument) {
+    if (!sheetHref && s.ownerNode) {
       // an inline <style> element
-      sheetHref = s.ownerNode.ownerDocument.location.href;
+      sheetHref = s.ownerNode.baseURI;
     }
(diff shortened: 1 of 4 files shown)


packages/rrweb-snapshot/src/utils.ts
```diff
@@ -120,7 +120,7 @@ export function stringifyStylesheet(s: CSSStyleSheet): string | null {
     let sheetHref = s.href;
     if (!sheetHref && s.ownerNode && s.ownerNode.ownerDocument) {
       // an inline <style> element
-      sheetHref = s.ownerNode.ownerDocument.location.href;
+      sheetHref = s.ownerNode.ownerDocument.baseURI;
```
eoghanmurray:
```suggestion
      sheetHref = s.ownerNode.baseURI;
```

I don't think the ownerDocument bit is necessary anymore!

eoghanmurray (dismissed):

packages/rrweb-snapshot/src/utils.ts
megboehlert: Good call out! I opened a new [PR](https://github.com/rrweb-io/rrweb/pull/1705) with this change.

megboehlert:
new PR with requested changes [here](https://github.com/rrweb-io/rrweb/pull/1705)

PR #1704, opened 2025-06-12
Umd folder

eoghanmurray:
Create a new `umd` folder alongside `dist` for output of UMD files with a plain `.js` instead of `.cjs` extension, as the latter won't be served with the correct mime type by jsdelivr
    
 - #1687 (just rename `.cjs` to `.js`) was rejected due to the the [dual package hazard](https://nodejs.org/en/learn/modules/publishing-a-package#the-dual-package-hazard), and produces a warning when run through publint.dev (which was the original motivation for changing to \.cjs)
 - jsdelivr won't be serving `.cjs` with the correct mime type: https://github.com/jsdelivr/jsdelivr/issues/18584


packages/rrweb/src/record/index.ts
```diff
@@ -209,7 +209,7 @@ function record<T = eventWithTime>(
       e.type !== EventType.FullSnapshot &&
       !(
         e.type === EventType.IncrementalSnapshot &&
-        e.data.source === IncrementalSource.Mutation
+        e.data.source ===IncrementalSource.Mutation
```
Juice10:
```suggestion
        e.data.source === IncrementalSource.Mutation
```

vite.config.default.ts
```diff
@@ -51,22 +51,36 @@ function minifyAndUMDPlugin({
...
             await buildFile({
               name,
               input: inputFilePath,
-              output: `${outputFilePath}.umd.cjs`,
+              output: outUmd,
               minify: false,
               isCss: false,
               outDir,
             });
+            copyFileSync(
```
Juice10, line 69:
```suggestion
            // Workaround because jsdelivr does use correct mime types for .umd.cjs
            // More info: https://github.com/jsdelivr/jsdelivr/issues/18584 https://github.com/rrweb-io/rrweb/pull/1704 
            copyFileSync(
```

Juice10:
I just realized this is missing changes to package.json file which include the umd files in npm packages (otherwise this won't be saved by jsdelivr:
We currently have this in `package.json`:
```
"files": [
    "dist",
    "package.json"
  ],
```
it needs to be this instead:
```
"files": [
    "dist",
    "umd",
    "package.json"
  ],
```

Issue #1707, opened 2025-06-20
[Bug]: Memory leak and high CPU usage for link with ref preload


maksimr:
### Preflight Checklist

- [x] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

2.0.0-alpha.18

### Expected Behavior

Prevent memory leak.

1. First of all we should cleanup `unload` listener when try to make snapshot again
2. It's unclear why we should at all try to listen `preload` stylesheet doesn't contains sheet at all and only preload it so maybe we should just drop this check or upgrade it to `rel="preload stylesheet"` which preload an apply styles.  

### Actual Behavior

Memory Leak:
`<link rel="preload" href="./style.css" as="style">` always returns `link.sheet` equals to `null`. So we repeatedly [call `setTimeout` and `addListener`](https://github.com/rrweb-io/rrweb/blob/4db9782d1278a2b7235ed48162ccedf0e0952113/packages/rrweb-snapshot/src/snapshot.ts#L382-L386) for such link and don't cleanup added event listener that lead to memory leak.

Hight CPU usage:
The previous case with memory leak lead to degradation of performance if repeatedly stop and start recording. For example we stop recording when page invisible and start again when visible. In this case leaked timer hold previous record and now we have N(number of restarts) timers which constantly call `setTimeout` and add listeners so every `stylesheetLoadTimeout` we call N timers which add N listeners. In worst case the page is freez and consume 100% of CPU because we have a lot of timers and a lot of listeners.

### Steps to Reproduce

index.html
```html
<html>

  <link rel="preload" href="./style.css" as="style">

  <body>
    <script type="module" src="./index.js"></script>
  </body>

</html>
```

index.js
```js
// Just Memory Leak
import { record } from "rrweb";

record({
  emit: () => void 0,
  checkoutEveryNms: 5000,
});

// Run getEventListeners every 5 seconds in DevTools
// getEventListeners(document.querySelector('link'));
```

<img width="1438" alt="Image" src="https://github.com/user-attachments/assets/c4fdf0e2-2ea0-4c26-94d0-ede7abe8d7a8" />

index.js
```js
// Memory leak + high CPU usage
import { record } from "rrweb";

async function main() {
  const stopRef = { current: null };

  const pageVisibilityChangeNTimes = 5000;
[...]

PR #1708, opened 2025-06-23
fix: memory leaks (#1707)

maksimr:


packages/rrweb/src/record/index.ts
```diff
@@ -375,6 +377,8 @@ function record<T = eventWithTime>(
     shadowDomManager.init();
 
     mutationBuffers.forEach((buf) => buf.lock()); // don't allow any mirror modifications during snapshotting
+    abortController.abort();
```
pauldambra, line 380: question from a place of ignorance, sorry...

i'm reading this as "if we start a new snapshot, abort the previous one". do we know that the recorder does not rely on the previous one finishing?

PR #1709, opened 2025-06-25
Optimized the user guide

Juntao666:
When I was following this user guide to use rrweb, which is an amazing tool, I found the installation process confusing, especially for developers with little experience like me. So, I added more descriptions for the different installation options. Also, some of the links to newer versions no longer work, so I included a link to an older version that is still supported in the user guide.


eoghanmurray:
We'll have to apologize on this one as the user guide will need to be updated to point to the right places when/if https://github.com/rrweb-io/rrweb/pull/1704 gets merged

PR #1712, opened 2025-07-01
fix: move a console log to the warn method so it can be suppressed 

pauldambra:
noticed this while doing something else
this call to `console.warn` is direct so it doesn't obey the provided config
let's make it play nicely

### DIFF
--- a/.changeset/tiny-singers-pump.md
+++ b/.changeset/tiny-singers-pump.md
@@ -1,0 +1,6 @@
+---
+"rrweb-snapshot": patch
+"rrweb": patch
+---
+
+Use the provided logger from player config instead of directly calling console.warn"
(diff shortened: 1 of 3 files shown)


.changeset/tiny-singers-pump.md
```diff
@@ -0,0 +1,6 @@
+---
+"rrweb-snapshot": patch
+"rrweb": patch
+---
+
+Use the provided logger from player config instead of directly calling console.warn"
```
Juice10:
```suggestion
Use the provided logger from player config instead of directly calling console.warn
```

Juice10 (approved):

eoghanmurray (approved):

PR #1718, opened 2025-07-22
fix (rrdom): Fix edge case where oldChild can become detach causing the recursive function to end early

arivyiii:
Fix an edge case in diffChildren: when we replace an old child via [oldTree.parentNode?.replaceChild(calibratedOldTree, oldTree);](https://github.com/rrweb-io/rrweb/blob/4db9782d1278a2b7235ed48162ccedf0e0952113/packages/rrdom/src/diff.ts#L145C1-L145C66) its nextSibling pointer is lost, causing the diff loop to exit early and inadvertently remove subsequent elements and styles.

### DIFF
--- a/packages/rrdom/src/diff.ts
+++ b/packages/rrdom/src/diff.ts
@@ -525,6 +525,24 @@ function diffChildren(
   let oldChild = oldTree.firstChild;
   let newChild = newTree.firstChild;
   while (oldChild !== null && newChild !== null) {
+    // For mismatched node types (e.g., Element vs Text), we create a new node and insert it before the old one.
+    // This preserves the oldChild reference needed for the recursive diff loop, which would break if we replaced the node directly.
+    const isMismatch = !sameNodeType(oldChild, newChild);
+
+    if (isMismatch) {
+      const newNode = createOrGetNode(newChild, replayer.mirror, rrnodeMirror);
+
+      try {
+        oldTree.insertBefore(newNode, oldChild);
+      } catch (e) {
+        console.warn(e);
+      }
+
+      diff(newNode, newChild, replayer, rrnodeMirror);
+      newChild = newChild.nextSibling;
+      continue;
+    }
+
     diff(oldChild, newChild, replayer, rrnodeMirror);
     oldChild = oldChild.nextSibling;
     newChild = newChild.nextSibling;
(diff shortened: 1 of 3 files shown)


packages/rrdom/src/diff.ts
```diff
@@ -525,7 +525,26 @@ function diffChildren(
...
   while (oldChild !== null && newChild !== null) {
+    // For mismatched node types (e.g., Element vs Text), we create a new node and insert it before the old one.
+    // This preserves the oldChild reference needed for the recursive diff loop, which would break if we replaced the node directly.
+    const isMismatch = !sameNodeType(oldChild, newChild);
+
+    if (isMismatch) {
+      const newNode = createOrGetNode(newChild, replayer.mirror, rrnodeMirror);
+
+      try {
+        handleInsertBefore(oldTree, newNode, oldChild);
```
pauldambra: may be a silly question... where is `handleInsertBefore` defined?
i don't find it in rrweb at all

eoghanmurray: Yes, maybe this PR is incomplete? @arivyiii is that a mispelling or something that was left out?

arivyiii: Sorry I missed your messages. This PR was originally using handleInsertBefore (a safe wrapper around oldTree.insertBefore from another commit), but since oldTree.insertBefore works fine for this use case, I've updated the PR accordingly.

eoghanmurray:
Your test case is very helpful.  
I'm assuming the `newList.splice(1, 0, { tagName: 'STYLE', id: 1 });` line is necessary to replicate the bug.
It is not clear to me how that scenario would arise 'in the wild'?  Would it be possible to explain the motivating scenario a little more?

arivyiii:
> It is not clear to me how that scenario would arise 'in the wild'? Would it be possible to explain the motivating scenario a little more?

Yeah, it was a strange case that occurred in a few of our replays when users were using this extension: https://chromewebstore.google.com/detail/dark-reader/eimadpbcbfnmbkopoojfekhnkhdbieeh?hl=en-US. For some reason, the style tags were being appended with the same node ID back to back. I wasn’t able to reproduce it on any site with the recorder, but while debugging the replay I noticed that the childNode had a nextSibling, which was lost once it returned to diffChildren. We felt it was a safe enough change since for most replays they will never enter this block of code and if this edge case occur it would get resolved versus our users having to deal with a distorted replay.

PR #1719, opened 2025-07-28
Create SECURITY.md

rotemreiss:
At the moment, there's no documentation on how to report security vulnerabilities to rrweb-io on the rrewb project.
Please review my suggestion, make edits if needed, and provide security researchers a responsible way to report security vulnerabilities in this widely used project.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,10 @@
+# Vulnerability Disclosure Policy
+This document outlines rrweb's vulnerability disclosure policy.
+
+## Reporting a Vulnerability
+Please do not report security vulnerabilities through public GitHub issues.
+Instead, please report them to our GitHub Security page. If you prefer to submit one without using GitHub, you can also email us at security@rrweb.io.
+We commit to acknowledging vulnerability reports immediately and will work to fix active vulnerabilities as soon as we can. We will publish resolved vulnerabilities as security advisories on our GitHub security page.
+
+We appreciate your help in making rrweb more secure for everyone.
+Thank you for your support and responsible disclosure.


SECURITY.md
```diff
@@ -0,0 +1,12 @@
+# Vulnerability Disclosure Policy
+
+This document outlines rrweb's vulnerability disclosure policy.
+
+## Reporting a Vulnerability
+
+Please do not report security vulnerabilities through public GitHub issues.
+Instead, please report them to our GitHub Security page. If you prefer to submit one without using GitHub, you can also email us at security@rrweb.io.
```
pauldambra: @Juice10 would that email route anywhere?
i'd guess it's better not to have an email route published since then someone needs to monitor that inbox

SECURITY.md
```diff
@@ -0,0 +1,12 @@
...
+We commit to acknowledging vulnerability reports immediately and will work to fix active vulnerabilities as soon as we can. We will publish resolved vulnerabilities as security advisories on our GitHub security page.
```
pauldambra:
```suggestion
We commit to acknowledging vulnerability reports and will work to fix active vulnerabilities as soon as we can (noting this is a community run project).

We will publish resolved vulnerabilities as security advisories on our GitHub security page.
```

pauldambra (approved):
email route worries me that it becomes an unmonitored black box 
and i'm not keen on saying "immediately" when the maintainers might be volunteering their time
but otherwise, this helps with clarity

SECURITY.md
eoghanmurray: we've set up a new private google group for this which will also email members of the core team directly

eoghanmurray: ( @YunFeng0817  has edited this PR from security@rrweb.io to rrweb-security@googlegroups.com )

eoghanmurray (approved):

SECURITY.md
rotemreiss: @pauldambra, @eoghanmurray - Thanks for merging it!
Could you also consider enabling the vulnerability report feature to allow private vulnerability reports directly via GitHub?

This is where you can find this config:
image

eoghanmurray: This has been enabled already I believe although I didn't do that.

rotemreiss: It's not. You do have the SECURITY.md file now, but there's no option to report vulnerabilities (create GHSA) via the security section.
You should have a button like this:
image

PR #1721, opened 2025-08-02
feat: adopt perf improvements from https://github.com/rrweb-io/rrweb/pull/1652

pauldambra:
adopts the recording performance improvements


pauldambra:
ugh, wrong repo :/

PR #1733, opened 2025-09-06
fix(rrweb-snapshot): absolutifyURLs function calculation error in some scenario

godky:
`absolutifyURLs` function calculation error when processing inline style sheets in hash mode spa. For example:

```javascript
//output: url(http://www.example.com/somepath/#/a.jpg)
absolutifyURLs('url("./a.jpg")', 'http://www.example.com/somepath/#/route')
```

But the expected result is  `url(http://www.example.com/somepath/a.jpg)`


packages/rrweb-snapshot/src/utils.ts
```diff
@@ -429,19 +429,10 @@ export function absolutifyURLs(cssText: string | null, href: string): string {
...
-        if (part === '.') {
-          continue;
-        } else if (part === '..') {
-          stack.pop();
-        } else {
-          stack.push(part);
-        }
-      }
-      return `url(${maybeQuote}${stack.join('/')}${maybeQuote})`;
+      return `url(${maybeQuote}${new URL(
```
eoghanmurray: We need to always call `new URL` within a try/catch ... we could maybe fallback to the old method

eoghanmurray (changes requested):

PR #1735, opened 2025-09-08
fix: css syntax error: 'Unclosed string'

### LINKED ISSUE #1734: [Bug]: Empty css rules break replay with "CssSyntaxError CssSyntaxError: Unclosed string" error
### Preflight Checklist

rrweb

### Version

latest

### Expected Behavior

"Record" produces valid css rules without empty ones, which are invalid and can't be parsed by "replay"

### Actual Behavior

Because "cssText" is invalid css, "replay" fails with:
```
CssSyntaxError
CssSyntaxError: Unclosed string
    at Input.error (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:1103:17)
    at unclosed (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:2041:18)
    at Object.nextToken (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:2125:15)
    at Parser.other (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:2708:30)
    at Parser.parse (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:2749:16)
    at parse$3$1 (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:2878:13)
    at new LazyResult (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:3010:17)
    at Processor.process (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:3513:14)
    at adaptCssForReplay (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:3733:6)
    at applyCssSplits (webpack-internal:///../../node_modules/@rrweb/replay/dist/replay.js:3759:45)
```

### Steps to Reproduce

1) Open any web page

2) Execute following script:

```js
const stylesElement = document.createElement("style");
stylesElement.innerHTML = ".cl {border: 1px solid var(--border-color); border-color: var(--border-color);}";
document.head.appendChild(stylesElement);
console.log(stylesElement.sheet.rules[0].style.cssText);
```

3) "cssText" would output:
```
border-top-style: ; border-top-width: ; border-right-style: ; border-right-width: ; border-bottom-style: ; border-bottom-width: ; border-left-style: ; border-left-width: ; border-image-source: ; border-image-slice: ; border-image-width: ; border-image-outset: ; border-image-repeat: ; border-color: var(--border-color);
```

4) Later it would be saved that way into rrweb snapshot, and, when replaying with "replay", it would fail with "CssSyntaxError: Unclosed string" because of empty css rules.

### Testcase Gist URL

_No response_

### Additional Information

_No response_

KuznetsovRoman:
Fixes https://github.com/rrweb-io/rrweb/issues/1734

### DIFF
--- a/packages/rrweb-snapshot/src/utils.ts
+++ b/packages/rrweb-snapshot/src/utils.ts
@@ -150,7 +150,9 @@ export function stringifyRule(rule: CSSRule, sheetHref: string | null): string {
     }
     return importStringified;
   } else {
-    let ruleStringified = rule.cssText;
+    // Removes incomprehensible empty .cssText rules
+    // see https://github.com/rrweb-io/rrweb/issues/1734
+    let ruleStringified = fixEmptyCssStyles(rule.cssText);
     if (isCSSStyleRule(rule) && rule.selectorText.includes(':')) {
       // Safari does not escape selectors with : properly
       // see https://bugs.webkit.org/show_bug.cgi?id=184604
@@ -167,6 +169,18 @@ export function fixSafariColons(cssStringified: string): string {
   // Replace e.g. [aa:bb] with [aa\\:bb]
   const regex = /(\[(?:[\w-]+)[^\\])(:(?:[\w-]+)\])/gm;
   return cssStringified.replace(regex, '$1\\$2');
+}
+
+export function fixEmptyCssStyles(cssStringified: string) {
+  const quickTestRegExp = /:\s*;/;
+
+  if (quickTestRegExp.test(cssStringified)) {
+    // Remove e.g. "border-top-style: ;"
+    const regex = /(?<=^|\{|;)\s*[_a-zA-Z][_a-zA-Z0-9-]*:\s*;/gm;
+    return cssStringified.replace(regex, '');
+  }
+
+  return cssStringified;
 }
 
 export function isCSSImportRule(rule: CSSRule): rule is CSSImportRule {
--- a/.changeset/smooth-ghosts-wonder.md
+++ b/.changeset/smooth-ghosts-wonder.md
@@ -1,0 +1,5 @@
+---
+"@rrweb/record": patch
+---
+
+fix 'CssSyntaxError: Unclosed string' for invalid css rules
--- a/packages/rrweb-snapshot/test/utils.test.ts
+++ b/packages/rrweb-snapshot/test/utils.test.ts
@@ -6,6 +6,7 @@ import {
   escapeImportStatement,
   extractFileExtension,
   fixSafariColons,
+  fixEmptyCssStyles,
   isNodeMetaEqual,
   stringifyStylesheet,
 } from '../src/utils';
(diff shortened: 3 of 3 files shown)


packages/rrweb-snapshot/src/utils.ts
```diff
@@ -169,6 +172,12 @@ export function fixSafariColons(cssStringified: string): string {
   return cssStringified.replace(regex, '$1\\$2');
 }
 
+export function fixEmptyCssStyles(cssStringified: string): string {
+  // Remove e.g. "border-top-style: ;"
+  const regex = /(?<=^|\{|\;)\s*[_a-zA-Z][_a-zA-Z0-9-]*:\s*;/gm;
```
KuznetsovRoman, line 177: regex breakdown:
- This regex should have "new line" or "{" or ";" before itself (which is precedes css rule name) `(?<=^|\{|\;)`
- Regex might start with empty symbols (between "new line / { / ;" and actual css rule name) `\s*`
- Then it matches css rule name `[_a-zA-Z][_a-zA-Z0-9-]*`
- After css rule name, there should be ":" symbol (between css rule name and its value) `:`
- Then it might have some empty symbols `\s*`
- Then its essential to end css rule, without any characters, which might be interpreted as css value `;`

All occurrences of that regex are invalid css properties and have to be removed

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -169,6 +171,18 @@ export function fixSafariColons(cssStringified: string): string {
   return cssStringified.replace(regex, '$1\\$2');
 }
 
+export function fixEmptyCssStyles(cssStringified: string) {
+  const quickTestRegExp = /:\s*;/;
+
+  if (quickTestRegExp.test(cssStringified)) {
+    // Remove e.g. "border-top-style: ;"
+    const regex = /(?<=^|\{|\;)\s*[_a-zA-Z][_a-zA-Z0-9-]*:\s*;/gm;
```
KuznetsovRoman: regex breakdown:
- We should see "new line" or "{" or ";" before regex starts (which is precedes css rule name) `(?<=^|\{|\;)`
- Regex might start with empty symbols (between "new line / { / ;" and actual css rule name) `\s*`
- Then it matches css rule name `[_a-zA-Z][_a-zA-Z0-9-]*`
- After css rule name, there should be ":" symbol (between css rule name and its value) `:`
- Then it might have some empty symbols `\s*`
- Then its essential to end css rule, without any characters, which might be interpreted as css value `;`

All occurrences of that regex are invalid css properties and have to be removed

packages/rrweb-snapshot/src/utils.ts
```diff
@@ -169,6 +171,18 @@ export function fixSafariColons(cssStringified: string): string {
   return cssStringified.replace(regex, '$1\\$2');
 }
 
+export function fixEmptyCssStyles(cssStringified: string) {
+  const quickTestRegExp = /:\s*;/;
+
+  if (quickTestRegExp.test(cssStringified)) {
```
KuznetsovRoman, line 177: The bug is rare, so it worth to check "if original css rules even has pattern of `:\s*;`"

If it hasn't, then we don't need to execute regex.

This will save some CPU cycles

PR #1737, opened 2025-09-10
image currentSrc property can sometimes be undefined

QuentinLowe:
image currentSrc property can sometimes be undefined


eoghanmurray:
This PR could really do with a test to recreate exactly the scenario where `currentSrc` would be undefined; is it because of replay in a particular browser, or is there some recording that can reliably recreate the scenario in all browsers?  A test is more useful than a fix as inline images has been completely overhauled in #1475

QuentinLowe:
When the virtual DOM mode is triggered, this problem will arise.

Issue #1742, opened 2025-09-19
[Bug]: rrweb record minification size


eoghanmurray:
### Preflight Checklist

- [x] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

@2.0.0-alpha.18

### Expected Behavior

There's a large jump in record.js size between alpha.15 and alpha.18:

@2.0.0-alpha.15  -> 81532 bytes
@2.0.0-alpha.17  -> 205883 bytes
@2.0.0-alpha.18  -> 207283 bytes

as per this package:
https://www.jsdelivr.com/package/npm/@rrweb/record

I believe this may correspond to the addition of the postcss parser in #1458 
Which now is being included in the record output, even though it's only needed in replay.

### Actual Behavior

postcss packaged and included in record.js

### Steps to Reproduce

Download alpha.15 and alpha.18 and compare the files

PR #1743, opened 2025-09-19
Trying to expunge postcss from record build output

eoghanmurray:
Re #1742

See changes, trying to output record.js without postcss included

Thought that the `record` package would import from new `rrweb-record.js` output after these changes, but that doesn't appear to be happening.

I'm getting a some weird new files in the output, and these are the ones including postcss :
```
rrweb/packages/rrweb$ grep postcss -l dist/*
grep: dist/assets: Is a directory
dist/base64-arraybuffer.es5-D-OFpZ9a.cjs
dist/base64-arraybuffer.es5-D-OFpZ9a.cjs.map
dist/base64-arraybuffer.es5-D-OFpZ9a.umd.cjs
dist/base64-arraybuffer.es5-D-OFpZ9a.umd.cjs.map
dist/base64-arraybuffer.es5-D-OFpZ9a.umd.min.cjs
dist/base64-arraybuffer.es5-D-OFpZ9a.umd.min.cjs.map
dist/base64-arraybuffer.es5-DwKE5w4P.js
dist/base64-arraybuffer.es5-DwKE5w4P.js.map
```

### DIFF
--- a/packages/record/src/index.ts
+++ b/packages/record/src/index.ts
@@ -1,3 +1,3 @@
-import { record } from 'rrweb';
+import record from 'rrweb/record';
 
 export { record };
--- a/packages/rrweb/package.json
+++ b/packages/rrweb/package.json
@@ -34,7 +34,6 @@
   "main": "./dist/rrweb.umd.cjs",
   "module": "./dist/rrweb.js",
   "unpkg": "./dist/rrweb.umd.cjs",
-  "typings": "dist/rrweb.d.ts",
   "exports": {
     ".": {
       "import": {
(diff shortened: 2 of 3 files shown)


eoghanmurray:
Claude's analysis:

```
 Here's what I found. The problem is clear and has a straightforward root cause:                                                                                                       
                                         
  The Import Chain                                                                                                                                                                      
                      
  @rrweb/record/src/index.ts                                                                                                                                                            
    └─ imports { record, getWindowScroll, ... } from 'rrweb'                                                                                                                            
         └─ rrweb/src/index.ts (barrel export) imports BOTH:
              ├─ ./record  (what you need)
              └─ ./replay  (what you don't need)
                   └─ imports { adaptCssForReplay, buildNodeWithSN } from 'rrweb-snapshot'
                        └─ rrweb-snapshot/src/rebuild.ts
                             └─ import postcss from 'postcss'  ← HERE

  packages/record/src/index.ts:1-6 imports from 'rrweb' — the main barrel export. That barrel (packages/rrweb/src/index.ts:1-8) unconditionally imports the Replayer, SyncReplayer, and
  canvasMutation from ./replay, which pulls in rrweb-snapshot's rebuild.ts, which imports postcss. Tree-shaking can't eliminate it because the replay module likely has side effects the
   bundler can't prove are safe to remove.

  158 occurrences of "postcss" in each of the record dist files confirms it's fully inlined.

  The irony

  The rrweb package already has a record-only entry point at packages/rrweb/src/entries/record.ts that imports only ../record without touching replay at all — but nobody is using it.

  The fix

  The @rrweb/record package needs to stop importing from rrweb's barrel. It should either:

  1. Import record from rrweb's record-specific entry/module, and getWindowScroll/getWindowHeight/getWindowWidth directly from rrweb's utils — bypassing the barrel entirely
  2. Or restructure rrweb-snapshot's index.ts to not eagerly import rebuild.ts (which contains postcss)

  Option 1 is much simpler and more surgical. Want me to look into implementing it?

```

Issue #1682, opened 2025-04-16
rrweb cannot get the correct document object in the js sandbox


HappyPot:
I use a micro-front-end framework wujie, which generates an iframe to carry the js part and act as a js sandbox. Any content in the body tag of this iframe is placed in the shadow dom. As a result, rrweb cannot get the correct doucment object and thus cannot record the screen. How to solve this problem, thank you

wfk007:
@HappyPot which version are you using?  have you tried 2.0.0-alpha.18

HappyPot:
@wfk007 I'm glad to hear that. I'm using version 2.0.0-alpha.4 right now, and I plan to give the version you mentioned a try

PR #1763, opened 2025-11-21
fix: wujie shadow root

wfk007:
fix：https://github.com/rrweb-io/rrweb/issues/1682

reproduce step: https://github.com/wfk007/rrweb-wujie

### DIFF
--- a/.changeset/wet-bottles-flash.md
+++ b/.changeset/wet-bottles-flash.md
@@ -1,0 +1,6 @@
+---
+"rrweb": patch
+"@rrweb/utils": patch
+---
+
+fix: wujie monkeypatches ownerDocument


wfk007:
Discussion needed, It is not a elegant commit.

packages/rrweb/src/utils.ts
```diff
@@ -541,5 +541,6 @@ export function shadowHostInDom(n: Node): boolean {
 export function inDom(n: Node): boolean {
   const doc = n.ownerDocument;
   if (!doc) return false;
-  return dom.contains(doc, n) || shadowHostInDom(n);
+  // fix: https://github.com/rrweb-io/rrweb/issues/1682
+  return doc.contains(n) || shadowHostInDom(n);
```
Juice10: `dom.contains` was trying to get the original unproxied `.contains`.
Changing it to use the regular `doc.contains` could fix it for wujie but would break it for other frameworks that monkeypatch `.contains` and make it do non-standard behavior. 
To fix this correctly check out how `dom.contains` is implemented, it sounds like wujie is getting around our proxy bypass

wfk007: Happy New Year Justin! Thanks for your great suggestion, in fact, wujie modified the `this` reference in the iframe, binding it to the `this` reference in `shadowRoot`. To be honest, I didn't think it was a very elegant fix from the start and want to discuss with you.

packages/rrweb/src/utils.ts
```diff
@@ -541,5 +541,6 @@ export function shadowHostInDom(n: Node): boolean {
 export function inDom(n: Node): boolean {
   const doc = n.ownerDocument;
```
Juice10: It turns out the real problem is here, wujie monkey patches `n.ownerDocument`, we should make sure access to this method goes through `getUntaintedAccessor` like we do with many other accessors, then this should record ok:
https://github.com/rrweb-io/rrweb/blob/c059dbf41d32a4b05eaa564e7349065173b63549/packages/utils/src/index.ts#L172-L186

Juice10 (approved):

PR #1766, opened 2025-12-02
feat: Add option to add CSP to the replay iframe

KristianGrafana:
This PR adds support to add a Content-Security-Policy to the iframe. This is if users want to further tighten the security of the iframe by being able to block images, videos, objects, iframes and more. 

Let me know if you have any questions!


Juice10:
Hi @KristianGrafana, thanks for this addition. Since the csp attribute is [lacking some major browser support](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/csp#browser_compatibility) it would be good to document this so people who use it don't get a false sense of security. And in general it would be good to document this feature in the Guide.md and rrweb-player readme files

PR #1784, opened 2026-02-06
Remove postcss from dist/record.js (it was only intended for replay)

eoghanmurray:
This succeeds #1743 

Thanks to #1630 we should be able to see the effect of this in the CI as part of this PR 

## Tree-shaking postcss

The `rrweb-snapshot` package contains both snapshot (record-time) and rebuild
(replay-time) code. The rebuild module imports `postcss`, a large dependency
that is only needed during replay. Because `rrweb-snapshot` re-exports
everything from a single `index.ts` barrel file, bundlers that consume the
pre-built package see it as one indivisible chunk — they cannot tree-shake out
`rebuild.ts` and its `postcss` import, even when only snapshot functions are
used.

We initially tried adding `/*#__PURE__*/` annotations to side-effecting
expressions (e.g. `new RegExp(...)`) inside `rrweb-snapshot/src/css.ts`, hoping
Rollup would then drop the unused modules. This did not work because by the
time this package imports from the published `rrweb-snapshot` package, the
sources have already been bundled together — the individual module boundaries
that Rollup needs for tree-shaking no longer exist.

The solution (in `vite.config.ts`) uses two techniques:

1. **A custom `resolveId` plugin** redirects imports of `rrweb`, `rrweb-snapshot`,
   and `rrdom` to their original **source** entry points instead of the
   pre-built packages. This restores individual module boundaries so Rollup can
   see exactly which source files are used and which are not.

2. ~**`treeshake.moduleSideEffects: false`**~  [update - moved to `"sideEffects": false,` in rrweb-snapshot/package.json] tells Rollup it is safe to drop
   ~any module~ [the rrweb-snapshot modules] whose exports are not consumed, even if that module contains
   top-level code. This allows `rebuild.ts` (and therefore `postcss`) to be
   eliminated entirely from the record bundle.

Additionally, `packages/rrweb/src/entries/record.ts` was changed from a
default-import-then-re-export pattern to direct named re-exports, which gives
Rollup a clearer view of what is actually used.

A test in `packages/record/test/record.test.ts` asserts that no output bundle
file contains the string "postcss", guarding against future regressions.

Co-Authored by Claude AI

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -15,7 +15,6 @@ jobs:
     name: Tests
     permissions:
       contents: read
-      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -15,6 +15,7 @@ jobs:
     name: Tests
     permissions:
       contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
--- a/.github/workflows/style-check.yml
+++ b/.github/workflows/style-check.yml
@@ -3,6 +3,36 @@ name: Prettier Check
 on: [push, pull_request]
 
 jobs:
+  
+
+  bundle_size:
+    # Only runs on PRs (needs a base branch to compare against)
+    if: github.event_name == 'pull_request_target'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    name: Check Bundle Sizes
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.head_ref }}
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+          cache: 'yarn'
+      - name: Check bundle sizes
+        uses: preactjs/compressed-size-action@v2
+        with:
+          install-script: 'yarn install --frozen-lockfile'
+          build-script: 'build:all'
+          compression: 'none'
+          pattern: '**/dist/*.{js,cjs,mjs,css}'
+        env:
+          PUPPETEER_SKIP_DOWNLOAD: true
+
   prettier_check:
     # In the forked PR, it's hard to format code and push to the branch directly, so the action only check the format correctness.
     if: github.event_name != 'push' && github.event.pull_request.head.repo.full_name != 'rrweb-io/rrweb'
--- a/.github/workflows/style-check.yml
+++ b/.github/workflows/style-check.yml
@@ -3,36 +3,6 @@ name: Prettier Check
 on: [push, pull_request]
 
 jobs:
-  
-
-  bundle_size:
-    # Only runs on PRs (needs a base branch to compare against)
-    if: github.event_name == 'pull_request_target'
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      pull-requests: write
-    name: Check Bundle Sizes
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
-          ref: ${{ github.head_ref }}
-      - name: Setup Node
-        uses: actions/setup-node@v3
-        with:
-          node-version: lts/*
-          cache: 'yarn'
-      - name: Check bundle sizes
-        uses: preactjs/compressed-size-action@v2
-        with:
-          install-script: 'yarn install --frozen-lockfile'
-          build-script: 'build:all'
-          compression: 'none'
-          pattern: '**/dist/*.{js,cjs,mjs,css}'
-        env:
-          PUPPETEER_SKIP_DOWNLOAD: true
-
   prettier_check:
     # In the forked PR, it's hard to format code and push to the branch directly, so the action only check the format correctness.
     if: github.event_name != 'push' && github.event.pull_request.head.repo.full_name != 'rrweb-io/rrweb'
--- a/packages/record/README.md
+++ b/packages/record/README.md
@@ -55,9 +55,8 @@ record({
 ## Notes
 
 Currently this package is really just a wrapper around the `record` function in the main `rrweb` package.
-All `record` related code will get moved here in the future.  There are a few peculiarities documented in the [build notes](BUILD.md) related to this.
-
-
+All `record` related code will get moved here in the future. There is custom vite config for this package in
+order to optimize the build output. This is documented in the rrweb-snapshot [build notes](../rrweb-snapshot/BUILD.md).
 
 ## Sponsors
 
--- a/packages/record/README.md
+++ b/packages/record/README.md
@@ -93,7 +93,6 @@ Rollup a clearer view of what is actually used.
 A test in `packages/record/test/record.test.ts` asserts that no output bundle
 file contains the string "postcss", guarding against future regressions.
 
-
 ## Sponsors
 
 [Become a sponsor](https://opencollective.com/rrweb#sponsor) and get your logo on our README on Github with a link to your site.
--- a/packages/record/README.md
+++ b/packages/record/README.md
@@ -55,43 +55,9 @@ record({
 ## Notes
 
 Currently this package is really just a wrapper around the `record` function in the main `rrweb` package.
-All `record` related code will get moved here in the future.
+All `record` related code will get moved here in the future.  There are a few peculiarities documented in the [build notes](BUILD.md) related to this.
 
-## Tree-shaking postcss
 
-The `rrweb-snapshot` package contains both snapshot (record-time) and rebuild
-(replay-time) code. The rebuild module imports `postcss`, a large dependency
-that is only needed during replay. Because `rrweb-snapshot` re-exports
-everything from a single `index.ts` barrel file, bundlers that consume the
-pre-built package see it as one indivisible chunk — they cannot tree-shake out
-`rebuild.ts` and its `postcss` import, even when only snapshot functions are
-used.
-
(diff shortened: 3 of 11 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     name: Tests
     permissions:
       contents: read
+      pull-requests: write
```
Juice10: Can we remove this?

packages/record/vite.config.ts
Juice10, line 1: Let's try and generalize the changes here and move them to `/vite.config.default.ts`.
One of the reasons to move towards a unified build process was to reduce package specific bundle code.

packages/record/vite.config.ts
```diff
@@ -1,4 +1,59 @@
...
+  const baseConfig =
+    typeof baseConfigFn === 'function' ? baseConfigFn(env) : baseConfigFn;
+  return mergeConfig(baseConfig, {
+    plugins: [recordOnlyResolvePlugin()],
+    build: {
+      rollupOptions: {
+        treeshake: {
+          // Allow Rollup to drop modules whose exports are unused, even
+          // if they contain top-level code (e.g. rebuild.ts / postcss).
+          moduleSideEffects: false,
```
Juice10, line 51: How do we know this blanket statement doesn't have any unintended consequences?

packages/record/README.md
```diff
@@ -26,6 +26,42 @@ record({
...
+   top-level code. This allows `rebuild.ts` (and therefore `postcss`) to be
+   eliminated entirely from the record bundle.
+
+Additionally, `packages/rrweb/src/entries/record.ts` was changed from a
+default-import-then-re-export pattern to direct named re-exports, which gives
+Rollup a clearer view of what is actually used.
+
+A test in `packages/record/test/record.test.ts` asserts that no output bundle
+file contains the string "postcss", guarding against future regressions.
+
```
Juice10: It feels like the Readme.md isn't the right place for this description. It's quite helpful for as context for this PR, but it feels a little out of place here. Especially since the rest of this Readme is so bare bones

packages/record/vite.config.ts
Juice10: I had AI go through the whole repo, it seems like the risk is mostly minimal/hypothetical currently.

Juice10 (changes requested):

packages/record/README.md
eoghanmurray: Ya was thinking of this as a bit of a temporary thing, happy to remove from this PR

packages/record/vite.config.ts
eoghanmurray: Right, but I think this could also be solved with a more in-depth refactor of `rrweb-snapshot`, to make a cleaner separation of `snapshot` and `rebuild` functionality (specifically not having a shared `utils.ts` for both sides)

The `packages/record` and `packages/replay` are intended to be the final destination for the `packages/rrweb/src/record` and `packages/rrweb/src/replay` code, so that could help too in that we could then remove this 'hack'.  From that pov I'd vote to go forward with the exceptions in this PR only living in the ./packages/record folder and checking after those refactors whether we can remove

packages/record/vite.config.ts
eoghanmurray: I moved it up to root to see how it affects build for all projects; 
It won't work for `rrweb-player` as it results in the Sveldt library being dropped from the output:

You get a diff e.g. in rrweb-player.cjs where the following bit is removed:

```
  const PUBLIC_VERSION = "4";                                                                                                                                                           
  if (typeof window !== "undefined")                                                                                                                                                    
    (window.__svelte || (window.__svelte = { v: new Set() })).v.add(PUBLIC_VERSION);           
[+ many many lines]                                                                                     
```

Presumably that's Svelte making itself available (a side effect!).

There are also changes to the chrome extension, presumably because it pulls in rrweb-player but I didn't check

eoghanmurray: One other change  (maybe minor, maybe not) which affects all build outputs in all packages is the addition of a `/* @__PURE */` marker to  certain parts, e.g. in packages/all/dist/all.cjs:
```
var stringify_1$1;
var hasRequiredStringify$1;
function requireStringify$1() {
  if (hasRequiredStringify$1) return stringify_1$1;
  hasRequiredStringify$1 = 1;
  let Stringifier = /* @__PURE__ */ requireStringifier$1();
  function stringify(node2, builder) {
    let str = new Stringifier(builder);
    str.stringify(node2);
  }
  stringify_1$1 = stringify;
  stringify.default = stringify;
  return stringify_1$1;
}
var node$1;
var hasRequiredNode$1;
function requireNode$1() {
  if (hasRequiredNode$1) return node$1;
  hasRequiredNode$1 = 1;
  let { isClean, my } = /* @__PURE__ */ requireSymbols$1();
  let CssSyntaxError = /* @__PURE__ */ requireCssSyntaxError$1();
  let Stringifier = /* @__PURE__ */ requireStringifier$1();
  let stringify = /* @__PURE__ */ requireStringify$1();
  function cloneNode(obj, parent) {
    let cloned = new obj.constructor();
    for (let i2 in obj) {
      if (!Object.prototype.hasOwnProperty.call(obj, i2)) {
        continue;
```

I don't know if this is also a problem?

eoghanmurray: > How do we know this blanket statement doesn't have any unintended consequences?

We know from checking (at this point in time of the codebase) that it doesn't have any unintended consequences for the record package — perhaps it might have an unintended consequence in future, but that will be discovered as part of any PR that introduces a side-effecting-module for the record side by the record/dist/ output not working (assuming we are using the record package for testing).  I think this could be a good thing that we aggressively tree shake packages/record by default?

For now anyhow [I've removed](https://github.com/rrweb-io/rrweb/pull/1784/commits/c824c465a52a7720fd8013d4bc512d5d79922bc4) this blanket statement, but moved it into the rrweb-snapshot package.json instead as `sideEffects: false` there. 

I've added notes in record/vite.config.ts and also in rrweb-snapshot/BUILD.md to document this.

packages/record/README.md
eoghanmurray: Ultimately moved into rrweb-snapshot/BUILD.md and rewritten slightly

packages/record/vite.config.ts
eoghanmurray: So in latest the `rollupOptions are gone (commented out) from record/vite.config.ts 

That just leaves `recordOnlyResolvePlugin` as the main customization in that config.

>  Rollup plugin that redirects workspace package imports to their source entry
> points during JS bundling. This lets Rollup see individual source modules
> and tree-shake out replay-only code (rebuild.ts / postcss).

Is this a candidate for moving to the top level?  As I understand it, it would need to be rewritten as 'resolve to individual source modules instead of the 'barrel' outputs?  Is this a known technique?

packages/record/README.md
eoghanmurray: have moved into rrweb-snapshot as that's where the setting is located now

packages/record/vite.config.ts
Juice10: > Right, but I think this could also be solved with a more in-depth refactor of rrweb-snapshot, to make a cleaner separation of snapshot and rebuild functionality (specifically not having a shared utils.ts for both sides)

I created this PR to do some steps in that direction: https://github.com/eoghanmurray/rrweb/pull/2

packages/record/vite.config.ts
```diff
@@ -1,4 +1,66 @@
...
+        return aliases[source];
+      }
+      // rrdom is replay-only; stub it out so none of its code is bundled.
+      if (source === 'rrdom') {
+        return '\0rrdom-stub';
+      }
+    },
+    load(id) {
+      if (id === '\0rrdom-stub') {
+        return 'export class BaseRRNode {}; export class RRNode {}; export class RRIFrameElement {};';
```
Juice10: This is brittle. I get we don't really need rrdom for recording (there are only a few instances where it is used to do an instanceof check), however is manually removing it a substantial win for file size? Compared to the extra brittleness of having to add an rrdom stub that recreates an empty rrdom here.

eoghanmurray: Thanks, I've addressed by omitting any concrete (non type) rrdom imports and duck-typing the BaseRRNode check now in the latest forced commit

.github/workflows/style-check.yml
```diff
@@ -3,6 +3,36 @@ name: Prettier Check
...
+      - name: Check bundle sizes
+        uses: preactjs/compressed-size-action@v2
+        with:
+          install-script: 'yarn install --frozen-lockfile'
+          build-script: 'build:all'
+          compression: 'none'
+          pattern: '**/dist/*.{js,cjs,mjs,css}'
+        env:
+          PUPPETEER_SKIP_DOWNLOAD: true
+
```
Juice10:
```suggestion  
```

Added in another PR, this doesn't work anymore (no pull_request_target anymore)

packages/rrweb/src/utils.ts
```diff
@@ -414,8 +414,11 @@ export function hasShadowRoot<T extends Node | RRNode>(
   n: T,
 ): n is T & { shadowRoot: ShadowRoot } {
   if (!n) return false;
-  if (n instanceof BaseRRNode && 'shadowRoot' in n) {
-    return Boolean(n.shadowRoot);
+  const isRRDom = !(n instanceof Node);
```
Juice10, line 417: This change doesn't work with iframes since `globalThis.Node !== instanceOf iframeEl.contentWindow.Node`

Juice10 (changes requested):

Juice10:
~The rrweb-player numbers going down so heavily is amazing but also a little suspect since it should include postcss~
image
_EDIT: that is related to the vite6 update, and reduction in base64 encoded workers_

packages/rrweb/src/utils.ts
eoghanmurray: did this come up in a test or do I need to add one?

PR #1812, opened 2026-03-20
fix: always use native Proxy

juliecheng:
specific site was running into loading errors/crashing when recording, turn outs it's using a very old version of a 3rd party library (jspdf) that overwrites Proxy. whenever we ran `new Proxy`, it would result in a bunch of errors and cause the site to stop loading. this PR is to protect against overwritten Proxy and make sure we always use native Proxy 🫡

### DIFF
--- a/.changeset/brown-colts-join.md
+++ b/.changeset/brown-colts-join.md
@@ -1,0 +1,6 @@
+---
+"rrweb": patch
+"@rrweb/utils": patch
+---
+
+protect against overwritten Proxy and make sure we always use native Proxy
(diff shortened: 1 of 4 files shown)


Juice10:
Thanks for submitting this @juliecheng, I refactored it a little bit since there was some duplication and the tests were not passing here. Wanted to push it to your original branch but since its the Pendo org that wasn't allowed so I started a new PR for it: https://github.com/rrweb-io/rrweb/pull/1934/
Can you check to see if that works for you as well?

Closing this PR for now (unless you'd rather pull in my changes in which case I'd be more than happy to re-open this one!
Thanks again!

PR #1825, opened 2026-05-06
fix(snapshot): use link.sheet for CSS capture on WebKit

d0ugal:
## Problem

When `inlineStylesheet` is enabled, rrweb searches `document.styleSheets` to find the `CSSStyleSheet` for each `<link>` element and inline its rules as `_cssText`.

On WebKit (Safari), at `readyState: "interactive"`, external stylesheets loaded via `<link>` are **not yet listed** in `document.styleSheets` — only inline `<style>` elements appear. This means full snapshots taken at page load capture `_cssText: ""` for all external stylesheets, producing unstyled replays.

However, `HTMLLinkElement.sheet` **is** already populated at this point on WebKit.

## Fix

Check `link.sheet` before falling back to `document.styleSheets.find()`:

```typescript
const stylesheet =
  (n as HTMLLinkElement).sheet ||
  Array.from(doc.styleSheets).find((s) => {
    return s.href === (n as HTMLLinkElement).href;
  });
```

## Testing

Verified with Angular apps across Chrome, Firefox, and Safari (Playwright WebKit)

### DIFF
--- a/.changeset/shiny-socks-pull.md
+++ b/.changeset/shiny-socks-pull.md
@@ -1,0 +1,5 @@
+---
+'rrweb-snapshot': patch
+---
+
+Fix inline stylesheet capture on WebKit by reading CSS rules from `HTMLLinkElement.sheet` before falling back to `document.styleSheets`.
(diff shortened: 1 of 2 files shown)


eoghanmurray:
That's super and yes I think `.sheet` is fully usable now (maybe it always was?)

I've made a similar change in the StyleSheet Assets branch which will become rrweb 2.1:
https://github.com/rrweb-io/rrweb/commit/4206030844c41b27264e21893a9bb8f989255e28

I think you might not need the fallback to searching the array?

PR #1826, opened 2026-05-06
fix(utils): use Zone.js unpatched MutationObserver on Safari

d0ugal:
## Problem

When Angular's Zone.js is present, it monkey-patches `MutationObserver`. rrweb's `getUntaintedPrototype` detects the patched version and falls back to creating a hidden iframe, then uses `iframe.contentWindow.MutationObserver` to get an unpatched observer.

On Safari/WebKit, a `MutationObserver` obtained from an iframe's `contentWindow` **silently never fires callbacks** when observing nodes in the parent document. This means rrweb captures zero DOM mutations — full snapshots work, but all incremental changes are lost.

## Fix

Zone.js stores original unpatched globals under `globalThis[Zone.__symbol__(key)]`. Before attempting the iframe fallback, check if Zone.js has stored the native `MutationObserver` and use it directly.

```typescript
const zoneAlternative = angularZoneUnpatchedAlternative(key);
if (zoneAlternative) {
  return zoneAlternative.prototype;
}
```

This is the same approach used by PostHog: https://github.com/PostHog/posthog-js/pull/1687

## Testing

Verified with Angular apps (Zone.js present) across Chrome, Firefox, and Safari (Playwright WebKit). After the fix, Safari sessions capture 146+ mutations where previously they captured zero.

### DIFF
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -51,6 +51,22 @@ export const isAngularZonePresent = (): boolean => {
   return !!(globalThis as { Zone?: unknown }).Zone;
 };
 
+// When Angular Zone.js is present, it stores the original unpatched globals
+// under a symbol key. Use this to get the native MutationObserver (etc.)
+// without the iframe fallback, which breaks on Safari.
+// See: https://github.com/PostHog/posthog-js/pull/1687
+function angularZoneUnpatchedAlternative<T extends keyof BasePrototypeCache>(
+  key: T,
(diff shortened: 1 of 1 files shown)


eoghanmurray:
This was fixed by #1854 which I believe fixes both the Zone case and other types of MutationObserver  monkey patching on Safari?  Apologies I had not seen this PR prior to addressing that.

Is there still merit in avoiding the <iframe> workaround altogether for Zone given we can get to the unpatched versions?

PR #1829, opened 2026-05-14
[config] Deprecate inline asset recording options

Juice10:
## Summary

- Mark `recordOptions.inlineImages` and `recordOptions.inlineStylesheet` as deprecated in the public rrweb types.
- Update English and Chinese guide option tables to say both options remain supported in 2.0.0 but are planned to be superseded by future `captureAssets` asset recording APIs.
- Add a patch changeset for `rrweb`.

## Why

This lets the project defer the asset-events work from #1475 to a later 2.x release while giving users a 2.0.0 migration signal. Runtime behavior is unchanged.

## Validation

- `git diff --check`
- Attempted `yarn workspace rrweb check-types` via the bundled Yarn release, but this checkout does not have dependencies installed, so `tsc` was not available (`/bin/sh: tsc: command not found`).


YunFeng0817 (approved):

guide.md
```diff
@@ -221,14 +221,14 @@ The `record` function accepts the following options.
...
 | slimDOMOptions           | {}                 | remove unnecessary parts of the DOM <br />refer to the [list](https://github.com/rrweb-io/rrweb/blob/588164aa12f1d94576f89ae0210b98f6e971c895/packages/rrweb-snapshot/src/types.ts#L97-L108)  |
 | dataURLOptions           | {}                 | Canvas image format and quality ,This parameter will be passed to the OffscreenCanvas.convertToBlob(),Using this parameter effectively reduces the size of the recorded data                  |
-| inlineStylesheet         | true               | whether to inline the stylesheet in the events                                                                                                                                                |
+| inlineStylesheet         | true               | Deprecated since 2.0.0. Still supported, but planned to be superseded by future `captureAssets` asset recording APIs.                                                                         |
```
eoghanmurray, line 224: I don't think this is really deprecated in the assets branch; rather it sets an equivalent set of config options in captureAssets for you

guide.md
```diff
@@ -221,14 +221,14 @@ The `record` function accepts the following options.
...
 | recordCrossOriginIframes | false              | Whether to record cross origin iframes. rrweb has to be injected in each child iframe for this to work. Available options:<br/>`false`, <br/>`true`                                           |
 | recordAfter              | 'load'             | If the document is not ready, then the recorder will start recording after the specified event is fired. Available options: `DOMContentLoaded`, `load`                                        |
-| inlineImages             | false              | whether to record the image content                                                                                                                                                           |
+| inlineImages             | false              | Deprecated since 2.0.0. Still supported, but planned to be superseded by future `captureAssets` asset recording APIs.                                                                         |
```
eoghanmurray, line 231: (not 100% on this but) ditto for inlineImages which sets `{ objectURLs: true, origins: true }` for captureAssets which isn't the same as `{ images: true }` which captures all images.

guide.md
eoghanmurray: we'll keep it as a first class option as it sets the config so that "it will only inline stylesheets that are CORs accessible" i.e. same as current `inlineStylesheet: true` behaviour

PR #1833, opened 2026-05-15
[codex] Add asset event types

Juice10:
## Summary

- Extract the minimal public asset-event type contract from #1475 for the 2.0 event surface
- Add `EventType.Asset`, `assetEvent`, `assetEventWithTime`, `assetParam`, and `assetCallback` to `@rrweb/types`
- Factor `SerializedBlobArg` and add `SerializedCssTextArg` so the asset payload type can reference stylesheet payloads without pulling in the broader asset runtime

## Why

Adding a new event kind after 2.0 can break consumers with exhaustive event handling. This lands only the compatibility-sensitive type surface before the stable release, without pulling in #1475 runtime behavior or the broader `captureAssets` API.

## Not included

- No `captureAssetsParam` or `recordOptions.captureAssets`
- No runtime asset manager behavior
- No snapshot/rebuild manager API
- No docs, package versions, lockfile, or #1475 branch drift

## Validation

- `git diff --check`
- Subagent ran focused compile/Prettier checks for the changed type surface
- Full repo `yarn` checks were not runnable in this Codex shell because `yarn`, `npm`, `npx`, and `corepack` are unavailable in PATH and the project rejects `pnpm`

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -6,6 +6,7 @@ export enum EventType {
   Meta,
   Custom,
   Plugin,
+  Asset,
 }
 
 export type domContentLoadedEvent = {
@@ -57,6 +58,15 @@ export type pluginEvent<T = unknown> = {
     plugin: string;
     payload: T;
   };
+};
+
+export type assetEvent = {
+  type: EventType.Asset;
+  data: assetParam;
+};
+
+export type assetEventWithTime = assetEvent & {
+  timestamp: number;
 };
 
 export enum IncrementalSource {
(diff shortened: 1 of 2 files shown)


eoghanmurray (approved):

eoghanmurray:
The `cssTexts: []` was something we discussed Justin; the fact that it's an array is so that processor's will expect an array, but usually it will have one item.  The case for multiple items is multiple textContent children on a `<style>` element which is only possible with programmatic childNode appending to a style element (`<link>` els always have multiple.   Just to note.

packages/types/src/index.ts
```diff
@@ -382,16 +393,23 @@ export enum CanvasContext {
   WebGL2,
 }
 
+export type SerializedCssTextArg = {
+  rr_type: 'CssText';
+  cssTexts: string[];
```
eoghanmurray:
```suggestion
  cssTexts: string[]; // usually just a single item
```

eoghanmurray (dismissed):

eoghanmurray (approved):

PR #1835, opened 2026-05-19
[codex] Tree-shake postcss from record bundle

Juice10:
## Summary
- Marks `rrweb-snapshot` as side-effect-free so unused rebuild modules can be dropped.
- Resolves `rrweb`, `rrweb-snapshot`, and `rrdom` to local source entrypoints during the `@rrweb/record` build.
- Adds regression coverage that record bundles do not contain `postcss` and that `dist/record.js` stays at least 200 KiB below the measured baseline.

## Why
`@rrweb/record` imports through pre-built package barrels, which hides module boundaries from Rollup and keeps replay-only `rrweb-snapshot` rebuild code, including `postcss`, in record-time bundles. The record build now sees the source module graph and can drop replay-only code without changing the public package API.

## Validation
- `yarn workspace rrweb-snapshot check-types`
- `yarn workspace @rrweb/record build`
- `yarn workspace @rrweb/record test`
- `rg -n "postcss" packages/record/dist --glob '*.{js,cjs}'` found no matches

Bundle size check:
- Before: `packages/record/dist/record.js` was 397373 bytes
- After: `packages/record/dist/record.js` is 161287 bytes
- Reduction: 236086 bytes

### DIFF
--- a/.changeset/record-tree-shake-postcss.md
+++ b/.changeset/record-tree-shake-postcss.md
@@ -1,0 +1,7 @@
+---
+"@rrweb/record": patch
+"rrweb": patch
+"rrweb-snapshot": patch
+---
+
+Tree-shake replay-only `postcss` code from the `@rrweb/record` bundle.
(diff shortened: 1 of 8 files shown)


eoghanmurray:
This is an alternative to #1784

Juice10:
Superseded by an origin-hosted branch/PR so checks run against rrweb-io default workflows instead of fork default workflows.

PR #1854, opened 2026-06-02
Fix broken MutationObserver when page is monkey patched in Webkit/Safari

eoghanmurray:
#1509 added support for recording on SalesForce LWC (Lightning Web Security) pages and all similar frameworks that monkey patch built in functions that rrweb relies on
#1597 extended this to include AngularJS (which appears to have untainted methods but in fact doesn't)

Unfortunately both of these silenced mutation recording events (type: IncrementalSnapshot, source: Mutation) on Safari due to a difference in how WebKit handles lifecycle of the mutation observer: https://webkit.org/b/179224. The following versions were confirmed to have no mutation recording on an Angular website:
    
     Mozilla/5.0 (iPhone; CPU iPhone OS 18_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.3 Mobile/15E148 Safari/604.1
     Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.3 Mobile/15E148 Safari/604.1

This PR introduces a new test regime involving Docker in order to test the WebKit browser engine.   I haven't attempted to run the entire test suite using this method, but it's possible we'd uncover other webkit bugs if we did so.


packages/rrweb/package.json
```diff
@@ -72,6 +73,7 @@
     "ignore-styles": "^5.0.1",
     "inquirer": "^9.0.0",
     "jest-image-snapshot": "^6.2.0",
+    "playwright": "^1.44.0",
```
Juice10: 1.60.0 is the latest, had some issues with this older version and ci a couple days ago

packages/rrweb-snapshot/package.json
```diff
@@ -7,6 +7,7 @@
     "prepack": "npm run build",
     "retest": "vitest run",
     "test": "yarn build && vitest run",
+    "test:webkit": "yarn build && vitest run --config vitest.config.webkit.ts",
```
Juice10, line 10: `vitest.config.webkit.ts` doesn't exist in this pr, I haven't double checked but I doubt we already had it

packages/utils/src/index.ts
```diff
@@ -93,17 +96,33 @@ export function getUntaintedPrototype<T extends keyof BasePrototypeCache>(
...
+      iframeEl.remove();
+      return defaultPrototype;
+    }
+
+    // WebKit/Safari: WebKit tears down an iframe's ScriptExecutionContext when it is
+    // detached from the DOM. MutationObserver.deliver() silently drops callbacks when
+    // m_callback->scriptExecutionContext() returns null (webkit.org/b/179224).
+    // Keep the iframe attached so its context stays live, and expose a cleanup fn.
+    const ua = navigator.userAgent;
+    if (ua.includes('Safari') && !ua.includes('Chrome')) {
```
Juice10, line 118: Can we use whatever was breaking down to detect if we need to go down this code path? Seeing as us detection [is notoriously flaky](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/User-agent_reduction)

eoghanmurray: Ya I considered this but have discounted it thus far; we'd need to set up the mutation observer and check whether it can capture a mutation without the iframe present (an async process).  So we'd have to wait some length of time to decide that no mutations are getting captured, and then start again and re-execute the setup code (put the iframe permanently back on the page, and set up the mutation observer again).

Another alternative would be to remove the iframe for those browsers that are _known_ to work okay (although similarly brittle to current situation).

Given this is a known Safari bug vs. spec behaviour, I figured the explicit ua detection is a reasonable accommodation.

A further option is to just always have the iframe present on the page.

packages/rrweb-snapshot/package.json
eoghanmurray: Thanks, I forgot to add them; have added in latest commit

packages/rrweb/package.json
eoghanmurray: Thanks, the docker image was installing 1.60 ... should have been 1.60 here too, fixed in latest

packages/rrweb/test/record/monkey-patched.test.ts
```diff
@@ -0,0 +1,156 @@
...
+}
+
+async function waitForRAF(page: Page): Promise<void> {
+  await page.evaluate(
+    () =>
+      new Promise<void>((resolve) => {
+        requestAnimationFrame(() => requestAnimationFrame(() => resolve()));
+      }),
+  );
+}
```
Juice10: I think you can pull in these functions from our other test helpers. (Maybe waitForRAF will need type tweaking to also allow for playwright's Page types)

PR #1871, opened 2026-06-09
Clarify that `inlineStylesheet` and `inlineImages` options are not deprecated

eoghanmurray:
This reverts  #1829 which was released in 2.0.0

The intention in #1475 is to keep them as simple versions of the config, and that if on, they'll set sensible defaults for the `assetConfig` option.


Juice10:
> The intention in https://github.com/rrweb-io/rrweb/pull/1475 is to keep them as simple versions of the config, and that if on, they'll set sensible defaults for the assetConfig option. 

In that case we should update #1475 because once that gets merged it will mark it as deprecated again.

eoghanmurray:
Having second thoughts on this PR as with Assets, they will never actually be "inlined"

Want to also note that the same two options when supplied in the rrweb-snapshot branch directly will not be deprecated (#1829 didn't touch [those options](https://github.com/rrweb-io/rrweb/blob/main/packages/rrweb-snapshot/src/snapshot.ts#L1237))

eoghanmurray:
As per above comment; we'll keep them deprecated but support them for a long time

Issue #1820, opened 2026-04-15
Session Replay (rrweb) blocks main thread when Zendesk Web Widget is present


sbimochan:
## Is there an existing issue for this?

- [x] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [x] I have reviewed the documentation https://docs.sentry.io/
- [x] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

## Environment

- **SaaS** (sentry.io)
- **SDK**: `@sentry/react` (also reproducible with `@sentry/browser`)
- **SDK Version**: _[fill in your version, e.g. 8.x.x]_
- **Browser**: Chrome 130+ (reproducible across browsers)
- **Framework**: React (Vite dev server, localhost:3000)

## Problem Statement

When Session Replay is enabled (`replaysSessionSampleRate > 0` or `replaysOnErrorSampleRate > 0`), interacting with the **Zendesk Web Widget** causes significant main thread blocking. The UI becomes unresponsive — clicks are delayed, scrolling stutters, and Sentry itself reports `ui.long-task` spans with the transaction name `Main UI thread blocked`.

**Setting both `replaysSessionSampleRate` and `replaysOnErrorSampleRate` to `0` completely eliminates the thread blocking**, confirming that rrweb's DOM observation/serialization is the root cause — not the Zendesk widget's own JavaScript.

## Steps to Reproduce

1. Initialize Sentry with Session Replay enabled:

```javascript
Sentry.init({
  dsn: "YOUR_DSN",
  replaysSessionSampleRate: 1.0,
  replaysOnErrorSampleRate: 1.0,
  integrations: [
    Sentry.replayIntegration(),
  ],
});
```

2. Load the Zendesk Web Widget on the same page (standard snippet from Zendesk):

```html
<script id="ze-snippet" src="https://static.zdassets.com/ekr/snippet.js?key=YOUR_KEY"></script>
```

3. Click on the Zendesk widget launcher button to open the widget.
4. Observe main thread blocking in DevTools Performance tab and Sentry logs.

## Observed Behavior

Console output from Sentry debug mode shows continuous span creation:

```
Sentry Logger [log]: [Tracing] Starting 'http.client' span on transaction 'GET https://ekr.zdassets.com/compose/...'
Sentry Logger [log]: [Tracing] Starting 'ui.long-task' span on transaction 'Main UI thread blocked'
[...]

Issue #1447, opened 2024-04-11
[Bug]: Performance issue with Mutation observer


NatanCoralogix:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb

### Version

2.0.0-alpha.11

### Expected Behavior

putting rr-block on a complex elements, such as highcharts svg charts,
should not only block element from the Replay, but also skip any manipulations done by rrweb

### Actual Behavior

We have a page with many Highcharts charts, and the page has lag, also noticed by profiling the page, even if we add rr-block to the parent

### Steps to Reproduce

create a page with complex elements, that will reach a large number of mutations,
in our case its many charts in same page.
add rr-block to the parent, and see it doesnt help for perofrmance, the page is laggy.
all rr-block is doing is to remove element from replay later.

### Additional Information

We monitor the page and listen to the mutation observer,
whenever there is a large number 10k+ the page gets stuck due to many rrweb manipulations.
Intuitive, i would think adding rr-block will skip any rrweb logic on this elements.

maybe there is other solution to really exclude heavy elements from rrweb?

eoghanmurray:
The mutation observer is set up to monitor the entire page (a single set up of an observer for root node + descendants).
When a mutation happens on a blocked element, or on an element whose ancestor is blocked, then the element still needs to be checked to figure out that the mutation should be ignored.
I imagine that's what's going on, but I can't think of an easy way to mitigate.

eoghanmurray:
@mdellanoce has made some PRs related to mutation performance which have yet to land in trunk.
It would be helpful to those PRs if you could provide timings for your test case before/after you cherry-pick https://github.com/rrweb-io/rrweb/pull/1368 and https://github.com/rrweb-io/rrweb/pull/1398

wardpeet:
We are using posthog and have a similar issue.  Seems like it's the emit function. I'm not sure how I can test your patches within posthog though.

![image](https://github.com/user-attachments/assets/43f46460-f7a9-4db7-a22a-b36f5ca909e7)

PR #1873, opened 2026-06-10
feat(record): add `attributeFilter` option for native MutationObserver filtering

kunalfermat:
## Summary

Add an `attributeFilter` option to `record()` that delegates attribute filtering to the browser's native `MutationObserver.observe({ attributeFilter })`. When set, unlisted attributes never fire the JS callback, eliminating CPU overhead from high-frequency style animations entirely — not just filtering after the fact.

## Problem

CSS animations on carousels, sliders, and animated badges generate 1,000+ `style` attribute mutations per second with zero user interaction. On sites like littlesleepies.com, **99% of all rrweb events are attribute mutations** from a product carousel (7,200 mutations/5s) and animated VIP badges (1,200 mutations/5s).

Today there is no way to prevent these from reaching the MutationObserver callback:
- `ignoreCSSAttributes` only filters CSSOM `CSSStyleDeclaration.setProperty()` calls, not inline `style` attribute changes observed by MutationObserver
- `hooks.mutation` fires after collection — the expensive processing has already happened
- `blockSelector` hides the element entirely from replay

Related issues: #197, #221, #668, #1447, #1820
Complementary to: #1694 (mutation emission throttling)

## Solution

Pass `attributeFilter` through to the native `MutationObserver.observe()` call. This is the most performant approach — the browser never fires the JS callback for filtered-out attributes, so there is zero CPU overhead per-mutation for unlisted attributes.

```js
record({
  emit(event) { /* ... */ },
  // Only observe class, value, and form-state attributes.
  // 'style' mutations (animations) never fire — zero CPU cost.
  attributeFilter: ['class', 'value', 'checked', 'selected', 'disabled', 'src', 'href'],
});
```

When `attributeFilter` is omitted or undefined, all attributes are observed (existing default behavior — fully backwards compatible).

## Changes

- `packages/rrweb/src/types.ts` — Added `attributeFilter?: string[]` to `recordOptions`, `observerParam`, and `MutationBufferParam`
- `packages/rrweb/src/record/observer.ts` — Pass `attributeFilter` to `observer.observe()` when provided
- `packages/rrweb/src/record/index.ts` — Wire `attributeFilter` from record options through to observer params
- `.changeset/add-attribute-filter.md` — Minor changeset

## Trade-offs

- **Opt-in only**: No behavior change unless `attributeFilter` is explicitly set
- **All-or-nothing per attribute name**: If you filter out `style`, ALL style mutations are invisible (not just animation ones). Users who need some style mutations recorded should not filter `style`
- **Replay fidelity**: Filtered attributes won't appear in replay. This is acceptable for `style` animations (visually indistinguishable at 0/s vs 1,400/s) but users should be aware

## Real-world impact

Tested on littlesleepies.com (product carousel + VIP badges):
- Without filter: 988 mutation events in 15s idle, 15,763 attribute changes
- With `attributeFilter: ['class']`: MutationObserver callback never fires for style changes → ~99% reduction in mutation processing


eoghanmurray:
This looks like an interesting idea and cuts off the problem at the source, preventing unnecessary processing in rrweb.
Critique; this is putting a big burden on the user of rrweb to correctly enumerate all attributes that may an effect on the playback.

~An alternate take that I haven't had a chance to revisit recently is #1694 (throttling mutation emission)
If you hadn't already seen that, it would be great to see if it alleviates the problem for your use case.~

Edit: I see you mentioned #1694 already — did you test out any throttling values for your problem website?

eoghanmurray:
There is also (undocumented I now realize) `record.freezePage` option which stops emission of mutations (although doesn't stop the processing which appears to be a factor in your case) ... I'd usually have `freezePage` kick in after a certain period of user inactivity.

kunalfermat:
> This looks like an interesting idea and cuts off the problem at the source, preventing unnecessary processing in rrweb. Critique; this is putting a big burden on the user of rrweb to correctly enumerate all attributes that may an effect on the playback.
> 
> ~An alternate take that I haven't had a chance to revisit recently is #1694 (throttling mutation emission) If you hadn't already seen that, it would be great to see if it alleviates the problem for your use case.~
> 
> Edit: I see you mentioned #1694 already — did you test out any throttling values for your problem website?

Thanks for the feedback @eoghanmurray !                                                                                                                        
                                                                         
On the ergonomics concern — totally fair. An allowlist puts the burden on users to enumerate every attribute they care about, which is error-prone. A simpler alternative would be an inverse option like ignoreAttributes: ['style'] — users only list what to exclude, which is a much shorter list (usually just style).  That's easier to reason about and harder to get wrong. Happy to rework the API in that direction if you think it's a better fit.                              
                                                                                                                                                                
 On #1694 (mutation emission throttling) — we actually implemented something very similar on our side. Our pixel has an emit-callback throttle that drops style-only attribute mutations firing faster than 100ms apart. It works well for bandwidth/storage (~85% reduction in mutation events on our noisiest brand site), but doesn't help with CPU because rrweb's MutationObserver callback still fires 1,400+/s and processes every mutation before our throttle sees it. That's what motivated the native attributeFilter approach — the browser never fires the JS callback for filtered-out attributes, so the expensive processMutation path (getAttribute, style diffing, etc.) never runs.                                                                                          
                                                                                                                                                              
That said, #1694 and this PR are complementary — throttling controls how often mutations are emitted, while attributeFilter/ignoreAttributes controls which mutations are observed. Both have value for different use cases.

kunalfermat:
> There is also (undocumented I now realize) `record.freezePage` option which stops emission of mutations (although doesn't stop the processing which appears to be a factor in your case) ... I'd usually have `freezePage` kick in after a certain period of user inactivity.

`record.freezePage` — interesting, didn't know about that. For our use case (session recording on e-commerce sites), the animations fire continuously even while users are actively browsing/adding to cart, so freezing on inactivity wouldn't catch the overlap period. But it's a good tool for other scenarios.

eoghanmurray:
Ok I'm leaning towards wontfix as I believe any sort of attributeFilter means we can't discover *new* attributes being added to an element, so we'd get a hidden gap.  The problem is that it's really hard to explain this in the config docs.  As you have to understand how https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe works anyway to use this safely, I think it's fair to recommend wiring this up directly in a fork of the repo (where `.observe()` is called).

My other objection is that you are setting a filter to target e.g. *all* style attributes, whereas a subset might be the problem.  A no-config solution to that would be to detect the high frequency/noisy element+attribute combinations and take them 'offline', but we've currently just one root observer set up and I dunno how a single element could be omitted.

Happy to listen to your thoughts or more detail on your noisy  example.

eoghanmurray:
> On mutation emission throttling — we actually implemented something very similar on our side. Our pixel has an emit-callback throttle that drops style-only attribute mutations firing faster than 100ms apart. It works well for bandwidth/storage (~85% reduction in mutation events on our noisiest brand site), but doesn't help with CPU because rrweb's MutationObserver callback still fires 1,400+/s and processes every mutation before our throttle sees it.

I just realized that #1933 might fix the worst of this.  Basically the style mutation is special in that it's processing is more complicated (it drops in a dummy element so as to construct a before/after comparison of styles).  Before #1933 this happens every time a mutation comes in, but afterwards it only happens upon emit.  If you also add the 100ms (or higher) mutation throttling config now available in that PR, it should pick out the noisy elements only and do that style construction only once per 100ms and I think it should have a major relief on cpu.

New PRs now give you a pre-built package to try it out:
https://github.com/rrweb-io/rrweb/pull/1933#issuecomment-5601408890
Would you be able to test it from a CPU performance perspective prior to merge?

PR #1887, opened 2026-06-27
release 2.1.0

Juice10:
## What changed

- Configure Changesets to only major-bump peer dependents when the new peer version is outside the declared range.
- Regenerate the pending release output as 2.1.0 instead of 3.0.0.
- Keep the existing fixed package group and the main release branch config intact.

## Why

The network plugin changes add new packages and new exported @rrweb/types shapes, which should be a minor release. Changesets was conservatively escalating a peer dependency minor bump to major, and the fixed group propagated that to every package.

## Impact

The release PR stays on the 2.x line while still updating all fixed-group package versions and internal dependency ranges together.

## Validation

- Reproduced the existing release plan resolving every package to 3.0.0 before the config change.
- Verified the release plan resolves every releasable package to 2.1.0 after the config change.
- Ran `GITHUB_TOKEN="$(gh auth token)" yarn changeset version`.
- Verified all fixed-group package versions are 2.1.0 and no internal dependency points at 3.0.0.
- Ran `git diff --check`.

## Follow-up note

The new network plugin packages manually tighten peer dependency ranges to `^2.1.0`. In particular, they require the `@rrweb/types` network exports introduced in this release, so advertising compatibility with `@rrweb/types@^2.0.1` would be too broad even though Changesets leaves in-range peer dependencies unchanged.


Juice10:
Superseded by the automatically generated release PR #1889, which now includes the network plugin peer dependency updates.

PR #1900, opened 2026-07-14
feat(plugin): add window popup plugin to record popup events

smeng9:
Record `window.alert` / `window.confirm` / `window.prompt` events as part of UI interaction


smeng9:
Hi @Juice10 

It has been a while since this PR is opened. This PR makes the plugin suite more complete by adding the capability to record user's interaction and response in a popup. It resolves the issue that the recording shows no interaction for a while when user makes a decision for the native popup. Thanks for your time to look into this PR.

PR #1901, opened 2026-07-14
Update vite config to prevent Svelte's SSR lifecycle stubs being bundled

### LINKED ISSUE #1872: [Bug]: rrweb-player 2.0.0 & 2.0.1 published `dist` is broken — Replayer is never created, `play()` throws `controller.play` of undefined
### Preflight Checklist

rrweb-player

### Affected versions
- ❌ `rrweb-player@2.0.1` (latest)
- ❌ `rrweb-player@2.0.0`
- ✅ `rrweb-player@2.0.0-alpha.20` (works)
- ✅ `rrweb-player@2.0.0-alpha.18` (works)

### Expected Behavior

The published build artifact (`dist/rrweb-player.js` / `dist/rrweb-player.umd.cjs`) for
2.0.0 and 2.0.1 is missing all Svelte lifecycle code. The internal rrweb `Replayer` is
never instantiated, so the controller is never rendered, and calling `play()` throws.
The source in `packages/rrweb-player/src/Player.svelte` is correct — only the published
bundle is defective, which points to a build/publish regression (likely the Svelte/Vite
compile during the "Module Distribution Overhaul" in the 2.0.0 release that split out
`@rrweb/replay`).

### Reproduction
```js
import rrwebPlayer from 'rrweb-player';

const player = new rrwebPlayer({
  target: document.getElementById('player'),
  props: { events, showController: true, autoPlay: false },
});

player.play();
// → Uncaught TypeError: Cannot read properties of undefined (reading 'play')
//      at Player.play (rrweb-player.js)  -> controller.play()
```

`autoPlay: true` is equally broken (nothing plays), because the Replayer is never created.

### Root cause (evidence from the published artifact)
In `node_modules/rrweb-player/dist/rrweb-player.js` (2.0.1):

1. The Player component declares `let replayer;` but **never assigns it** — there is no
   `new Replayer(...)` anywhere in the bundle, and **no `onMount`** at all
   (`grep -c onMount dist/rrweb-player.js` → `0`).
2. Because `replayer` stays `undefined`, the `{#if replayer}` block that mounts the
   `Controller` (and sets the `controller` binding) never runs → `controller` is `undefined`.
3. Public `play()` does `controller.play()` → `TypeError`.
4. Even the `Controller`'s methods are empty stubs in this build:
   ```js
   const play  = () => { { return; } };
   const pause = () => { { return; } };
   const toggle = () => {};
   ```

Compare the **working** `2.0.0-alpha.20` build, where the same functions are intact:
```js
// Player onMount creates the Replayer; Controller.play is real:
const play = () => {
  if (playerState !== "paused") return;
  if (finished) { replayer.play(); ... }
};
```

| Version | `onMount` in dist | `new Replayer` in dist | Playback |
|---|---|---|---|
| 2.0.0-alpha.18 | 3 | yes | works |
| 2.0.0-alpha.20 | 3 | yes | works |
| 2.0.0 | 0 | no | broken |
| 2.0.1 | 0 | no | broken |

This matches `Player.svelte` on `master`, which still creates the Replayer in `onMount`
(`replayer = new Replayer(events, { speed, root: frame, unpackFn: unpack, ...$$props })`)
and defines `play()` as `controller.play()`. So the regression is in how 2.0.0/2.0.1 were
built/published, not in the source.

### Expected
The published `dist` for 2.0.0/2.0.1 should contain the `onMount` lifecycle that
instantiates the Replayer, and functional `play()/pause()/toggle()`, as the alpha builds and
the current source do.

### Environment
- rrweb-player 2.0.1, bundled via an app build (Angular) importing the ESM `dist/rrweb-player.js`
- Reproducible with both the ESM (`dist/rrweb-player.js`) and UMD (`dist/rrweb-player.umd.cjs`) entries

### Workaround
Pin to `rrweb-player@2.0.0-alpha.20` (latest version with a working build).
```

### PR DESCRIPTION
**Pull request by kevinvandijk**
Fixes #1872 

This needs a bit more verification but as far as I can see this fixes the issue mentioned above

### DIFF
--- a/packages/rrweb-player/vite.config.ts
+++ b/packages/rrweb-player/vite.config.ts
@@ -1,6 +1,11 @@
 import path from 'path';
 import glob from 'fast-glob';
-import { Plugin } from 'vite';
+import {
+  defaultClientConditions,
+  defineConfig,
+  mergeConfig,
+  Plugin,
+} from 'vite';
 import config from '../../vite.config.default';
 import { svelte } from '@sveltejs/vite-plugin-svelte';
 import sveltePreprocess from 'svelte-preprocess';
@@ -30,6 +35,7 @@ async function generateDts(inputPath: string) {
 function viteSvelteDts(): Plugin {
   return {
     name: 'vite-plugin-svelte-dts',
+    apply: 'build',
     async buildStart(options) {
       console.log('Generating .d.ts files for Svelte components...');
 
@@ -60,11 +66,21 @@ function viteSvelteDts(): Plugin {
   };
 }
 
-export default config(path.resolve(__dirname, 'src/main.ts'), 'rrwebPlayer', {
-  plugins: [
-    viteSvelteDts(),
-    svelte({
-      preprocess: [sveltePreprocess({ typescript: true })],
-    }),
-  ],
-});
+const sveltePlugins = svelte({
+  preprocess: [sveltePreprocess({ typescript: true })],
+}) as Plugin[];
+const baseConfig = config(
+  path.resolve(__dirname, 'src/main.ts'),
+  'rrwebPlayer',
+  {
+    plugins: [viteSvelteDts(), ...sveltePlugins],
+  },
+);
+
+export default defineConfig((env) =>
+  mergeConfig(typeof baseConfig === 'function' ? baseConfig(env) : baseConfig, {
+    resolve: {
+      conditions: [...defaultClientConditions],
+    },
+  }),
+);
--- a/.changeset/fix-rrweb-player-browser-runtime.md
+++ b/.changeset/fix-rrweb-player-browser-runtime.md
@@ -1,0 +1,5 @@
+---
+"rrweb-player": patch
+---
+
+Fix the rrweb-player production build so Svelte lifecycle APIs resolve to the browser runtime instead of to stubs
--- a/packages/rrweb-player/package.json
+++ b/packages/rrweb-player/package.json
@@ -16,7 +16,8 @@
     "svelte-preprocess": "^5.0.3",
     "svelte2tsx": "^0.7.30",
     "tslib": "^2.0.0",
-    "vite": "^6.0.1"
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by eoghanmurray** [comment]
> ## Bundle Size Changes
> 
> **Size change:** +650.04 kB (+6.65%) | **Total size:** 10425.80 kB
> `rrweb-player` 🔍 - 993.84 kB -> 1643.88 kB (+650.04 kB (+65.41%))
> File 	Base 	PR 	Diff
> `rrweb-player.cjs` 	287.43 kB 	473.37 kB 	+185.93 kB (+64.69%) 🔍
> `rrweb-player.js` 	287.31 kB 	473.24 kB 	+185.93 kB (+64.71%) 🔍
> `rrweb-player.umd.cjs` 	289.69 kB 	476.16 kB 	+186.47 kB (+64.37%) 🔍
> `rrweb-player.umd.min.cjs` 	129.40 kB 	221.11 kB 	+91.71 kB (+70.87%) 🔍

2.0.0-alpha.18 had the following `dist` sizes, so this increase seems to me to be due to the fix rather than any real increase:

```
449K rrweb-player.cjs
449K rrweb-player.js
452K rrweb-player.umd.cjs
218K rrweb-player.umd.min.cjs
```

File: packages/rrweb-player/vite.config.ts [hunk 1]
```diff
@@ -60,11 +66,21 @@ function viteSvelteDts(): Plugin {
...
+const baseConfig = config(
+  path.resolve(__dirname, 'src/main.ts'),
+  'rrwebPlayer',
+  {
+    plugins: [viteSvelteDts(), ...sveltePlugins],
+  },
+);
+
+export default defineConfig((env) =>
+  mergeConfig(typeof baseConfig === 'function' ? baseConfig(env) : baseConfig, {
```
Juice10, line 81: Doesn't `baseConfig(env)` always output a function?
Curious why this ternary statement is in here

Juice10 (approved):

PR #1904, opened 2026-07-17
fix(rrweb-plugin-console-record): bind wrapped console methods to the logger

### LINKED ISSUE #1772: [Bug]: Importing getRecordConsolePlugin version 2.0.0-alpha.18, and referencing it crashes chrome extension content script.
### Preflight Checklist

Other (specify below)

### Version

@rrweb/rrweb-plugin-console-record: 2.0.0-alpha.18

### Expected Behavior

I want to import and use the console recorder plugin, as it provides in the recipes:

```
import rrweb from 'rrweb';
import { getRecordConsolePlugin } from '@rrweb/rrweb-plugin-console-record';

rrweb.record({
  emit: function emit(event) {
    // you should use console.log in this way to avoid errors.
    const defaultLog = console.log['__rrweb_original__']
      ? console.log['__rrweb_original__']
      : console.log;
    defaultLog(event);
  },
  // to use default record option
  plugins: [getRecordConsolePlugin()],
});
```

### Actual Behavior

All is well until I run my extension in the dev environment, I find that the content script doesn't run at all, possibly indicating a crash during the build process, perhaps some operation is being run within the package version that throws an error. My content script widget doesn't appear.

### Steps to Reproduce

- Create an extension project
- Install @rrweb/rrweb-plugin-console-record: 2.0.0-alpha.18 as a dependency
- Use this in a content script
- Go to browser to see if it crashes

### Testcase Gist URL

_No response_

### Additional Information

I downgraded to version @rrweb/rrweb-plugin-console-record: 2.0.0-alpha.15, and my code is now working as expected.

yashrao2607:
## Summary

`getRecordConsolePlugin` patches each console method with an arrow function that calls `original.apply(this, args)`. Arrow functions ignore the call-site receiver entirely and use their *lexically enclosing* `this` — here that's the `replace()` function, which is invoked as a plain function call, so `this` is `undefined` in this module's strict mode. In other words `this` was **never** the logger, regardless of how the wrapped method was actually called.

Native console implementations can throw `TypeError: Illegal invocation` when a method is applied with the wrong receiver. That's consistent with the reported behavior: importing this plugin silently crashed a Chrome extension's content script (isolated-world consoles appear to enforce this more strictly than a page's main-world console, which is likely why this wasn't caught earlier).

Fixed by binding explicitly to `_logger` (the actual logger object passed into `replace()`), both in the normal call path and in the internal-error fallback (`catch` block), which had the identical bug (`original('rrweb logger error:', ...)`, an unbound plain call).

## Test plan

- [x] Added `test/this-binding.test.ts`: patches a fake logger whose method records `this`, calls it, and asserts `this` is the logger instance. Verified this test fails against the pre-fix code (`this === undefined`) and passes after the fix.
- [x] `yarn vitest run` for this package: all existing tests (including the puppeteer-based `test/index.test.ts` integration tests) still pass.
- [x] `tsc -noEmit` clean.

Fixes #1772

### DIFF
--- a/packages/plugins/rrweb-plugin-console-record/src/index.ts
+++ b/packages/plugins/rrweb-plugin-console-record/src/index.ts
@@ -188,7 +188,12 @@ function initLogObserver(
       level,
       (original: (...args: Array<unknown>) => void) => {
         return (...args: Array<unknown>) => {
-          original.apply(this, args);
+          // `this` here is not the logger (this arrow fn's lexical `this` is
+          // whatever `replace` was called with, not the console instance).
+          // Native console methods can throw "Illegal invocation" when
+          // called with the wrong receiver (observed in extension content
+          // scripts), so bind explicitly to `_logger`.
+          original.apply(_logger, args);
 
           if (level === 'assert' && !!args[0]) {
             // assert does not log if the first argument evaluates to true
@@ -230,7 +235,8 @@ function initLogObserver(
               });
             }
           } catch (error) {
-            original('rrweb logger error:', error, ...args);
+            // same `this`-binding requirement as above
+            original.apply(_logger, ['rrweb logger error:', error, ...args]);
           } finally {
             inStack = false;
           }
--- a/packages/plugins/rrweb-plugin-console-record/src/index.ts
+++ b/packages/plugins/rrweb-plugin-console-record/src/index.ts
@@ -188,11 +188,6 @@ function initLogObserver(
       level,
       (original: (...args: Array<unknown>) => void) => {
         return (...args: Array<unknown>) => {
-          // `this` here is not the logger (this arrow fn's lexical `this` is
-          // whatever `replace` was called with, not the console instance).
-          // Native console methods can throw "Illegal invocation" when
-          // called with the wrong receiver (observed in extension content
-          // scripts), so bind explicitly to `_logger`.
           original.apply(_logger, args);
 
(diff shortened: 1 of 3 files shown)


packages/plugins/rrweb-plugin-console-record/src/index.ts
```diff
@@ -188,7 +188,12 @@ function initLogObserver(
       level,
       (original: (...args: Array<unknown>) => void) => {
         return (...args: Array<unknown>) => {
-          original.apply(this, args);
+          // `this` here is not the logger (this arrow fn's lexical `this` is
+          // whatever `replace` was called with, not the console instance).
+          // Native console methods can throw "Illegal invocation" when
+          // called with the wrong receiver (observed in extension content
+          // scripts), so bind explicitly to `_logger`.
```
Juice10: Useful for reading this PR, but not permanently in the source code
```suggestion
```

packages/plugins/rrweb-plugin-console-record/src/index.ts
```diff
@@ -230,7 +235,8 @@ function initLogObserver(
               });
             }
           } catch (error) {
-            original('rrweb logger error:', error, ...args);
+            // same `this`-binding requirement as above
```
Juice10:
```suggestion
```

Juice10 (approved):

PR #1905, opened 2026-07-17
fix(rrweb-snapshot): omit srcdoc attribute when rebuilding iframe elements

### LINKED ISSUE #1736: [Bug]: iframe with srcdoc throwing error TypeError: Failed to execute 'insertBefore' on 'Node': parameter 1 is not of type 'Node'.
### Preflight Checklist

rrweb

### Version

latest

### Expected Behavior

It should not crash or throw error.

### Actual Behavior

It should not throw error, also crashes the browser.
If we add try catch block then player wont work.

### Steps to Reproduce

Here is the JSON

https://jsfiddle.net/cwt8xgku/1/

### Testcase Gist URL

_No response_

### Additional Information

_No response_

yashrao2607:
## Summary

When rebuilding a serialized `<iframe>` that has a `srcdoc` attribute, `buildNode` was setting `srcdoc` on the live element via `node.setAttribute(name, value.toString())` like any other attribute. Setting `srcdoc` makes the browser **asynchronously parse and load its own document** into the iframe's `contentDocument` — but rrweb *also* separately reconstructs that iframe's document from its own recorded child nodes/mutations (`attachDocumentToIframe` / the mirror). These two reconstructions race, and the browser's native srcdoc load can clobber (or be clobbered by) rrweb's own tree, desyncing the mirror from the live DOM. A later mutation then targets a node that no longer exists, throwing e.g.:

```
Failed to execute 'insertBefore' on 'Node': parameter 1 is not of type 'Node'.
```

...which crashes the replayer (per the report, wrapping it in try/catch just breaks the player instead).

`rrdom`'s diff-based renderer already special-cases this exact issue (see `packages/rrdom/src/diff.ts:354` and its CHANGELOG: *"Omit the 'srcdoc' attribute of iframes to avoid overwriting content"*), but the plain browser-DOM rebuild path in `rrweb-snapshot` (used by the default web replayer) never got the same treatment. This PR applies the same fix there: skip setting `srcdoc`, since rrweb's own reconstruction is the source of truth for the iframe's contents.

## Test plan

- [x] Added a test in `rebuild.test.ts` that builds an `<iframe srcdoc="...">` node via `buildNodeWithSN` and asserts the resulting live element has no `srcdoc` attribute. Verified it fails without the fix and passes with it.
- [x] Full `rrweb-snapshot` test suite (`rebuild.test.ts`, `css.test.ts`, `snapshot.test.ts`) passes: 72/72.
- [x] `tsc -noEmit` clean.

Fixes #1736

### DIFF
--- a/.changeset/common-pants-crash.md
+++ b/.changeset/common-pants-crash.md
@@ -1,0 +1,5 @@
+---
+"rrweb-snapshot": patch
+---
+
+Omit srcdoc attribute when rebuilding iframe elements
--- a/.changeset/dark-rooms-tease.md
+++ b/.changeset/dark-rooms-tease.md
@@ -1,0 +1,5 @@
+---
+"rrweb-snapshot": patch
+---
+
+Omit the `srcdoc` attribute when rebuilding iframe elements, so the browser doesn't race its own async document load against rrweb's own reconstruction of the iframe's contents (which could desync the mirror and throw `insertBefore` errors)
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -426,6 +426,19 @@ function buildNode(
               'rrweb-original-srcset',
               n.attributes.srcset as string,
             );
+          } else if (tagName === 'iframe' && name === 'srcdoc') {
+            /**
+             * Setting `srcdoc` makes the browser asynchronously parse and load
+             * its own document into the iframe, racing against (and getting
+             * clobbered by or clobbering) the document we reconstruct for this
+             * iframe from its separately-recorded child nodes/mutations. That
+             * race can leave the mirror out of sync with the live DOM, causing
+             * later mutations to target nodes that no longer exist (e.g.
+             * insertBefore throwing "parameter 1 is not of type 'Node'").
+             * Omit it; our own reconstruction is the source of truth.
+             * @see https://github.com/rrweb-io/rrweb/issues/1736
+             */
+            // ignore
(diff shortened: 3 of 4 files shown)


Juice10 (dismissed):

Juice10 (dismissed):

eoghanmurray:
Hi @yashrao2607 

It's possible that an iframe has not been captured due to `.contentDocument` not being available on it (but srcdoc is):
   
```                                                                                                                                                                                    
... if the iframe carries a sandbox attribute that does not include allow-same-origin, the browser loads the 
srcdoc into an opaque (unique) origin. That origin is cross-origin relative to the parent, so the same-origin 
policy blocks you and iframe.contentDocument returns null:

  <!-- contentDocument === null -->
  <iframe sandbox srcdoc="<!doctype html><html>...</html>"></iframe>
  <iframe sandbox="allow-scripts" srcdoc="..."></iframe>

  <!-- contentDocument accessible again -->
  <iframe sandbox="allow-same-origin" srcdoc="..."></iframe>
  <iframe srcdoc="..."></iframe>
```

Could you expand this PR to check for either the absense of `sandbox` attribute or the presence of `allow-same-origin` on that attribute before removing srcdoc?

Also the check could be moved to record side as in this scenario we will effectively be recording the same document content twice (content might be substantial).  If we know we've got the iframe DOM serialized using same-document descent, we should blank out the srcdoc at record time to save space.

eoghanmurray (changes requested):
see my comment re. `.contentDocument`

Issue #1692, opened 2025-05-19
[Bug]: CSS splitting causes corrupted css that causes a processing error failing to replay


ShaharMintz-checksum:
### Preflight Checklist

- [x] I have searched the [issue tracker](https://www.github.com/rrweb-io/rrweb/issues) for a bug report that matches the one I want to file, without success.

### What package is this bug report for?

rrweb-snapshot

### Version

2.0.0-alpha.18

### Expected Behavior

Snapshot should be built successfully and can be played with the rrweb replayer.

**More details:**

buildStyleNode receives cssText with /* rr_split */ as following (using ... before and after the split because it's too long - see events.json in the rrweb debuger) and should be able to handle and process into a valid css to be applied

`'.pfruits-widget-newsfeed { position: fixed; top: 0px; right: 0px; left: 0px; bottom: unset; z-index: 10000991; display: flex; max-height: 100vh; min-height: 100%; min-width: 100%; flex-direction: column; overflow: hidden; border-width: 1px; border-style: solid; --tw-border-opacity: 1; border-color: rgb(243 244 246 / var(--tw-border-opacity)); --tw-bg-opacity: 1; background-color: rgb(255 255 255 / var(--tw-bg-opacity)); transition-property: opacity, transform; transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1); transition-duration: 150ms; }@media ....ot(:where([class~="not-prose"], [class~="not-prose"/* rr_split */] *)) { color: inherit; }.pfruits-banner-v3 .banner-content-wrapper :where(blockquote strong):not(:where([class~="not-prose"], ... rgba(0, 0, 0, 0.15); }.pfruits-banner-v3 .banner-main-container.limit-height { max-height: 250px; }'
`

### Actual Behavior

In reality, building the snapshot fails and the replayer final HTML is corrupted, throwing an error

```
CssSyntaxError: <css input>:25:1093: Unclosed bracket
    at Input.error (rrweb-snapshot.js:2378:17)
    at Parser.unclosedBracket (rrweb-snapshot.js:4117:22)
    at Parser.other (rrweb-snapshot.js:3986:35)
    at Parser.parse (rrweb-snapshot.js:4024:16)
    at parse$3$1 (rrweb-snapshot.js:4153:13)
    at new LazyResult (rrweb-snapshot.js:4285:17)
    at Processor.process (rrweb-snapshot.js:4788:14)
[...]

PR #1906, opened 2026-07-17
test(rrweb-snapshot): lock in the #1734/#1692 css crash fixes with regression tests

yashrao2607:
## Summary

Both #1734 and #1692 report the replayer crashing with an uncaught `CssSyntaxError` from postcss:

- **#1734**: empty longhand declarations the browser's CSSOM produces for shorthand properties combined with `var()` (e.g. `border-top-style: ;`), causing `"Unclosed string"`.
- **#1692**: a `/* rr_split */` marker landing inside a quoted attribute selector (`[class~="..."]`), so one half of the split is invalid CSS on its own, causing `"Unclosed bracket"`.

I went looking for the root cause and, checking against current `main`, both are already fixed by two earlier PRs:

- #1600 ("Fix adapt css with split") — `applyCssSplits` now rejoins all splits (`cssTextSplits.join('')`) *before* handing the text to `adaptCssForReplay`/postcss, so a split fragment is never parsed on its own. I reproduced #1692's exact failure shape (a split landing mid-attribute-selector) and confirmed it rejoins correctly without throwing.
- #1580 ("Warn instead of fail on exceptions thrown from postcss") — `adaptCssForReplay` wraps the postcss call in try/catch and falls back to the original, unmodified CSS text on failure instead of throwing. I reproduced the *exact* `"Unclosed string"` error from #1734's stack trace directly against `adaptCssForReplay` and confirmed it now degrades to a `console.warn` + original text instead of propagating and crashing the replayer.

Since I couldn't get the reporters' full page CSS (their repros are truncated CSSOM snippets from a much larger stylesheet), I can't reproduce their exact crash byte-for-byte, but I've reproduced the identical *failure mode* from each report (same postcss error type, same code path) and confirmed both are handled gracefully on `main`.

This PR doesn't change any behavior — it adds these cases as permanent regression tests (using the reporters' own snippets where possible) so the #1600/#1580 fixes can't silently regress, and so #1734 and #1692 have a paper trail for closing them.

## Test plan

- [x] New tests in `css.test.ts`:
  - `applyCssSplits css rejoiner > rejoins a split that lands inside a quoted attribute selector without corrupting it (#1692)`
  - `adaptCssForReplay with malformed css (#1734) > does not throw on empty longhand declarations produced by the browser CSSOM`
  - `adaptCssForReplay with malformed css (#1734) > falls back to the original text instead of throwing on genuinely unparseable css` — this one reproduces the literal `"Unclosed string"` postcss error from #1734's stack trace
- [x] Full `rrweb-snapshot` suite passes: 93/93 (1 pre-existing skip).
- [x] `tsc -noEmit` clean.

Addresses #1734, #1692


yashrao2607:
No changeset added here on purpose: this PR only adds regression tests (`css.test.ts`), it doesn't change any published package's runtime behavior, so it shouldn't trigger a version bump. See the PR description for why — both #1734 and #1692 turned out to already be fixed on `main` by #1600 and #1580; this just locks that in.

Juice10 (dismissed):

Juice10 (dismissed):

PR #1914, opened 2026-07-31
Click track plugin (Heatmaps)

eoghanmurray:
Add a click track plugin which generates an additional event for each click (touch or mouse) in addition to the regular rrweb [click event](https://rrweb.com/docs/observer).
The regular click event only makes sense in relation to the replay state at the time of the click (viewport position, and the internal-to-rrweb id of the target element), whereas this plugin aims to extract 'semantic' information about the click, with the primary intended use-case being Heatmaps.

The plugin uses a css selector to identify ('triangulate') an element independently of the initial FullSnapshot, without regard to uniqueness on the page. I have developed and separately published the [semantic-selector](https://github.com/eoghanmurray/semantic-selector) library to generate meaningful css selectors with the intention that the selector is useful longer term (in terms of page lifecycle) and is not overly affected by the current state of the page.
    
What it does:

  - record pointer type (touch/mouse/pen)
  - when selector is non-unique, record the index and total count of other similar elements
  - sometimes walk up the DOM tree to prefer a more significant target with a href attached (configurable, but e.g. prefer a button instead of an icon inside the button)
  - recursively walk down again to record the original `inner` target, and in the case of shadowDom, to record the target from within the shadow root 
  - record an x/y in percentage terms of the target element only (not the viewport)
  - record aspect ratio of the click target
  - record (a substring of) innerText for a significant element (configurable)
    
This was originally explored as #503 and #659

This is the first record-only plugin, there is deliberately no replay version as the existing click event covers that, but that is not stopping the plugin events from being made use of in or adjacent to the replayer e.g. for displaying enriched timeline info.


.changeset/add-click-track.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rrweb/rrweb-plugin-click-track-record": minor
```
Juice10, line 2: Just realized @rrweb/rrweb- is pretty redundant, but it is our convention currently

packages/plugins/rrweb-plugin-click-track-record/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/rrweb-io/rrweb/issues"
+  },
+  "homepage": "https://github.com/rrweb-io/rrweb#readme",
+  "dependencies": {
+    "semantic-selector": "^0.3.1"
+  },
+  "devDependencies": {
+    "rrweb": "^2.0.1",
```
Juice10, line 55: I'm surprised this doesn't also use rrweb types.

Edit: this package does import directly from rrweb types but it's not in the dependency list, we should add it

packages/plugins/rrweb-plugin-click-track-record/src/index.ts
```diff
@@ -0,0 +1,400 @@
...
+export type InnerBoundary = 'shadow';
+
+/**
+ * A resolved element paired with where within it the click landed: a CSS
+ * selector locating the element (within its own root), the click position
+ * relative to that element's box, and — recursively — the next, more precise
+ * target below it. Used for the top-level target and for every hop down toward
+ * the real clicked node (a shadow host, or a plain descendant).
+ */
+export type PositionedTarget = {
```
Juice10, line 24: A lot of these types would be better suited for the rrweb/types package as it allows you to do background processing on them without having to load every single plugin & rrweb package to be able to identify and process them

PR #1920, opened 2026-09-02
fix(rrweb-snapshot): align _cssText split points with css rule boundaries

creasty:
`applyCssSplits` hands each `/* rr_split */` part to one of the `<style>`'s text nodes. The whole reason those parts are kept separate is the one written in `buildStyleNode`'s docstring — "in case they are modified by subsequent mutations". But the split points recorded by `markCssSplits` regularly land in the **middle of a rule**, and when a mutation then inserts a sibling text node between two parts, the result is invalid css.

## The problem

`splitCssText` doesn't know where the rule boundaries are. It locates each split point by searching for a text node's content inside the browser-serialized stylesheet, normalising whitespace to compare, then converting the offset back. That lands close to the right place but not exactly on it — routinely a few characters short, in the middle of a declaration:

```
.a[aria-expanded="true"] { background-color: rgba(0, 0, 0, 0.05); color: rgba(0, 0, 0, 0.84
```

While the parts stay adjacent this is harmless: concatenating the text nodes gives back the original css, and that is what the browser parses. It stops being harmless as soon as something is inserted between them. A css-in-js library (styled-components here) adds rules at runtime by inserting a text node into the same `<style>`, positioned by `nextId` — right between two of the parts. What the browser gets is:

```
... color: rgba(0, 0, 0, 0.84.dqzkfX{cursor:pointer;-webkit-user-select:none; ...
```

The declaration never closes, so the parser goes looking for the closing brace and **discards every rule until it recovers**.

In a 43-second recording of our app (~200 text nodes in one `<style>`, 45 split points), 57 of 311 rules were dropped this way. On screen, two navigation buttons lost their styling entirely and fell back to the UA button appearance (`border: outset 2px`, `background: buttonface`) while their neighbours were fine — the difference being only whether their rule happened to sit after a splice point.

Reproducing it needs nothing exotic:

```js
// a split point that landed inside `.b`
applyCssSplits(sn, '.a { color: red; }.b { col/* rr_split */or: green; }', false, cache);
// before: childNodes[0] = '.a { color: red; }.b { col'
//         childNodes[1] = 'or: green; }'
// a later mutation inserts a text node between the two and `.b` is cut in half
// after:  childNodes[0] = '.a { color: red; }.b { color: green; }'
//         childNodes[1] = ''
```

## The fix

Move each split point forward to the end of the rule it lands in, before distributing the parts (`snapCssSplitsToRuleBoundaries`). Braces inside strings and comments aren't counted, and nested blocks (`@media`, `@supports`, `@keyframes`) close at their outermost brace.

The exact position of a split point is a guess to begin with, so nothing is lost by moving it — the concatenation is unchanged, only which text node holds which rule. And once every part is a whole number of rules, a sibling inserted between any two of them is harmless.

The same alignment is applied to `endIndex` on the `hackCss` path: `adaptCssForReplay` changes the length of the css, so the existing `searchBit` scan is used as before and then snapped to a rule boundary in the rewritten string. This also covers the case where the next part is empty after snapping and the scan has nothing to search for.

## Why on the replay side

The split points are baked into `_cssText` at record time, so fixing `splitCssText` would only help recordings made after the change ships. Doing it in `applyCssSplits` fixes recordings that already exist, which matters given how long recorder rollout takes relative to a player deploy.

## Behaviour change

One existing expectation changes: `applies css splits correctly when split parts are invalid by themselves x3` asserted per-node contents that were half-rules. The parts are now whole rules; the test asserts both the new distribution and that the concatenation is unchanged. The other existing cases are unaffected (split points that already sit on a boundary don't move, and the remerge/too-few-nodes paths are untouched).

## Tests

In `applyCssSplits css rejoiner`:

- a split point which lands inside a rule is moved to the end of that rule
- the css survives a sibling text node being inserted between the parts (the actual failure)
- braces inside strings are not mistaken for the end of a rule

And a `snapCssSplitsToRuleBoundaries` describe covering: already-aligned splits are left alone, a split point moved forward, braces in strings and comments, nested at-rules kept together, trailing splits emptied when everything snapped into an earlier one, and the single-split no-op.

All ten fail on `main` and pass with the change. `packages/rrweb-snapshot`: `vitest run` 141 passed / 1 skipped, `tsc --noEmit` clean, `eslint src` clean. `packages/rrweb` is unchanged against `main` (the same 11 puppeteer benchmark/integration tests time out on both, in a container).

### DIFF
--- a/packages/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb-snapshot/src/rebuild.ts
@@ -12,6 +12,9 @@ import {
   Mirror,
   isNodeMetaEqual,
   extractFileExtension,
+  cssRuleBoundaries,
+  nextCssRuleBoundary,
+  snapCssSplitsToRuleBoundaries,
 } from './rebuild-utils';
 import postcss from 'postcss';
 
@@ -206,7 +209,7 @@ export function applyCssSplits(
       childTextNodes.push(scn);
     }
   }
-  const cssTextSplits = cssText.split('/* rr_split */');
+  let cssTextSplits = cssText.split('/* rr_split */');
   while (
     cssTextSplits.length > 1 &&
     cssTextSplits.length > childTextNodes.length
@@ -214,9 +217,16 @@ export function applyCssSplits(
     // unexpected: remerge the last two so that we don't discard any css
     cssTextSplits.splice(-2, 2, cssTextSplits.slice(-2).join(''));
   }
+  // the split points recorded by `markCssSplits` regularly land in the middle
+  // of a rule; move them to the end of that rule so that a sibling inserted
+  // between two of these text nodes by a later mutation can't cut a rule in
+  // half (see `snapCssSplitsToRuleBoundaries`)
+  cssTextSplits = snapCssSplitsToRuleBoundaries(cssTextSplits);
   let adaptedCss = '';
+  let adaptedBoundaries: number[] = [];
   if (hackCss) {
     adaptedCss = adaptCssForReplay(cssTextSplits.join(''), cache);
+    adaptedBoundaries = cssRuleBoundaries(adaptedCss);
   }
   let startIndex = 0;
   for (let i = 0; i < childTextNodes.length; i++) {
@@ -248,6 +258,13 @@ export function applyCssSplits(
         // something went wrong, put a similar sized chunk in the right place
         endIndex += cssTextSplits[i].length;
       }
+      // `adaptCssForReplay` rewrites selectors, so the split point has to be
+      // re-aligned with the rewritten css as well
+      endIndex = nextCssRuleBoundary(
+        adaptedBoundaries,
+        endIndex,
+        adaptedCss.length,
+      );
       childTextNode.textContent = adaptedCss.substring(startIndex, endIndex);
       startIndex = endIndex;
     } else {
--- a/packages/rrweb-snapshot/test/css.test.ts
+++ b/packages/rrweb-snapshot/test/css.test.ts
@@ -5,7 +5,11 @@ import { describe, it, beforeEach, expect } from 'vitest';
 import { mediaSelectorPlugin, pseudoClassPlugin } from '../src/css';
 import postcss, { type AcceptedPlugin } from 'postcss';
 import { JSDOM } from 'jsdom';
-import { splitCssText, stringifyStylesheet } from './../src/utils';
+import {
+  snapCssSplitsToRuleBoundaries,
+  splitCssText,
+  stringifyStylesheet,
+} from './../src/utils';
 import { applyCssSplits } from './../src/rebuild';
 import * as fs from 'fs';
 import * as path from 'path';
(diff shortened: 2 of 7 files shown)


eoghanmurray:
Thanks for your contribution!  This is a gnarly bit of rrweb and I originally authored it thinking I was covering some very rare edge cases for correctness. 

> Reproducing it needs nothing exotic

So that's reproduces the fault, but I would like to see what the original text nodes looked like in a minimal example.  I don't believe that `styled-components` would arbitrarily split inside a rule (and I don't think that's what your saying; but the tests start off with the bad output of the rrweb split-serialization, rather than starting off with good output and showing how we expect rrweb to serialize it).

I think maybe a new integration test showing browser behaviour might be more valuable to ensure we don't regress this longer term.  I'll add a few more comments inline

packages/rrweb-snapshot/test/css.test.ts
```diff
@@ -408,13 +412,74 @@ describe('applyCssSplits css rejoiner', function () {
...
+  });
+
+  it('survives a sibling text node being inserted between the splits', () => {
+    // a later mutation can insert a text node between these two, which is the
+    // whole reason the split is preserved; the css has to stay valid when it
+    // does
+    const markedCssText = [
+      '.a { color: red; }.b { col',
+      'or: green; }.c { color: blue; }',
+    ].join('/* rr_split */');
```
eoghanmurray, line 457: This is starting with a broken split if I understand it correctly — is this a simulation of the 'bad/inexact' splitting?

packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -248,6 +258,13 @@ export function applyCssSplits(
         // something went wrong, put a similar sized chunk in the right place
         endIndex += cssTextSplits[i].length;
```
eoghanmurray, line 259: I'm assuming all the bad cases are produced in here ("something went wrong") so possibly we could confine the nextCssRuleBoundary call to only in here.
I'd also be interested in the input data that caused the above search to go wrong.

eoghanmurray: Above comment was written before I realized we are in rebuild here not snapshot ...

packages/rrweb-snapshot/test/css.test.ts
```diff
@@ -408,13 +412,74 @@ describe('applyCssSplits css rejoiner', function () {
       '/* rr_split */',
     );
     applyCssSplits(sn3, markedCssText, true, mockLastUnusedArg);
+    // the split points move to the end of the rule they landed in, so each
+    // text node holds whole rules rather than half of one
     expect((sn3.childNodes[0] as textNode).textContent).toEqual(
-      badStartThird.replace('.a:hover', '.a:hover,\n.a.\\:hover'),
+      '.a:hover,\n.a.\\:hover { background-color: red; }',
```
eoghanmurray, line 418: so IIRC the idea with this test was to test actual weird text nodes that *don't* land in logical css split points ... i.e. the original text nodes *were* injected arbitrarily ... it's a contrived example designed to show that we don't throw an error trying to parse any of the parts as valid CSS, so I'd reject this test modification, but rather ask you to find the root cause of what the styled-components was actually injecting (was there ever invalid css produced at record time ... if not we should not produce invalid css in our splitting ... that's the challenge)

eoghanmurray:
> Why on the replay side

I get that you want to fix at replay for your existing recordings, but this needs to be fixed at record time, so I'd love to see an example of the 'off by x chars' splits being produced, as that is not the intention for it to be inexact.

eoghanmurray (changes requested):
Request changes as I don't think moving boundaries at replay time is correct for all recorded material, and could introduce new errors.
I'd like to see a test page (even as simple as pointing me to a failing production site) so I can see where the algorithm is going wrong in the record time splitting.  If that can be identified, along with the associated insertions which are causing the trouble, then we might be able to apply the replay time fix less invasively, e.g. by doing it the prior way first, catching an exception and only then trying to shift things around

eoghanmurray:
Hi @creasty I've gone ahead and fixed on the record side in #1948, as the intention was never to have the splits ' regularly land in the middle of a rule' if they weren't authored that way.  In fact my previous #1640 (an earlier fix) introduced this new error.

Based on the findings from what generates it there, I came up with an in-algorithm fix which focuses on finding the first incorrect `0px`  while rebuilding and re-examines each split after encountering the first bad one.  Would you be able to review and deploy that #1948 PR and I'll get it merged if it still fixes all your historic recordings?

PR #1948, opened 2026-09-18
Fix bad splitCssText due to '0px' normalization

eoghanmurray:
The '0px' -> '0' normalization added in #1640 introduced a bug which caused rules to misalign as shown in the following:
    
      ┌──────────────────────┬─────────┬───────────────────────┬────────┐
      │      raw prefix      │ raw len │ normalized (baseline) │ length │
      ├──────────────────────┼─────────┼───────────────────────┼────────┤
      │ .ab { margin: 0p     │ 16      │ .ab{margin:0p         │ 13     │ <-- same length as we haven't gotten to 'x' yet so can't remove 'px' which would bring string to 12 chars
      ├──────────────────────┼─────────┼───────────────────────┼────────┤
      │ .ab { margin: 0px; } │ 20      │ .ab{margin:0}         │ 13     │
      └──────────────────────┴─────────┴───────────────────────┴────────┘
    
This starts with the record-time fix and also has a targeted replay fix for existing broken recordings.  The motivating #1920 fix/report is replay time only, and it potentially realigns all css data at replay time, even if the original css was not authored that way.


packages/rrweb-snapshot/src/rebuild.ts
```diff
@@ -219,24 +219,26 @@ export function applyCssSplits(
...
   for (let i = 0; i < childTextNodes.length; i++) {
     if (i === cssTextSplits.length) {
       break;
     }
     const childTextNode = childTextNodes[i];
-    if (!hackCss) {
-      childTextNode.textContent = cssTextSplits[i];
-    } else if (i < cssTextSplits.length - 1) {
+    let cssTextSection = cssTextSplits[i];
+    let nextTextSection =
```
eoghanmurray:
```suggestion
    const nextTextSection =
```

